PR #2, opened 2022-02-23
Create `python-publish.yml`

vtomole:


vtomole:
:+1: . Will merge once i finish testing this action locally.

PR #13, opened 2022-05-12
Add shell: bash to windows pytest

PalashGoiporia:


vtomole (changes requested):
Why are you adding another macos check? We already have one.

PalashGoiporia:
> Why are you adding another macos check? We already have one.

I wanted to add `shell: bash` to the windows test. I think I accidentally added the commit for macos again, but I don't think the code will get duplicated.

vtomole (approved):

PR #14, opened 2022-05-22
Delete `python-app.yml` in `supermarq`

vtomole:
The real `check` is here: https://github.com/SupertechLabs/SupermarQ/blob/main/.github/workflows/ci.yml


dowusu-antwi (approved):
LGTM.

Issue #16, opened 2022-06-09
Turn `SupermarQ.ipynb` into `submit.py` and `score.py`


teaguetomesh:
The main functions performed by `SupermarQ.ipynb` are:

Create a `service` object
Submit benchmarks through `cirq-superstaq`
Collect benchmark results and compute the score

We'll split these functionalities up into two separate files for now

vtomole:
Notebook was split into `.py` files and `SupermarQ.ipynb` doesn't exist anymore.

PR #18, opened 2022-06-09
Quantum Program Profiles

teaguetomesh:
Adds `features.py` to compute the different axes of the program profiles. And also adds `feature_plotting.py` to generate the radio plots as shown in the paper.

Currently, the `radar_factory` function within `feature_plotting.py` is too complex - not sure what to change to accommodate this.

Also, the function `features.cirq_to_qiskit` is needed to convert to `qiskit.QuantumCircuit` objects because that is how the profiles are generated. Does SuperstaQ already have a similar function defined elsewhere? I can edit the code to use that functionality instead.


supermarq/features.py
```diff
@@ -0,0 +1,167 @@
+from typing import Union
+
+import cirq
+import networkx as nx
+import numpy as np
+import qiskit
+
+
+def cirq_to_qiskit(circuit: cirq.Circuit) -> qiskit.circuit.QuantumCircuit:
```
vtomole, line 9: We might want to remove this and use SuperstaQ's `cirq_to_qiskit` which is more robust. This will require a token to use though. Maybe we want to keep it for people who don't have a token?

requirements.txt
```diff
@@ -1,2 +1,2 @@
-cirq>=0.13.0,<=0.14.0.dev20220105215702
-qiskit~=0.28.0
+cirq==0.15.0.dev20220420201205
```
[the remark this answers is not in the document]
teaguetomesh: I just updated them to match what's required by SuperstaQ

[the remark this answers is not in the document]
teaguetomesh: I don't think the bounds for SupermarQ need to be especially strict. As long as cirq and qiskit are "new enough" we have the functionality we need. So maybe we can change these requirements to `>=0.13` and `>=0.28`?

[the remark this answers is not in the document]
teaguetomesh: ah that's a good point - yes, I am planning to complete #16 which will involve writing functions to automatically submit circuits via `cirq-superstaq` or `qiskit-superstaq`.

What's the correct version requirement for them? Is there a specific one we need?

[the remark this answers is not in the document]
vtomole: Let's do  `qiskit-superstaq~=0.21.0` and `cirq-superstaq~=0.1.25`.

supermarq/features.py
teaguetomesh: @vtomole this is a dumb question but: in order for a user to use `superstaq.cirq_to_qiskit` they would need to import SuperstaQ, but it's a private repo?

How can we give external users access to this function?

vtomole: We'll add a new [API endpoint](https://superstaq.super.tech/api). A good reference is to rip off one of the `_compile` ones and call off to  `superstaq.cirq_to_qiskit` under the hood.

vtomole: Does this hacky `cirq_to_qiskit` work for everything here now? I'm starting to lean that we let it be for agility's sake and address the endpoints once we have the tutorial set up. The endpoints are a bonus.

teaguetomesh: It works for the current SupermarQ benchmarks. This is the same file I used to generate the feature plots shown in the paper.

vtomole: Okay. Let's get this in and prioritize putting together the tutorial for now. Let me review other parts of this PR.

teaguetomesh: Sounds good, I will actually move `cirq_to_qiskit` out into its own file: `converters.cirq_to_qiskit` because we will also want to add the option to convert circuits when the benchmarks themselves are created.

vtomole: Let's get the checks to pass as well.

teaguetomesh: > Currently, the radar_factory function within feature_plotting.py is too complex - not sure what to change to accommodate this.

Could you take a look at this

supermarq/feature_plotting.py
```diff
@@ -0,0 +1,132 @@
...
+    num_vars : int
+        Number of variables for radar chart.
+    frame : {'circle', 'polygon'}
+        Shape of frame surrounding axes.
+
+    """
+    # calculate evenly-spaced axis angles
+    theta = np.linspace(0, 2 * np.pi, num_vars, endpoint=False)
+
+    class RadarAxes(PolarAxes):
```
vtomole, line 30: Move this class outside this function.

requirements.txt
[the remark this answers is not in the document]
teaguetomesh: @vtomole did you mean `qiskit-superstaq~=0.1.21`?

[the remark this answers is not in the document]
vtomole: I did, yup.

supermarq/feature_plotting.py
teaguetomesh: Is it possible to keep `RadarAxes` inside of `radar_factory`? I think this is how these matplotlib factories are typically defined. This is because `RadarAxes` needs to know the value of `num_vars`, `frame`, and `theta`, but it is only instantiated on line 29: 

`fig, ax = plt.subplots(dpi=150, subplot_kw=dict(projection="radar"))`

So defining the class within `radar_factory` gives it access to those local variables

teaguetomesh: I can discuss this more if you'd like to hop on a call quick. But I just tried moving it outside and passing those variables to the constructor but this caused other problems with the variables that matplotlib is passing to `RadarAxes` under the hood

vtomole: What we discussed: Create 

```
class RadarAxesMeta(PolarAxes):
```
and 
```
def radar_factory(num_vars, frame="circle"):
      class RadarAxes(RadaAxesMeta):
```

vtomole (commented):
Please ad a comment justifying each `pragma: no cover` and i'll see if we really need to ignore it.

supermarq/features.py
```diff
@@ -0,0 +1,163 @@
...
+
+    dag = qiskit.converters.circuit_to_dag(circ)
+    dag.remove_all_ops_named("barrier")
+
+    return len(dag.two_qubit_ops()) / len(dag.gate_nodes())
+
+
+def compute_depth(circuit: Union[cirq.Circuit, qiskit.circuit.QuantumCircuit]) -> float:
+    """
+    depth feature = # of 2-qubit gates along the critical path / total # of 2-qubit gates
```
vtomole, line 137: Change docstring to be in the style of Google's recommendation: https://google.github.io/styleguide/pyguide.html

supermarq/features.py
```diff
@@ -0,0 +1,163 @@
...
+            if op.name == "reset":
+                reset_present = True
+        if reset_present:
+            reset_moments += 1
+
+    return reset_moments / gate_depth
+
+
+def compute_entanglement(circuit: Union[cirq.Circuit, qiskit.circuit.QuantumCircuit]) -> float:
+    """
```
vtomole, line 116: Change docstring to be in the style of Google's recommendation: https://google.github.io/styleguide/pyguide.html

supermarq/features.py
```diff
@@ -0,0 +1,163 @@
...
+        circ = circuit
+
+    dag = qiskit.converters.circuit_to_dag(circ)
+    dag.remove_all_ops_named("barrier")
+    return max(1 - (circ.depth() / len(dag.gate_nodes())), 0)
+
+
+def compute_measurement(circuit: Union[cirq.Circuit, qiskit.circuit.QuantumCircuit]) -> float:
+    """
+    measurement feature = # of layers of mid-circuit measurement / circuit depth
```
vtomole, line 87: Change docstring to be in the style of Google's recommendation: https://google.github.io/styleguide/pyguide.html

supermarq/features.py
```diff
@@ -0,0 +1,163 @@
...
+    if isinstance(circuit, cirq.Circuit):
+        circ = sm.converters.cirq_to_qiskit(circuit)
+    else:
+        circ = circuit
+
+    dag = qiskit.converters.circuit_to_dag(circ)
+    dag.remove_all_ops_named("barrier")
+
+    reset_moments = 0
+    gate_depth = dag.depth() - 1  # assumes that circuit measures all qubits at the end
```
vtomole, line 102: Does qiskit have a way to assert that measurements are indeed terminal before we call this? Seems like currently there is a way for this to silently fail.

supermarq/features.py
```diff
@@ -0,0 +1,163 @@
...
+        for op in layer["partition"]:
+            for qubit in op:
+                activity_matrix[qubit.index, i] = 1
+
+    return np.sum(activity_matrix) / (N * dag.depth())
+
+
+def compute_parallelism(circuit: Union[cirq.Circuit, qiskit.circuit.QuantumCircuit]) -> float:
+    """
+    parallelism feature = max(1 - depth / # of gates, 0)
```
vtomole, line 68: Change docstring to be in the style of Google's recommendation: https://google.github.io/styleguide/pyguide.html

supermarq/features.py
```diff
@@ -0,0 +1,163 @@
...
+    -----
+    circ : A cirq or qiskit quantum circuit
+    """
+
+    if isinstance(circuit, cirq.Circuit):
+        circ = sm.converters.cirq_to_qiskit(circuit)
+    else:
+        circ = circuit
+
+    N = circ.num_qubits
```
vtomole, line 52: Variable names shouldn't be upper case like this.

supermarq/features.py
```diff
@@ -0,0 +1,163 @@
...
+        G.add_edge(q1.index, q2.index)
+
+    degree_sum = sum([G.degree(n) for n in G.nodes])
+
+    return degree_sum / (N * (N - 1))
+
+
+def compute_liveness(circuit: Union[cirq.Circuit, qiskit.circuit.QuantumCircuit]) -> float:
+    """
+    liveness feature = sum of all entries in the liveness matrix / (num_qubits * depth)
```
vtomole, line 40: Change docstring to be in the style of Google's recommendation: https://google.github.io/styleguide/pyguide.html

supermarq/features.py
```diff
@@ -0,0 +1,163 @@
...
+    -----
+    circ : A cirq or qiskit quantum circuit
+    """
+
+    if isinstance(circuit, cirq.Circuit):
+        circ = sm.converters.cirq_to_qiskit(circuit)
+    else:
+        circ = circuit
+
+    N = circ.num_qubits
```
vtomole, line 24: Variables names shouldn't be upper case

supermarq/features.py
```diff
@@ -0,0 +1,163 @@
...
+    dag = qiskit.converters.circuit_to_dag(circ)
+    dag.remove_all_ops_named("barrier")
+
+    G = nx.Graph()
```
vtomole, line 28: Variable names shouldn't be upper case

supermarq/features.py
```diff
@@ -0,0 +1,163 @@
...
+import cirq
+import networkx as nx
+import numpy as np
+import qiskit
+import supermarq as sm
+
+
+def compute_communication(circuit: Union[cirq.Circuit, qiskit.circuit.QuantumCircuit]) -> float:
+    """
+    communication feature = circuit's average qubit degree / degree of a complete graph
```
vtomole, line 12: Change docstring to be in the style of Google's recommendation: https://google.github.io/styleguide/pyguide.html

supermarq/feature_plotting.py
```diff
@@ -0,0 +1,135 @@
...
+
+    if show:
+        plt.show()  # pragma: no cover
+
+    plt.close()
+
+
+def radar_factory(num_vars: int, frame: str = "circle") -> np.ndarray:
+    """
+    (https://matplotlib.org/stable/gallery/specialty_plots/radar_chart.html)
```
vtomole, line 61: Change docstring to be in the style of Google's recommendation: https://google.github.io/styleguide/pyguide.html

supermarq/feature_plotting.py
```diff
@@ -0,0 +1,135 @@
...
+
+def plot_benchmark(
+    data: List[Union[str, List[str], List[List[float]]]],
+    show: bool = True,
+    savefn: Optional[str] = None,
+    spoke_labels: Optional[List[str]] = None,
+    legend_loc: Tuple[float, float] = (0.75, 0.85),
+) -> None:
+    """
+    Create a radar plot of the given benchmarks.
```
vtomole, line 19: Change docstring to be in the style of Google's recommendation: https://google.github.io/styleguide/pyguide.html

supermarq/features.py
teaguetomesh: Good catch -- updated this so instead of potentially dividing by zero, we will remove the final measurements in the circuit if they exist.

vtomole:
Looks like docstrings haven't been implemented Google style yet. Here is an example:

```python
def fetch_smalltable_rows(table_handle: smalltable.Table,
                          keys: Sequence[Union[bytes, str]],
                          require_all_keys: bool = False,
) -> Mapping[bytes, tuple[str, ...]]:
    """Fetches rows from a Smalltable.

    Retrieves rows pertaining to the given keys from the Table instance
    represented by table_handle.  String keys will be UTF-8 encoded.

    Args:
      table_handle:
        An open smalltable.Table instance.
      keys:
        A sequence of strings representing the key of each table row to
        fetch.  String keys will be UTF-8 encoded.
      require_all_keys:
        If True only rows with values set for all keys will be returned.

    Returns:
      A dict mapping keys to the corresponding table row data
      fetched. Each row is represented as a tuple of strings. For
      example:

      {b'Serak': ('Rigel VII', 'Preparer'),
       b'Zim': ('Irk', 'Invader'),
       b'Lrrr': ('Omicron Persei 8', 'Emperor')}

      Returned keys are always bytes.  If a key from the keys argument is
      missing from the dictionary, then that row was not found in the
      table (and require_all_keys must have been False).

    Raises:
      IOError: An error occurred accessing the smalltable.
    """
```

supermarq/features_test.py
```diff
@@ -0,0 +1,70 @@
+import cirq
+import qiskit
+import supermarq as sm
+
+
+qiskit_circuit = qiskit.QuantumCircuit(2, 1)
```
vtomole, line 6: Get rid of `qiskit-circuit`. As we discussed IRL, it's redundant.

supermarq/features.py
```diff
@@ -0,0 +1,170 @@
...
+    else:
+        circ = circuit
+
+    dag = qiskit.converters.circuit_to_dag(circ)
+    dag.remove_all_ops_named("barrier")
+
+    return len(dag.two_qubit_ops()) / len(dag.gate_nodes())
+
+
+def compute_depth(circuit: Union[cirq.Circuit, qiskit.circuit.QuantumCircuit]) -> float:
```
vtomole, line 141: Get rid of `qiskit.QuantumCircuit`

supermarq/features.py
```diff
@@ -0,0 +1,170 @@
...
+    """Compute the critical depth of the given quantum circuit.
+
+    Critical depth = # of 2-qubit gates along the critical path / total # of 2-qubit gates.
+
+    Input
+    -----
+    circuit : A cirq or qiskit quantum circuit
```
vtomole, line 148: Update comment to reflect the removal of Qiskit circuit input.

supermarq/features.py
```diff
@@ -0,0 +1,170 @@
...
+    """
+
+    if isinstance(circuit, cirq.Circuit):
```
vtomole, line 151: No need for this `if else` when we change this to only Cirq circuits.

vtomole (changes requested):
I've left some comments that should apply to all the feature sections.

supermarq/feature_plotting.py
```diff
@@ -0,0 +1,128 @@
...
+        spoke_labels = ["Connectivity", "Liveness", "Parallelism", "Measurement", "Entanglement"]
+
+    N = len(spoke_labels)
+    theta = radar_factory(N)
+
+    _, ax = plt.subplots(dpi=150, subplot_kw=dict(projection="radar"))
+    # fig.subplots_adjust(wspace=0.25, hspace=0.20, top=0.85, bottom=0.05)
+
+    _, labels, case_data = data
+    # ax.set_rgrids([0.2, 0.4, 0.6, 0.8])
```
vtomole, line 38: Remove this.

supermarq/feature_plotting.py
```diff
@@ -0,0 +1,128 @@
...
+    ax.set_rgrids([0.0, 0.2, 0.4, 0.6, 0.8, 1.0])
+    # ax.set_title(title, weight='bold', size='medium', position=(0.5, 1.1),
```
vtomole, line 40: Remove

supermarq/feature_plotting.py
```diff
@@ -0,0 +1,128 @@
...
+    #             horizontalalignment='center', verticalalignment='center')
```
vtomole, line 41: Remove

supermarq/features.py
teaguetomesh: Other users may generate qiskit circuits and still want to compute these features though.

teaguetomesh:
> Looks like docstrings haven't been implemented Google style yet. Here is an example:
> 
> ```python
> def fetch_smalltable_rows(table_handle: smalltable.Table,
>                           keys: Sequence[Union[bytes, str]],
>                           require_all_keys: bool = False,
> ) -> Mapping[bytes, tuple[str, ...]]:
>     """Fetches rows from a Smalltable.
> 
>     Retrieves rows pertaining to the given keys from the Table instance
>     represented by table_handle.  String keys will be UTF-8 encoded.
> 
>     Args:
>       table_handle:
>         An open smalltable.Table instance.
>       keys:
>         A sequence of strings representing the key of each table row to
>         fetch.  String keys will be UTF-8 encoded.
>       require_all_keys:
>         If True only rows with values set for all keys will be returned.
> 
>     Returns:
>       A dict mapping keys to the corresponding table row data
>       fetched. Each row is represented as a tuple of strings. For
>       example:
> 
>       {b'Serak': ('Rigel VII', 'Preparer'),
>        b'Zim': ('Irk', 'Invader'),
>        b'Lrrr': ('Omicron Persei 8', 'Emperor')}
> 
>       Returned keys are always bytes.  If a key from the keys argument is
>       missing from the dictionary, then that row was not found in the
>       table (and require_all_keys must have been False).
> 
>     Raises:
>       IOError: An error occurred accessing the smalltable.
>     """
> ```

This will be handled in #26

supermarq/features.py
teaguetomesh: Since the features need `qiskit.QuantumCircuit` anyway, it seems very easy to support both. I agree that we can keep the benchmark generators simple -> only generate cirq circuits. But the features can easily handle both

supermarq/feature_plotting.py
```diff
@@ -0,0 +1,128 @@
...
+
+    def plot(self, *args: Any, **kwargs: Any) -> None:
+        """Override plot so that line is closed by default"""
+        lines = super().plot(*args, **kwargs)
+        for line in lines:
+            self._close_line(line)
+
+    def _close_line(self, line: matplotlib.lines.Line2D) -> None:
+        x, y = line.get_data()
+        # FIXME: markers at x[0], y[0] get doubled-up
```
vtomole, line 113: File an issue for this and link to it in this comment

supermarq/features.py
vtomole: Ah, sure. I see now. This is fine.

supermarq/features_test.py
vtomole: nvm this comment. Was resolved above,

supermarq/features.py
vtomole: Actually, still haven't' decided. I think we want people to do `compute_depth(cirq_to_qiskit(qiskit_circuit)`. T

teaguetomesh: If our benchmarks are always returning cirq circuits, it doesn't make sense to me to force them to add this extra code when they want to compute the feature vectors. They should be able to enter in the exact output of the benchmark without needing to alter it

teaguetomesh: If our benchmarks are always returning cirq circuits, it doesn't make sense to me to force them to add this extra code when they want to compute the feature vectors. They should be able to enter in the exact output of the benchmark without needing to alter it

vtomole (approved):
Have to merge due to time constraints.

teaguetomesh:
Per our discussion: wrappers make sense here to keep sdks separate for simplicity's sake.

LGTM

PR #39, opened 2022-07-12
26 update docstrings to google style

alexabuko:
These are the docstrings I have changed so far. Let me know if there is anything wrong!


vtomole:
I won't forget to re-open this when we do something similar to https://github.com/SupertechLabs/general-superstaq/pull/218 before the repo-merge.

PR #40, opened 2022-07-13
updates in light of check script changes

perlinm:
Two points I'm unsure about in this PR:
- looks like `supermarq` already has a `dev` version, but it's called `dev_env`.  Do we want to rename this to mirror `app-ss[dev] / css[dev] / qss[dev]`?
- should we require `css[dev]` and `qss[dev]` in `dev-requirements.txt`, or `app-ss[dev]`?

UPDATE: we will make it `supermarq[dev]`, and I made `dev-requirements.txt` require `app-ss[dev]`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
           PRIVATE_REPOS_KEY: ${{ secrets.PRIVATE_REPOS_KEY }}
         run: |
           python -m pip install --upgrade pip
-          pip install .[dev_env]
+          pip install .[dev]
       - name: Format
         run: |
           check/format_.py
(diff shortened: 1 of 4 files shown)


vtomole:
>  Do we want to rename this to mirror app-ss[dev] / css[dev] / qss[dev]

Yup.

richrines1:
> should we require `css[dev]` and `qss[dev]` in `dev-requirements.txt`, or `app-ss[dev]`?

if it works the latter seems ideal to me. right now it shouldn't matter but maybe someday we'll have some other requirements for checks in `[cq]ss` which we shouldn't need to pull in here

perlinm:
I think the checks for this PR won't work until we deploy `css` and `qss`

perlinm:
@richrines1 @vtomole ready for review.  Note: aside from `[dev_env] --> [dev]`, changes to `setup.py` were made largely to have it look the same as in `css/qss`

vtomole (approved):
LGTM

PR #56, opened 2022-10-25
Update css/qss requirements

paaige:


perlinm (approved):
Just curious: is there a reason we need standalone updates?  I feel like we can just update requirements in PRs (such as #55) that make other changes.

paaige:
> Just curious: is there a reason we need standalone updates? I feel like we can just update requirements in PRs (such as #55) that make other changes.

True, I'm wary of adding things irrelevant to the main issue to a PR but in this case it's probably fine.

PR #70, opened 2023-02-08
Update dependencies to match SuperstaQ

teaguetomesh:
Quick fix to keep SupermarQ dependencies up-to-date with SuperstaQ. 

@vtomole I saw another PR #62 and some others in SuperstaQ that you have an automated way to update dependencies. Could we use that here to keep SupermarQ up to date with the latest css and qss deployments? 

The css and qss packages are mainly used for circuit submissions so automatically updating them shouldn't introduce breaking changes, plus we should probably keep SupermarQ up-to-date with them regardless.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-cirq~=1.0.0
-cirq-superstaq~=0.3.16
+cirq~=1.2.0.dev2022
+cirq-superstaq~=0.3.17
 qiskit-superstaq~=0.3.16
 scikit-learn~=1.0  # supports python 3.7


vtomole (approved):
SupermarQ has [requirements](https://github.com/SupertechLabs/SupermarQ/blob/main/check/requirements.py) checks

requirements.txt
```diff
@@ -1,4 +1,4 @@
-cirq~=1.0.0
-cirq-superstaq~=0.3.16
+cirq~=1.2.0.dev2022
```
richrines1, line 1: does supermarq actually need the pre-release version of cirq? if not we should get rid of this requirement so it just uses the cirq version required by cirq-superstaq

perlinm: Imo we should opt to have loose requirements in public-facing repos.  For example [`css` only requires `cirq>=0.14.0`](https://github.com/SupertechLabs/cirq-superstaq/blob/main/requirements.txt).  It's fine to update to, say, `>=1.0.0` or `1.2.0` if we rely on features therein, but I agree with @richrines1 and think that pre-releases should be avoided.

PR #72, opened 2023-02-16
Update requirements

teaguetomesh:
Responding to @richrines1's comment on #70, I tested behavior with and without the cirq dev requirement. All of the checks still pass so creating this PR to remove the explicit dependency.


richrines1 (approved):
lgtm! (also just confirmed the tests still pass with cirq 0.14.0, which is the oldest version accepted by css)

teaguetomesh:
@richrines1 @perlinm yesterday I removed the explicit cirq dependence in SupermarQ and instead we just require css and qss. (@richrines1 and I checked that all the checks still pass with the versions of cirq that css requires) and then all of the checks were passing here, but once I merged into main it failed the Windows pytest -- why does this happen?

richrines1:
@teaguetomesh the same test passed when i reran it, so might have just been a weird fluke

vtomole:
Yeah it's a fluke. Let's be on the lookout if it happens again.

PR #77, opened 2023-02-28
remove generic types

richrines1:
extends https://github.com/SupertechLabs/general-superstaq/pull/167 to supermarq in anticipation of repo merge


vtomole (approved):
The fact that we used `collections.Counter` instead of a `Dict` in the first place is a whole crime :rofl:

PR #81, opened 2023-03-06
Docstring bugsquashing

teaguetomesh:


supermarq/converters.py
```diff
@@ -4,9 +4,10 @@
 import qiskit
 
 
-def cirq_to_qiskit(  # pylint: disable=missing-function-docstring
+def cirq_to_qiskit(
     circuit: cirq.Circuit,
 ) -> qiskit.circuit.QuantumCircuit:
+    """Convert a circuit from `cirq` to `qiskit`."""
```
vtomole, line 10: Why is pylint letting you get away with specifying `Args` and `Returns` here? CC @pgoiporia

richrines1: the current checks don't occur if there's no "Args:"/"Returns:"/"Raises:"/etc in the docstring. To require these sections be present whenever applicable, we also have to add these to setup.cfg:
```
accept-no-param-doc=no
accept-no-raise-doc=no
accept-no-return-doc=no
accept-no-yields-doc=no
```

paaige: Do we want to always explicitly specify the args / returns / raises when it's extremely obvious what they represent? I've basically been sort of leaving them blank unless it's not super obvious (for ex. a variable name/return type is not descriptive)

supermarq/simulation.py
```diff
@@ -4,9 +4,24 @@
 import numpy as np
 
 
-def get_ideal_counts(  # pylint: disable=missing-function-docstring
+def get_ideal_counts(
     circuit: cirq.Circuit,
 ) -> Dict[str, float]:
```
richrines1, line 9:
```suggestion
def get_ideal_counts(circuit: cirq.Circuit) -> Dict[str, float]:
```

supermarq/stabilizers.py
```diff
@@ -148,19 +224,32 @@ def transform_X_matrix_to_row_echelon_form(  # pylint: disable=missing-function-
                 apply_CNOT(measurement_circuit, j, i - N)
 
 
-def transform_X_matrix_to_reduced_row_echelon_form(  # pylint: disable=missing-function-docstring
+def transform_X_matrix_to_reduced_row_echelon_form(
     measurement_circuit: MeasurementCircuit,
 ) -> None:
```
richrines1, line 229:
```suggestion
def transform_X_matrix_to_reduced_row_echelon_form(measurement_circuit: MeasurementCircuit) -> None:
```

supermarq/stabilizers.py
```diff
@@ -148,19 +224,32 @@ def transform_X_matrix_to_row_echelon_form(  # pylint: disable=missing-function-
...
     for j in range(N - 1, 0, -1):
         for i in range(N, N + j):
             if measurement_circuit.get_stabilizer()[i, j] == 1:
                 apply_CNOT(measurement_circuit, j, i - N)
 
 
-def patch_Z_matrix(  # pylint: disable=missing-function-docstring
+def patch_Z_matrix(
     measurement_circuit: MeasurementCircuit,
 ) -> None:
```
richrines1, line 247:
```suggestion
def patch_Z_matrix(measurement_circuit: MeasurementCircuit) -> None:
```

supermarq/stabilizers.py
```diff
@@ -176,18 +265,27 @@ def patch_Z_matrix(  # pylint: disable=missing-function-docstring
             apply_S(measurement_circuit, i)
 
 
-def change_X_to_Z_basis(  # pylint: disable=missing-function-docstring
+def change_X_to_Z_basis(
     measurement_circuit: MeasurementCircuit,
 ) -> None:
```
richrines1, line 270:
```suggestion
def change_X_to_Z_basis(measurement_circuit: MeasurementCircuit) -> None:
```

supermarq/converters.py
richrines1: was wondering this as well. on one hand the consistency is nice, and combined with the checker it forces us to update the docstring if e.g. we add a new argument to a function. but it's also hard to say something like
```python
def qiskit_gate_to_cirq_gate(qiskit_gate: qiskit.circuit.Gate) -> cirq.Gate:
```
really needs much explanation

i think overall i'm in favor of requiring them always: i don't think there are \*that\* many cases where they aren't at least a little bit helpful, and it'd be nice to be able to have pylint require them so that they're never missing in places they actually are useful. could definitely be convinced otherwise though

paaige: yeah I agree, and it might even be useful in the simpler cases. I can see a situation where someone might change the return type, without also having to update the docstrings to represent the change it's easy to let hacky solutions slip through the cracks

supermarq/converters.py
```diff
@@ -4,9 +4,16 @@
 import qiskit
 
 
-def cirq_to_qiskit(  # pylint: disable=missing-function-docstring
+def cirq_to_qiskit(
     circuit: cirq.Circuit,
 ) -> qiskit.circuit.QuantumCircuit:
```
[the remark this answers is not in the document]
vtomole: Not related to this PR: a low priority intern project for this summer would be moving the SuperstaQ converters to SupermarQ.

richrines1 (approved):
lgtm!

PR #82, opened 2023-03-10
Add QEC benchmarks: surface code and Bacon-Shor code

perlinm:
Dumping scripts for now.  These need to be converted into benchmarks.


supermarq/benchmarks/bacon_shor_code.py
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env python3
+import cirq
+
+
+def get_code_cycle(rows: int, cols: int) -> cirq.Circuit:
```
teaguetomesh, line 5: These benchmarks look great! It'll be nice to have a few more ECC tailored apps.

First thing, define a `BaconShorCode` class to represent the benchmark. See for example the [GHZ benchmark](https://github.com/SupertechLabs/SupermarQ/blob/main/supermarq/benchmarks/ghz.py).

supermarq/benchmarks/bacon_shor_code.py
```diff
@@ -0,0 +1,44 @@
...
+    for (pauli, row, col), ancilla in ancillas.items():
+        qubit = cirq.GridQubit(row, col)
+        step = (0, 1) if pauli == cirq.Z else (1, 0)
+        pauli_ops = {qubit: pauli, qubit + step: pauli}
+        parity_op: cirq.PauliString[cirq.GridQubit] = cirq.PauliString(pauli_ops)
+        circuit += cirq.decompose_once(parity_op.controlled_by(ancilla))
+    circuit += hadamards
+
+    return circuit
+
```
teaguetomesh, line 34: Define a `circuit()` function which uses the functions you've already defined here and returns the benchmark circuit.

supermarq/benchmarks/bacon_shor_code.py
teaguetomesh: The `__init__()` function should take in all of the appropriate benchmark parameters as inputs.

supermarq/benchmarks/bacon_shor_code.py
```diff
@@ -0,0 +1,44 @@
...
+
```
teaguetomesh, line 35: Define a `score()` function that takes in a `Dict[str, int]` of counts and returns a float representing the performance of this benchmark. Thus far we've been defining all of the benchmark scores to be normalized to a range `[0,1]`, we can have some discussion about what the right metric here is. The only requirement is that _in principle_ it should be a scalable verification metric.

supermarq/benchmarks/bacon_shor_code.py
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env python3
```
teaguetomesh, line 1: This can be removed, unless it's necessary for something?

supermarq/benchmarks/surface_code.py
```diff
@@ -0,0 +1,146 @@
+#!/usr/bin/env python3
```
teaguetomesh, line 1: This can be removed, unless it's necessary for something?

supermarq/benchmarks/surface_code.py
```diff
@@ -0,0 +1,146 @@
...
+import functools
+from typing import Dict, Tuple
+
+import cirq
+
+
```
teaguetomesh, line 7: Same as above, define a `SurfaceCode` benchmark class with corresponding `__init__`

supermarq/benchmarks/surface_code.py
```diff
@@ -0,0 +1,146 @@
...
+        neighbors = [
+            neighbor for neighbor in _get_diagonal_neighbors(ancilla) if _is_on_this_patch(neighbor)
+        ]
+        pauli = _qubit_pauli(ancilla)
+        pauli_ops = {qubit: pauli for qubit in neighbors}
+        parity_op: cirq.PauliString[cirq.GridQubit] = cirq.PauliString(pauli_ops)
+        circuit += cirq.decompose_once(parity_op.controlled_by(ancilla))
+    circuit += hadamards
+
+    return circuit
```
teaguetomesh, line 41: Define a `circuit()` function

supermarq/benchmarks/surface_code.py
```diff
@@ -0,0 +1,146 @@
...
+    # transform data qubits to the ones used by the code cycle
+    current_qubits = sorted(circuit.all_qubits())
+    code_cycle_qubits = sorted(_get_data_qubits(rows, cols))
+    qubit_map: Dict[cirq.Qid, cirq.Qid] = dict(zip(current_qubits, code_cycle_qubits))
+    circuit = circuit.transform_qubits(qubit_map)
+
+    if variant_XZZX:
+        # hadamard transform every other data qubit
+        circuit += [cirq.H(qubit) for qubit in code_cycle_qubits if _qubit_pauli(qubit) == cirq.X]
+    return circuit
```
teaguetomesh, line 146: Define a `score()` function

teaguetomesh (changes requested):
Overall these look great. Main thing is to define a benchmark class for each, equipped with circuit and score functions.

PR #83, opened 2023-03-13
Discrete time crystal benchmark

teaguetomesh:
Another Hamiltonian simulation based benchmark. Could be a good addition with the added benefit that the simulation is exact in the sense that it doesn't require Trotterization.

Created the PR with plans to run some of these circuits on Hilbert soon. (And other devices, but those have easier time schedules)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-cirq-superstaq~=0.3.19
-matplotlib<=3.5.2
+cirq-superstaq~=0.3.20
+matplotlib>=3.5.2
 qiskit-superstaq~=0.3.19
 scikit-learn~=1.0  # supports python 3.7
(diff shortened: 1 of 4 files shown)


vtomole (commented):
Is this scalable?

teaguetomesh:
> Is this scalable?

Good question - so the score function I have right now is just comparing against noiseless simulation. In principle we can find scalable performance metrics because we know that every time a cycle is added the average polarization of each qubit should flip from +1 -> -1 -> +1 -> ... and so on. So in a noiseless setting these time crystals have periodic behavior that can be exploited wrt to scalable metrics.

The one caveat I'll throw in here is that usually these metrics like the number of [visible cycles](https://arxiv.org/pdf/2301.07625.pdf) are defined over multiple cycles -> which means that each `Benchmark` instance would actually be composed of multiple circuits (like the current VQE proxy). 

So I think that the specific `score` function for this benchmark will change over time depending on what we decide we want to do, but for now it's useful to have it available as another Hamiltonian simulation type of circuit that is distinct from TFIM.

PR #84, opened 2023-03-14
workflows for automated releases

richrines1:
(same as https://github.com/SupertechLabs/cirq-superstaq/pull/434 / https://github.com/SupertechLabs/qiskit-superstaq/pull/311 / https://github.com/SupertechLabs/general-superstaq/pull/220)


stephanielee9 (commented):
Compared to the files changed in https://github.com/SupertechLabs/cirq-superstaq/pull/434, and looks good. Also compared to `update-pypi-version.sh` file in `css` and looked the same, but didn't do any test to check if the publish works - are there any tests I can run or is comparing file contents good enough?

richrines1:
> Compared to the files changed in [SupertechLabs/cirq-superstaq#434](https://github.com/SupertechLabs/cirq-superstaq/pull/434), and looks good. Also compared to `update-pypi-version.sh` file in `css` and looked the same, but didn't do any test to check if the publish works - are there any tests I can run or is comparing file contents good enough?

i think as long as everything looks reasonable/consistent with the other repos that should be fine - it doesn't touch any of the actual code, so it seems like the best way to test it is to merge it and see what happens when we try to deploy

stephanielee9 (approved):

PR #85, opened 2023-03-14
clean up in anticipation of repo merge

richrines1:
* syncs .gitignore, license, and .github folder contents to align with those in *superstaq repos
* adds caching and tests for other python versions to ci.yml


LICENSE
```diff
@@ -178,15 +178,15 @@
...
+      boilerplate notice, with the fields enclosed by brackets "{}"
       replaced with your own identifying information. (Don't include
       the brackets!)  The text should be enclosed in the appropriate
       comment syntax for the file format. We also recommend that a
       file or class name and description of purpose be included on the
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright [yyyy] [name of copyright owner]
+   Copyright {yyyy} {name of copyright owner}
```
perlinm, line 189: Shouldn't we actually insert the year and name (Super.tech or ColdQuanta Inc.?) into here?

vtomole: No this is a template of what you are supposed to put on top of every source file in the repo; which we don't do, but should.

vtomole: Why is this file being touched in the first place?

richrines1: nope - this is supposed to be a template which should be filled in when copied into other files (which is described in the paragraph above). the license text itself is supposed to be included verbatim (interestingly qiskit-terra [gets this wrong](https://github.com/Qiskit/qiskit-terra/blob/bf3bb96cebc141f79ce46b5938724eef30778c8e/LICENSE.txt#L191) though)

vtomole: Given that we are so not lawyers, let's just default to the [Cirq way](https://github.com/quantumlib/Cirq/blob/master/LICENSE) of doing this here.

richrines1: > Why is this file being touched in the first place?

because our other repos already [use curly brackets](https://github.com/SupertechLabs/cirq-superstaq/blob/2f1d71ec31ae5401f4e88207e02f2dced27d40b5/LICENSE#L189) instead of square. (the point of this pr is to minimizes conflicts when everything is merged)

vtomole: Ah, good.

richrines1: yeah cirq is definitely right here - the license text itself stops making sense if you fill it out (and it's also what other every other non-qiskit repo does). seems like an oversight in qiskit

.gitignore
```diff
@@ -1,9 +1,8 @@
-.DS_Store
```
vtomole, line 1: Is this a duplicate or something?

vtomole (approved):
LGTM module one question.

.gitignore
richrines1: yep - it appears again [at the end of the file](https://github.com/SupertechLabs/SupermarQ/blob/2e2c049fa8c09fe941a9d8514c4b763f5c8cfdbe/.gitignore#L153)

vtomole: :+1:

PR #390, opened 2023-03-16
fix requirements check and sync versions

richrines1:
now that we sync package versions and publish everything at once, we should be putting the same version number in every package's requirements.txt. this fails the requirements check though, because we have to update the requirements before publishing anything to pypi

this is a hotfix which uses `gss.__version__` instead of the pypi version if the former is larger. we should really rethink how the requirements check should work now though


general-superstaq/general_superstaq/check/check_utils.py
```diff
@@ -13,10 +13,13 @@
...
-root_dir = subprocess.check_output(
-    ["git", "rev-parse", "--show-toplevel"], cwd=main_file_dir, text=True
-).strip()
+try:
+    main_file_dir = os.path.dirname(os.path.abspath(sys.argv[0]))
+    root_dir = subprocess.check_output(
+        ["git", "rev-parse", "--show-toplevel"], cwd=main_file_dir, text=True
+    ).strip()
+except subprocess.CalledProcessError:
+    root_dir = subprocess.check_output(["git", "rev-parse", "--show-toplevel"], text=True).strip()
```
richrines1, line 22: (this change just makes it so we can import `general_superstaq.check` in a python terminal, which is really nice for debugging)

general-superstaq/general_superstaq/check/.isort.cfg
```diff
@@ -0,0 +1 @@
+../../.isort.cfg
```
richrines1, line 1: (this is necessary so `general-superstaq/general_superstaq/check/setup.cfg` doesn't override `general-superstaq/.isort.cfg` in this directory)

vtomole (approved):

PR #394, opened 2023-03-16
make all other `_version.py` files symlinks to gss's

richrines1:
(they should always be the same so no reason to maintain four different files)


vtomole (approved):
LGTM

PR #403, opened 2023-03-17
clean up example-requirements

richrines1:


cirq-superstaq/example-requirements.txt
```diff
@@ -1,3 +1,2 @@
 notebook~=6.4.12
-pulser~=0.7.0
-qiskit>=0.35.0, <0.37.0
```
paaige, line 3: so we're no longer using pulser or was this package merged into qiskit-terra?

paaige (approved):

cirq-superstaq/example-requirements.txt
richrines1: no longer using it - we just missed these lines in https://github.com/SupertechLabs/cirq-superstaq/pull/439 & https://github.com/SupertechLabs/qiskit-superstaq/pull/315

PR #411, opened 2023-03-22
fix terms of use checking

richrines1:
previously [this check](https://github.com/SupertechLabs/client-superstaq/blob/main/general-superstaq/general_superstaq/superstaq_client.py#L363-L365) didn't catch anything, because `response.text` always contains the json-serialized response string, e.g.
```python
'"You must accept the Terms of Use (superstaq.super.tech/terms_of_use)."\n'
```
instead of just
```python
"You must accept the Terms of Use (superstaq.super.tech/terms_of_use)."
```
Calling `response.json()` instead deserializes the response content, and so returns the correct string for comparison

confirmed locally that after this change the message handling works as expected:
screenshot


knsmith (approved):
LGTM!

vtomole (approved):

PR #416, opened 2023-03-27
allow `--randomly-seed=` (and other pytest args) to be passed to coverage check

### LINKED ISSUE #414: Can't pass `--randomly-seed` to `check/coverage._.py`
### What is happening?

If `pytest-randomly` is installed, the coverage check should allow us to pick a seed. This will make it easier to reproduce failing/flaking tests.

`./check/coverage_.py --randomly-seed=3399423474` should work.

Manually add it to `pytest_args` in `gss`

### Any additional context?

_No response_

richrines1:
coverage doesn't really accept many other arguments we'd care about, so extra command line args should be passed to pytest instead. This way the same arguments (including `--randomly-seed=`) can be passed to `check/pytest_.py` and `check/coverage_.py`

fixes #414

### DIFF
--- a/general-superstaq/general_superstaq/check/coverage_.py
+++ b/general-superstaq/general_superstaq/check/coverage_.py
@@ -28,7 +28,7 @@ def run(  # pylint: disable=missing-function-docstring
 
     parser.add_argument("--enable-socket", action="store_true", help="Force-enable socket.")
 
-    parsed_args, args_to_pass = parser.parse_known_intermixed_args(args)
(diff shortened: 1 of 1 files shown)


perlinm (approved):
lgtm (not sure what's up with the failing notebook check though, looks unrelated to me)

richrines1:
> looks unrelated to me

yep, combination of https://github.com/SupertechLabs/client-superstaq/pull/418 (now fixed) and https://github.com/SupertechLabs/client-superstaq/issues/396 (still at large)

PR #419, opened 2023-03-28
mock environ in `test_find_api_key` so it passes when SUPERSTAQ_API_KEY is set locally

richrines1:
previously the part of the test which was supposed to read the key from a file would fail because the env variable took precedence

### DIFF
--- a/general-superstaq/general_superstaq/superstaq_client_test.py
+++ b/general-superstaq/general_superstaq/superstaq_client_test.py
@@ -783,12 +783,13 @@ def test_find_api_key() -> None:
         assert gss.superstaq_client.find_api_key() == "tomyheart"
 
     # find key in a config file
-    with mock.patch("pathlib.Path.is_file", return_value=True):
-        with mock.patch("builtins.open", mock.mock_open(read_data="tomyheart")):
-            assert gss.superstaq_client.find_api_key() == "tomyheart"
+    with mock.patch.dict(os.environ, clear=True):
+        with mock.patch("pathlib.Path.is_file", return_value=True):
+            with mock.patch("builtins.open", mock.mock_open(read_data="tomyheart")):
+                assert gss.superstaq_client.find_api_key() == "tomyheart"
 
     # fail to find an API key :(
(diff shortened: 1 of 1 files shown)


dowusu-antwi (approved):
LGTM.

general-superstaq/general_superstaq/superstaq_client_test.py
```diff
@@ -783,12 +783,13 @@ def test_find_api_key() -> None:
         assert gss.superstaq_client.find_api_key() == "tomyheart"
 
     # find key in a config file
-    with mock.patch("pathlib.Path.is_file", return_value=True):
-        with mock.patch("builtins.open", mock.mock_open(read_data="tomyheart")):
-            assert gss.superstaq_client.find_api_key() == "tomyheart"
+    with mock.patch.dict(os.environ, clear=True):
```
dowusu-antwi, line 786: Actually, to fix the Windows pytest error something like this could work (to keep [other environment variables](https://github.com/SupertechLabs/client-superstaq/actions/runs/4539557022/jobs/7999541808?pr=419#step:5:83) set, e.g., `USERPROFILE`, `HOMEPATH`):
```suggestion
    mock_environ = os.environ.copy()
    del mock_environ["SUPERSTAQ_API_KEY"]
    with mock.patch.dict(os.environ, mock_environ):
```

perlinm:
Thanks for catching (and fixing) this @richrines1!

general-superstaq/general_superstaq/superstaq_client_test.py
richrines1: ah yep, didn't realize windows needed those! looks like just setting SUPERSTAQ_API_KEY to an empty string works

dowusu-antwi: Nice, yeah way cleaner.

dowusu-antwi:
FYI there's an issue with [IonQ's device](https://github.com/SupertechLabs/client-superstaq/actions/runs/4545151415/jobs/8012089900?pr=419#step:6:37) causing the notebook check failure:
image

PR #420, opened 2023-03-29
fix failing notebook test

### LINKED ISSUE #396: finance notebook test fails when `ionq_ion_qpu` isn't available
### What is happening?

[`cirq-superstaq/cirq_superstaq/examples/finance.ipynb`](https://github.com/SupertechLabs/superstaq-client/blob/main/cirq-superstaq/examples/finance.ipynb) submits a job to `ionq_ion_qpu`, and so will fail if run at a time when it isn't available:

screenshot

run
```
./check/pytest_.py --notebook cirq-superstaq/cirq_superstaq/examples/finance.ipynb
```
at a time when `ionq_ion_qpu` isn't available

### What should happen?

notebook checks shouldn't depend on the time their run

### Environment

- **superstaq-client version**: 0.3.20

### Any additional context?

_No response_

richrines1:
fixes #396


perlinm (approved):
Nice, clean and easy fix.

PR #428, opened 2023-04-03
Move examples in `css` and `qss` to docs

stephanielee9:
Summary of changes:

- Updated `index.rst` to include examples
- Created subdirectories in `optimizations` for each device
   - Each subdirectory contains an `.rst` with links to `css` and `qss` versions of examples
- Minor updates to heading levels within Jupyter notebooks (so the docs sidebar looks correct)


richrines1:
currently `check/pytest_.py` [excludes `docs/*`](https://github.com/SupertechLabs/client-superstaq/blob/f12aa7e5fb21a25a1e3600ff950eefad454f21e5/check/pytest_.py#L12) - we should change that to exclude only the individual files that don't currently pass so we keep testing the ones moved in this pr

vtomole (approved):
LGTM

Issue #387, opened 2023-02-15
better handling of illegal inputs


richrines1:
### What is happening?

a lot of endpoints will accept easily-checkable invalid inputs, resulting in inconsistent behavior/opaque errors. When it's easy to do so we should validate these inputs client-side before sending anything to the server
```python
service = css.Service()  # (or qss.SuperstaQProvider)

service.cq_compilt([])  # correctly returns `CompilerOutput([], None, None, None, None)`
service.qscout_compilt([])  # correctly returns `CompilerOutput([], None, None, [], None)`
service.ibmq_compile([])  # raises KeyError: 'pulses' (client-side)
service.aqt_compile([])  # internal server error
service.aqt_compile_eca(circuit, num_equivalent_circuits=0)  # raises ValueError (client-side)
service.get_counts(circuit, repetitions=-1, target="ibmq_qasm_simulator")  # internal server error
```

### What are the current alternatives?

we could return descriptive messages from the server for invalid inputs, instead of validating them client-side

PR #435, opened 2023-04-08
Added Input Validation 

bharat-thotakura:
Added methods for both qss and css to check for basis invalid inputs per #387


cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -165,6 +173,46 @@ def _resolve_target(self, target: Union[str, None]) -> str:
...
+    def _validate_get_counts(self, repetitions: int) -> None:
+        """Validates that the number of repetitions is non-negative.
+
+        Args:
+            repetitions: The number of times to run the circuit.
+
+        Raises:
+            InvalidInputError: If input repetitions is negative.
+        """
+        if repetitions < 0:
```
richrines1, line 211: is there any use case for `repetitions=0`? if not,
```suggestion
        if repetitions <= 0:
```
(and replace "non-negative" with "positive" in docstrings/error message)

richrines1: also might be worth checking that it's an integer (or at least can safely be cast as one)? e.g.
```python
if repetitions != int(repetitions):
    raise ....

File: cirq-superstaq/cirq_superstaq/service.py [hunk 2]
```diff
@@ -98,6 +98,14 @@ def counts_to_results(
     return result
 
 
+class InvalidInputError(Exception):
```
richrines1, line 101: i think it's ok to just use the builtin `ValueError` for these - it's (i think) typically what's used for bad input values in python

cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -165,6 +173,46 @@ def _resolve_target(self, target: Union[str, None]) -> str:
 
         return target
 
+    def _validate_cirq_circuit(self, circuits: Union[cirq.Circuit, Sequence[cirq.Circuit]]) -> None:
+        """Validates that the input is either a single cirq.Circuit or a list of cirq.Circuit
```
richrines1, line 177:
```suggestion
        """Validates that the input is either a single `cirq.Circuit` or a list of `cirq.Circuit`
```
(elsewhere as well - code snippets should be wrapped in \`backticks\`)

cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -165,6 +173,46 @@ def _resolve_target(self, target: Union[str, None]) -> str:
...
+        instances.
+
+        Args:
+            circuit: The circuit to run.
+
+        Raises:
+            InvalidInputError: If the 'circuits' input is not a cirq.Circuit or a list of
+            cirq.Circuit instances.
+            InvalidInputError: If the 'circuits' input is a list containing elements that
+            are not cirq.Circuit instances.
```
richrines1, line 187: nits: (1) i think the second case is covered by the first, (2) code should be wrapped in backticks, and (2) subsequent lines should be indented once more
```suggestion
            InvalidInputError: If the input is not a `cirq.Circuit` or a list of `cirq.Circuit`
                instances.
```

qiskit-superstaq/qiskit_superstaq/superstaq_provider.py
```diff
@@ -102,6 +110,42 @@ def backends(self) -> List[qss.SuperstaQBackend]:
...
         return backends
 
+    def _http_headers(self) -> Dict[str, str]:
+        return {
+            "Authorization": self.get_access_token(),
+            "Content-Type": "application/json",
+            "X-Client-Name": "qiskit-superstaq",
+            "X-Client-Version": gss.API_VERSION,
+        }
+
```
richrines1, line 120: this was factored out in #432 (looks like a weird merge thing)
```suggestion
```

cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -165,6 +173,46 @@ def _resolve_target(self, target: Union[str, None]) -> str:
 
         return target
 
+    def _validate_cirq_circuit(self, circuits: Union[cirq.Circuit, Sequence[cirq.Circuit]]) -> None:
```
richrines1, line 176: nit: imo these should be moved outside the class

richrines1 (commented):
nice! a few nits, and also we'll need to add checks the validation functions to get full coverage (you can use `pytest.raises` to check that they raise errors in the correct instances)

cirq-superstaq/cirq_superstaq/service.py
bharat-thotakura: Yeah, was originally going to have a class that could be updated with various error codes to standardize error handling for the different methods so that it would be easier to maintain and update error messages later on. So a remnant of that.

cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -98,6 +98,54 @@ def counts_to_results(
...
+    ):
+        raise ValueError(
+            "Invalid 'circuits' input. Must be a `cirq.Circuit` or a "
+            "sequence of `cirq.Circuit` instances."
+        )
+
+
+def _validate_get_counts(repetitions: Any) -> None:
+    """Validates that the number of repetitions is positive an
+    integer.
```
paaige, line 128: "and an integer."

cirq-superstaq/cirq_superstaq/service_test.py
```diff
@@ -64,6 +64,37 @@ def test_counts_to_results() -> None:
...
+        ValueError,
+        match="Invalid 'circuits' input. Must be a `cirq.Circuit` or a "
+        "sequence of `cirq.Circuit` instances.",
+    ):
+        css.service._validate_cirq_circuit([circuit, "circuit_invalid"])
+
+
+def test_validate_get_counts() -> None:
+
+    with pytest.raises(TypeError, match="Repetitions must be an integer."):
```
paaige, line 88: I would also add a test for a string that you'd expect to pass like "10" or float (100.0)

paaige (commented):
Added some nits + suggestions.

cirq-superstaq/cirq_superstaq/service_test.py
bharat-thotakura: Currently, It'll accept "10" and floats like 100.0 since they can be safely cast into integers. But I added some more invalid inputs in 278940a

paaige: Yeah, I was suggesting to test valid inputs

bharat-thotakura: Added your suggestion with commit dab3ed2

cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -98,6 +98,54 @@ def counts_to_results(
...
+
+    Raises:
+        TypeError: If input is not an integer.
+        ValueError: If input is negative.
+    """
+
+    if not (
+        (isinstance(repetitions, int))
+        or (isinstance(repetitions, float) and repetitions.is_integer())
+        or (isinstance(repetitions, str) and repetitions.isdigit())
```
richrines1, line 141: isdigit() accepts some unicode values which can't be converted to ints, e.g. `"₁₂₃".isdigit()` returns True but `int("₁₂₃")` fails. isdecimal() is safer:
```suggestion
        or (isinstance(repetitions, (bytes, str)) and repetitions.isdecimal())
```
(also added `bytes` for completeness)

cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -98,6 +98,54 @@ def counts_to_results(
...
+        repetitions: The number of times to run the circuit.
+
+    Raises:
+        TypeError: If input is not an integer.
+        ValueError: If input is negative.
+    """
+
+    if not (
+        (isinstance(repetitions, int))
+        or (isinstance(repetitions, float) and repetitions.is_integer())
```
richrines1, line 140: this will fail for e.g. `repetitions = np.int_(123)`. i think this works though:
```suggestion
        (isinstance(repetitions, SupportsInt) and int(repetitions) == repetitions)
```
(where `SupportsInt` is imported from typing)

cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -98,6 +98,54 @@ def counts_to_results(
...
+            and all(isinstance(circuit, cirq.Circuit) for circuit in circuits)
+        )
+    ):
+        raise ValueError(
+            "Invalid 'circuits' input. Must be a `cirq.Circuit` or a "
+            "sequence of `cirq.Circuit` instances."
+        )
+
+
+def _validate_get_counts(repetitions: Any) -> None:
```
richrines1, line 126: can we give this a more generic name
```suggestion
def _validate_num_repetitions(repetitions: object) -> None:
```
and use it to validate `num_eca_circuits` as well? (if so we should also change the error messages to something more generic, e.g. `TypeError(f"{repetitions} cannot be safely cast as an integer.")`)

cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -98,6 +98,54 @@ def counts_to_results(
     return result
 
 
+def _validate_cirq_circuit(circuits: Any) -> None:
```
richrines1, line 101: nits: plural name, and avoid using `Any` (which is made unnecessary by the `isinstance` checks)
```suggestion
def _validate_cirq_circuits(circuits: object) -> None:
```

cirq-superstaq/cirq_superstaq/service_test.py
```diff
@@ -64,6 +64,44 @@ def test_counts_to_results() -> None:
...
+        match="Invalid 'circuits' input. Must be a `cirq.Circuit` or a "
+        "sequence of `cirq.Circuit` instances.",
+    ):
+        css.service._validate_cirq_circuit([circuit, "circuit_invalid"])
+
+
+def test_validate_get_counts() -> None:
+
+    invalid_inputs = [None, "reps", 1.5, "1.0", {1}, [1, 2, 3]]
+    valid_inputs = [1, 10, "10", 10.0]
```
richrines1, line 89:
```suggestion
    valid_inputs = [1, 10, "10", 10.0, np.int16(10)]
```

richrines1 (commented):
(added few more comments in `css`, the same apply to `qss` as well)

bharat-thotakura:
> (added few more comments in `css`, the same apply to `qss` as well)

Thanks for the comments!

cirq-superstaq/cirq_superstaq/service.py
bharat-thotakura: Ran into coverage issues that protocols could not be used with `isinstance()` and `SupportsInt` so opted for `hasattr()` instead in 5d342e0

cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -98,6 +98,59 @@ def counts_to_results(
...
+
+    if not (
+        hasattr(integer_param, "__int__")
+        and int(integer_param) == integer_param
+        or (
+            isinstance(integer_param, (bytes, str))
+            and (
+                str(integer_param, "utf-8") if isinstance(integer_param, bytes) else integer_param
+            ).isdecimal()
+        )
```
richrines1, line 146: ah sorry didn't realize bytes.isdecimal doesn't work in python 3.7. let's just go back to
```suggestion
        or (isinstance(integer_param, str) and integer_param.isdecimal())
```

cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -98,6 +98,59 @@ def counts_to_results(
...
+        integer_param: An input parameter.
+
+    Raises:
+        TypeError: If input is not an integer.
+        ValueError: If input is negative.
+    """
+
+    if not (
+        hasattr(integer_param, "__int__")
+        and int(integer_param) == integer_param
```
richrines1, line 140: nit (mostly because i can never remember and/or operator precedence):
```suggestion
        (hasattr(integer_param, "__int__") and int(integer_param) == integer_param)
```

cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -98,6 +98,59 @@ def counts_to_results(
...
+def _validate_cirq_circuits(circuits: object) -> None:
+    """Validates that the input is either a single `cirq.Circuit` or a list of `cirq.Circuit`
+    instances.
+
+    Args:
+        circuits: The circuit to run.
+
+    Raises:
+        ValueError: If the input is not a `cirq.Circuit` or a list of
+        `cirq.Circuit` instances.
```
richrines1, line 110: nit
```suggestion
        ValueError: If the input is not a `cirq.Circuit` or a list of `cirq.Circuit` instances.
```

qiskit-superstaq/qiskit_superstaq/superstaq_provider.py
```diff
@@ -24,6 +24,57 @@
...
+    ):
+        raise ValueError(
+            "Invalid 'circuits' input. Must be a `qiskit.QuantumCircuit` or a "
+            "sequence of `qiskit.QuantumCircuit` instances."
+        )
+
+
+def _validate_integer_param(integer_param: object) -> None:
+    """Validates that an input parameter is positive
+    and an integer.
```
richrines1, line 53:
```suggestion
    """Validates that an input parameter is positive and an integer.
```

qiskit-superstaq/qiskit_superstaq/superstaq_provider.py
```diff
@@ -24,6 +24,57 @@
...
+def _validate_qiskit_circuits(circuits: object) -> None:
+    """Validates that the input is either a single `qiskit.QuantumCircuit` or a list of
+    `qiskit.QuantumCircuit` instances.
+
+    Args:
+        circuits: The circuit to run.
+
+    Raises:
+        ValueError: If the input is not a `qiskit.QuantumCircuit` or
+        a list of `qiskit.QuantumCircuit` instances.
```
richrines1, line 36:
```suggestion
        ValueError: If the input is not a `qiskit.QuantumCircuit` or a list of
        `qiskit.QuantumCircuit` instances.
```

cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -98,6 +98,59 @@ def counts_to_results(
     return result
 
 
+def _validate_cirq_circuits(circuits: object) -> None:
+    """Validates that the input is either a single `cirq.Circuit` or a list of `cirq.Circuit`
+    instances.
+
+    Args:
+        circuits: The circuit to run.
```
richrines1, line 106:
```suggestion
        circuits: The circuit(s) to run.
```

qiskit-superstaq/qiskit_superstaq/superstaq_provider.py
```diff
@@ -24,6 +24,57 @@
 import qiskit_superstaq as qss
 
 
+def _validate_qiskit_circuits(circuits: object) -> None:
+    """Validates that the input is either a single `qiskit.QuantumCircuit` or a list of
+    `qiskit.QuantumCircuit` instances.
+
+    Args:
+        circuits: The circuit to run.
```
richrines1, line 32:
```suggestion
        circuits: The circuit(s) to run.
```

richrines1 (commented):
a few more nits but otherwise looks good!

PR #438, opened 2023-04-13
Daily test submit to aws and ibm

vtomole:


vtomole:
Passes https://github.com/SupertechLabs/client-superstaq/actions/runs/4684554678

paaige (approved):
lgtm

vtomole:
Nit after approval: submitted to the back-ends via a loop instead of individually.

PR #446, opened 2023-04-17
Add options to cq_compile

saeubank:
Address https://github.com/SupertechLabs/client-superstaq/issues/404

### DIFF
--- a/cirq-superstaq/cirq_superstaq/service.py
+++ b/cirq-superstaq/cirq_superstaq/service.py
@@ -544,13 +544,14 @@ class Service(finance.Finance, logistics.Logistics, user_config.UserConfig):
         serialized_circuits = css.serialization.serialize_circuits(circuits)
         circuits_is_list = not isinstance(circuits, cirq.Circuit)
 
-        json_dict = self._client.cq_compile(
-            {
-                "cirq_circuits": serialized_circuits,
-                "target": target,
-                "options": cirq.to_json(options),
-            }
-        )
+        request_json = {
+            "cirq_circuits": serialized_circuits,
+            "target": target,
+        }
+        if options:
+            request_json["options"] = cirq.to_json(options)
+
+        json_dict = self._client.cq_compile(request_json)
 
         return css.compiler_output.read_json_only_circuits(json_dict, circuits_is_list)
 
@@ -571,13 +572,14 @@ class Service(finance.Finance, logistics.Logistics, user_config.UserConfig):
         if not target.startswith("ibmq_"):
             raise ValueError(f"{target} is not an IBMQ target")
 
-        json_dict = self._client.ibmq_compile(
-            {
-                "cirq_circuits": serialized_circuits,
-                "target": target,
-                "options": cirq.to_json(options),
-            }
-        )
+        request_json = {
+            "cirq_circuits": serialized_circuits,
+            "target": target,
+        }
+        if options:
+            request_json["options"] = cirq.to_json(options)
+
+        json_dict = self._client.ibmq_compile(request_json)
 
         return css.compiler_output.read_json_ibmq(json_dict, circuits_is_list)
 
--- a/cirq-superstaq/cirq_superstaq/service.py
+++ b/cirq-superstaq/cirq_superstaq/service.py
@@ -461,7 +461,7 @@ class Service(finance.Finance, logistics.Logistics, user_config.UserConfig):
         options_dict = {}
         if options:
             options_dict.update(options)
-            
+
         if num_equivalent_circuits is not None:
             options_dict["num_eca_circuits"] = num_equivalent_circuits
         if random_seed is not None:
@@ -507,13 +507,14 @@ class Service(finance.Finance, logistics.Logistics, user_config.UserConfig):
         serialized_circuits = css.serialization.serialize_circuits(circuits)
         circuits_is_list = not isinstance(circuits, cirq.Circuit)
 
-        options = {} if options is None else options
-
         options_dict = {
             "mirror_swaps": mirror_swaps,
             "base_entangling_gate": base_entangling_gate,
-            **options,
         }
+
+        if options:
+            options_dict.update(options)
+
         json_dict = self._client.qscout_compile(
             {
                 "cirq_circuits": serialized_circuits,
--- a/cirq-superstaq/cirq_superstaq/service.py
+++ b/cirq-superstaq/cirq_superstaq/service.py
@@ -458,8 +458,10 @@ class Service(finance.Finance, logistics.Logistics, user_config.UserConfig):
         }
 
         options_dict: Dict[str, Union[float, Dict[str, Union[cirq.Gate, cirq.Operation, None]]]]
-        options_dict = {} if options is None else options
-
+        options_dict = {}
+        if options:
+            options_dict.update(options)
+            
         if num_equivalent_circuits is not None:
             options_dict["num_eca_circuits"] = num_equivalent_circuits
         if random_seed is not None:
@@ -542,7 +544,11 @@ class Service(finance.Finance, logistics.Logistics, user_config.UserConfig):
         circuits_is_list = not isinstance(circuits, cirq.Circuit)
 
         json_dict = self._client.cq_compile(
-            {"cirq_circuits": serialized_circuits, "target": target, "options": options}
+            {
+                "cirq_circuits": serialized_circuits,
+                "target": target,
+                "options": cirq.to_json(options),
+            }
         )
 
         return css.compiler_output.read_json_only_circuits(json_dict, circuits_is_list)
@@ -565,7 +571,11 @@ class Service(finance.Finance, logistics.Logistics, user_config.UserConfig):
             raise ValueError(f"{target} is not an IBMQ target")
 
         json_dict = self._client.ibmq_compile(
-            {"cirq_circuits": serialized_circuits, "target": target, "options": options}
+            {
+                "cirq_circuits": serialized_circuits,
+                "target": target,
+                "options": cirq.to_json(options),
+            }
         )
 
         return css.compiler_output.read_json_ibmq(json_dict, circuits_is_list)
--- a/cirq-superstaq/cirq_superstaq/service.py
+++ b/cirq-superstaq/cirq_superstaq/service.py
@@ -540,7 +540,6 @@ class Service(finance.Finance, logistics.Logistics, user_config.UserConfig):
         """
         serialized_circuits = css.serialization.serialize_circuits(circuits)
         circuits_is_list = not isinstance(circuits, cirq.Circuit)
-        options = {} if options is None else options
 
         json_dict = self._client.cq_compile(
             {"cirq_circuits": serialized_circuits, "target": target, "options": options}
@@ -565,7 +564,6 @@ class Service(finance.Finance, logistics.Logistics, user_config.UserConfig):
         if not target.startswith("ibmq_"):
             raise ValueError(f"{target} is not an IBMQ target")
 
-        options = {} if options is None else options
         json_dict = self._client.ibmq_compile(
             {"cirq_circuits": serialized_circuits, "target": target, "options": options}
         )
--- a/cirq-superstaq/cirq_superstaq/service.py
+++ b/cirq-superstaq/cirq_superstaq/service.py
@@ -574,7 +574,7 @@ class Service(finance.Finance, logistics.Logistics, user_config.UserConfig):
         request_json = {
             "cirq_circuits": serialized_circuits,
             "target": target,
-            "options": cirq.to_json({**kwargs}),
+            "options": cirq.to_json(kwargs),
         }
 
         json_dict = self._client.ibmq_compile(request_json)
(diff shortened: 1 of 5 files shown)


general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -272,7 +272,7 @@ def qscout_compile(
         """Makes a POST request to SuperstaQ API to compile a list of circuits for QSCOUT."""
         return self.post_request("/qscout_compile", json_dict)
 
-    def cq_compile(self, json_dict: Dict[str, Union[str, List[str]]]) -> Dict[str, str]:
+    def cq_compile(self, json_dict: Dict[str, Any]) -> Dict[str, str]:
```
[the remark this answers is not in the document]
saeubank: We loosen the types here since `options` is type `Dict[str, Any]` so this `json_dict` would need to be the same type. `post_request`'s `json_dict` is also type `Dict[str, Any]` so I don't think this would cause any issues. (`post_request` is called by this function)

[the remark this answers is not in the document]
richrines1: incidentally i'm not sure the Union is even necessary here; it looks like just `Dict[str, str]` should be fine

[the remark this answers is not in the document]
saeubank: Would it not be possible for the options to have nested objects? In that case we would need something like `Json = dict[str, 'Json'] | list['Json'] | str | int | float | bool | None` for strict typing (instead of using `Any`), but that would only be possible in python 3.11+ (I think)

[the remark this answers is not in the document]
saeubank: But also I just noticed the other compile methods have their options as arguments in the function, I can poke around to try to find what those are for `cq_compile` I assume that probably lives somewhere in the server code.

(same for `ibmq_compile`)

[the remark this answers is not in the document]
richrines1: > Would it not be possible for the options to have nested objects? In that case we would need something like `Json = dict[str, 'Json'] | list['Json'] | str | int | float | bool | None` for strict typing (instead of using `Any`), but that would only be possible in python 3.11+ (I think)

recursive type definitions are allowed prior to 3.11 i think (you just have to use the older syntax, e.g. `Union[Dict[str, "Json"], ...]` etc). i've actually attempted to do this in the past, but it gets a little complicated because (1) you have to deal with variance correctly everywhere (e.g. by using `Sequence`/`Mapping` instead of `List`/`Dict`), and (2) you have to handle cirq types (which can often be serialized with `cirq.to_json`, which we sometimes use in `css`). And overall we don't lose too much using `Dict[str, Any]`

either way it shouldn't matter for this function in particular - basically `css.Service.cq_compile()` can be annotated w/ `options: Dict[str, Any]`, but the dictionary it passes to this function should already have options serialized so will have type `Dict[str, str]`

(for some context: the options dictionary as implemented is not really intended to be permanent - it's a way make interfaces more flexible while the API is still very much in flux, but as things get more nailed down we'll probably want to do something more rigid)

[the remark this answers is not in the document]
richrines1: > But also I just noticed the other compile methods have their options as arguments in the function, I can poke around to try to find what those are for `cq_compile` I assume that probably lives somewhere in the server code.

i think doing both is ideal - common options can be made into arguments to make them more accessible/documented to users, but it's also useful to be able to augment this w/ arbitrary options (e.g. for temporary/trial parameters or options that only matter to us internally)

in this case i think the options i'd be most tempted to add as arguments are "grid_shape" and "control_radius"

saeubank:
I am not sure if what to do with these 2 pipeline failures. Could someone take a look and point me to what is causing these issues?

cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -544,7 +565,7 @@ def ibmq_compile(
             raise ValueError(f"{target} is not an IBMQ target")
 
         json_dict = self._client.ibmq_compile(
-            {"cirq_circuits": serialized_circuits, "target": target}
+            {"cirq_circuits": serialized_circuits, "target": target, "options": options}
```
richrines1, line 568:
```suggestion
            {"cirq_circuits": serialized_circuits, "target": target, "options": cirq.to_json(options)}
```

cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -509,28 +523,35 @@ def qscout_compile(  # pylint: disable=missing-param-doc,missing-raises-doc
...
+            options: dictionary of desired cq_compile options.
         Returns:
             object whose .circuit(s) attribute is an optimized cirq Circuit(s)
         """
         serialized_circuits = css.serialization.serialize_circuits(circuits)
         circuits_is_list = not isinstance(circuits, cirq.Circuit)
 
         json_dict = self._client.cq_compile(
-            {"cirq_circuits": serialized_circuits, "target": target}
+            {"cirq_circuits": serialized_circuits, "target": target, "options": options}
```
richrines1, line 545:
```suggestion
            {"cirq_circuits": serialized_circuits, "target": target, "options": cirq.to_json(options)}
```

cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -451,7 +458,7 @@ def _aqt_compile(
         }
 
         options_dict: Dict[str, Union[float, Dict[str, Union[cirq.Gate, cirq.Operation, None]]]]
-        options_dict = {}
+        options_dict = {} if options is None else options
```
richrines1, line 461: let's do something like this:
```suggestion
        options_dict = {}
        if options:
            options_dict.update(options)
```
that way we don't mutate input dictionaries passed by the user

general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -262,21 +262,19 @@ def resource_estimate(self, json_dict: Dict[str, str]) -> Dict[str, List[Dict[st
         """
         return self.post_request("/resource_estimate", json_dict)
 
-    def aqt_compile(self, json_dict: Dict[str, Union[int, str, List[str]]]) -> Dict[str, str]:
+    def aqt_compile(self, json_dict: Dict[str, Any]) -> Dict[str, str]:
```
richrines1, line 265: this shouldn't be necessary - the options dictionary is serialized prior to this so the annotation here shouldn't need to change
```suggestion
    def aqt_compile(self, json_dict: Dict[str, Union[int, str, List[str]]]) -> Dict[str, str]:
```
(ditto for other annotations in this module)

richrines1 (commented):
> I am not sure what to do with these 2 pipeline failures. Could someone take a look and point me to what is causing these issues?

i believe the input validation error is related to [this](https://github.com/SupertechLabs/client-superstaq/pull/446#discussion_r1171668341) - the server expects `json_dict["options"]` to be a json-deserializable string. it should be fixed by the suggestions below

general-superstaq/general_superstaq/superstaq_client.py
saeubank: Ah thanks, I missed that.

cirq-superstaq/cirq_superstaq/service.py
saeubank: Ah thanks, I missed that.

cirq-superstaq/cirq_superstaq/service.py
saeubank: Good point, updated.

general-superstaq/general_superstaq/superstaq_client.py
saeubank: since `request_json` in `service.py`'s `_aqt_compile()` is type `Dict[str,str]`

richrines1:
ok everything here looks good to me. the same set of changes should also be made in `qss.SuperstaQProvider` (but using `json.dumps()` instead of `cirq.to_json()` to serialize the dictionary)

cirq-superstaq/cirq_superstaq/service_test.py
```diff
@@ -503,7 +505,7 @@ def test_service_ibmq_compile(mock_ibmq_compile: mock.MagicMock) -> None:
         "final_logical_to_physicals": cirq.to_json([list(final_logical_to_physical.items())]),
     }
 
-    assert service.ibmq_compile(circuit).circuit == circuit
+    assert service.ibmq_compile(circuit, options={"test": "yes"}).circuit == circuit
```
paaige, line 508: off topic but we should definitely have a function validating that included options are supported.

saeubank: Agreed, this should definitely happen. I am not sure if we should be talking about the private repo's code in this public repo though. (not sure what the policy is here)

saeubank:
> ok everything here looks good to me. the same set of changes should also be made in `qss.SuperstaQProvider` (but using `json.dumps()` instead of `cirq.to_json()` to serialize the dictionary)

Out of curiosity should we be using some qiskit version of `json.dumps()` similar to how we are using cirq's `cirq.to_json()`?

cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -366,6 +367,7 @@ def aqt_compile(
                 implies css.CZ3 for all "CZ3/*" calibrations except "CZ3/C5T4", which will be
                 mapped to a css.CZ3_INV on qutrits (4, 5). Setting any calibration to None will
                 disable that calibration.
+            options: dictionary of desired aqt_compile options.
```
richrines1, line 370: nit: argument descriptions should be capitalized (though admittedly this isn't consistent across the repo yet)
```suggestion
            options: Dictionary of desired aqt_compile options.
```
(ditto elsewhere)

richrines1 (commented):
docstring nit then this seems good to me :)

one final thought before we merge this though - wdyt about using `**kwargs` instead of `options=`? e.g.
```python
    def qscout_compile(
        self,
        circuits: Union[cirq.Circuit, List[cirq.Circuit]],
        mirror_swaps: bool = True,
        base_entangling_gate: str = "xx",
        target: str = "sandia_qscout_qpu",
        **kwargs: Any,
    ) -> css.compiler_output.CompilerOutput:
        ....
        options_dict = {
            "mirror_swaps": mirror_swaps,
            "base_entangling_gate": base_entangling_gate,
            **kwargs,
        }
```
it simplifies the logic a bit in that we never have to check if options is None, and also provides a tiny bit of type validation (kwargs always a dictionary with strings for keys). Not sure if it makes the interface is better or worse though:
```python
service.cq_compile(circuits, options={"grid_shape": (3, 3)})  # options=

service.cq_compile(circuits, grid_shape=(3, 3))  # kwargs
```
the latter is a little bit cleaner (though it relies on undocumented keyword arguments), while the `options=` version is maybe nicer if users will frequently define a static options dictionary once and then use it multiple times

(@vtomole curious for your thoughts here as well)

richrines1:
> Out of curiosity should we be using some qiskit version of `json.dumps()` similar to how we are using cirq's `cirq.to_json()`?

unfortunately there isn't really a qiskit equivalent afaik (they do provide QPY for serializing circuits, but it doesn't work for other types). eventually we should probably implement our own or find a third-party package though - if nothing else to handle complex values and numpy arrays correctly

cirq-superstaq/cirq_superstaq/service.py
saeubank: Using `kwargs` means code that calls this function would not need to be updated when more options are added to the function signature, which seems like a plus. 

The only downside I can think of is that it might be confusing, but that can be kinda fixed through writing what `kwargs` does in the docstring.  

I like that suggestion, I'll update the code to use that unless there are any objections.

cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -475,66 +474,95 @@ def _aqt_compile(
...
         serialized_circuits = css.serialization.serialize_circuits(circuits)
         circuits_is_list = not isinstance(circuits, cirq.Circuit)
 
-        json_dict = self._client.cq_compile(
-            {"cirq_circuits": serialized_circuits, "target": target}
-        )
+        request_json = {
+            "cirq_circuits": serialized_circuits,
+            "target": target,
+            "options": cirq.to_json({**kwargs}),
```
richrines1, line 543: nit
```suggestion
            "options": cirq.to_json(kwargs),
```

cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -543,9 +571,13 @@ def ibmq_compile(
         if not target.startswith("ibmq_"):
             raise ValueError(f"{target} is not an IBMQ target")
 
-        json_dict = self._client.ibmq_compile(
-            {"cirq_circuits": serialized_circuits, "target": target}
-        )
+        request_json = {
+            "cirq_circuits": serialized_circuits,
+            "target": target,
+            "options": cirq.to_json({**kwargs}),
```
richrines1, line 577:
```suggestion
            "options": cirq.to_json(kwargs),
```

qiskit-superstaq/qiskit_superstaq/superstaq_provider.py
```diff
@@ -223,17 +236,31 @@ def ibmq_compile(
...
 
         serialized_circuits = qss.serialization.serialize_circuits(circuits)
 
-        json_dict = self._client.ibmq_compile(
-            {"qiskit_circuits": serialized_circuits, "target": target}
-        )
+        request_json = {
+            "qiskit_circuits": serialized_circuits,
+            "target": target,
+            "options": json.dumps({**kwargs}),
```
richrines1, line 260: nit
```suggestion
            "options": json.dumps(kwargs),
```

richrines1 (approved):
a few more nits then lgtm!

thanks :)

richrines1 (approved):

vtomole (approved):
LGTM

PR #452, opened 2023-05-03
Temporarily ignore failing notebooks

vtomole:
The deploy script fails because of these notebooks because there is a mismatch between the qiskit versions in the client and server. Will undo once we deploy the client.


pgoiporia (approved):
lgtm

PR #457, opened 2023-05-05
Missing compiled metadata

### LINKED ISSUE #453: Include metadata for compiled qiskit circuit
### What is happening?

When I use cq_compile on a qiskit QuantumCircuit that has metadata stored in circuit.metadata, the compiled circuit does not carry over that metadata.

I work around this inconvenience by keeping the original circuit around and accessing its metadata later, or I just set the compiled circuit's metadata to be equal to that of the original.

### Any additional context?

image

bharat-thotakura:
Fixes #453


dowusu-antwi (changes requested):
Nice. We should also do this for `ibmq_compile`, `aqt_compile`, etc.:
image

dowusu-antwi (commented):
Additionally, error message is fixed after updating branch / merging.

bharat-thotakura:
> Nice. We should also do this for `ibmq_compile`, `aqt_compile`, etc.: image

Should work for those too now with 113d45f

dowusu-antwi (approved):
Great, lgtm!

PR #466, opened 2023-05-18
consolidate cq_compile and ibmq_compile endpoints

richrines1:
(merge after depoloying server-side changes)


bharat-thotakura (approved):
LGTM

richrines1:
@dowusu-antwi this has a few implications for #473:
* cq_compile and ibmq_compile should now both call the generic `_client.compile()` instead of `_client.(cq|ibmq)_compile()`
* `_client.compile()` \*should\* also support any valid target other than qscout and aqt (e.g. ionq_ion_qpu)

PR #467, opened 2023-05-19
update docstrings and defaults for aqt_compile and qscout_compile

richrines1:
in response to bug reports from qscout users:
* changes default `mirror_swaps` value to False
* updates docstring of qscout_compile and aqt_compile (previously the qscout_compile contained an accidental reference to AQT)


vtomole (approved):
LGTM

PR #469, opened 2023-05-23
don't ignore relative phase of sub_gate in `QubitSubspaceGate`

richrines1:
currently `equal_up_to_global_phase` checks if the `sub_gate` s of two `QubitSubspaceGate`s are equal \*up to global phase\*, but this isn't correct because the global phase of the sub gates because physical when expanded to higher dimensions

this pr changes `QubitSubspaceGate._equal_up_to_global_phase_` to instead use the absolute comparison of the two sub gates

### DIFF
--- a/cirq-superstaq/cirq_superstaq/ops/qudit_gates.py
+++ b/cirq-superstaq/cirq_superstaq/ops/qudit_gates.py
@@ -422,11 +422,9 @@ class QubitSubspaceGate(cirq.Gate):  # pylint: disable=missing-class-docstring
         if other.subspaces != self.subspaces:
             return False
 
-        return cirq.equal_up_to_global_phase(
-            self.sub_gate, other.sub_gate, atol=atol
-        ) or cirq.equal_up_to_global_phase(
-            other.sub_gate, self.sub_gate, atol=atol
-        )  # Test both orders as a workaround for https://github.com/quantumlib/Cirq/issues/5980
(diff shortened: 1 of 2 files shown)


pgoiporia (approved):
I don't quite understand the physics, but the code lgtm lol

richrines1:
> I don't quite understand the physics

basically `QubitSubspaceGate` just embeds a qubit gate in a given subspace of higher-dimensional qudit operation, while performing an identity on the new energy levels. Which means for example the gate
```python
css.QubitSubspaceGate(cirq.X, (3,))
```
has the unitary
```
[[0.+0.j 1.+0.j 0.+0.j]
 [1.+0.j 0.+0.j 0.+0.j]
 [0.+0.j 0.+0.j 1.+0.j]],
```
while
```python
css.QubitSubspaceGate(cirq.rx(np.pi), (3,))
```
has the unitary
```
[[0.+0.j 0.-1.j 0.+0.j]
 [0.-1.j 0.+0.j 0.+0.j]
 [0.+0.j 0.+0.j 1.+0.j]].
```
So the different global phases of `cirq.X` and `cirq.rx(pi)` become different physical phases of the (0, 1) and (1, 0) terms relative to the (2, 2) term

pgoiporia:
> 

Ah I see. Makes sense. Thanks!

richrines1:
(will merge after server-side changes have been merged)

PR #471, opened 2023-05-25
allow ibmq_compile to succeed if the server doesn't send back pulses

richrines1:
allows users to get compiled circuits for open qasm/non-pulse-enabled backends


bharat-thotakura (approved):
lgtm

PR #473, opened 2023-05-25
Move compile endpoints to backend for `qss`

dowusu-antwi:
Fixes #206.

Adds `compile` method to backend object in `qss` for `qiskit` parity, but also maintains `provider_compile` endpoint (as a thin wrapper around `backend.compile`) for `css` parity. (Essentially amounts to moving code / tests.)

PR #476, opened 2023-06-01
add a generic compilation endpoint to `css.Service`

richrines1:
allows compilation to arbitrary targets (including aws/honeywell/etc)

(qss parity will come with #473)


paaige (approved):
lgtm

bharat-thotakura (approved):
LGTM

PR #479, opened 2023-06-02
Infra for load testing Superstaq

vtomole:


dev_tools/locustfile.py
```diff
@@ -0,0 +1,40 @@
...
+
+import logging
+from typing import TYPE_CHECKING
+
+import cirq_superstaq
+import requests
+from locust import HttpUser, events, task
+
+if TYPE_CHECKING:
+    from locust.env import Environment
```
richrines1, line 18: nit: why not just import
```suggestion
import locust.env
```
and use `locust.HttpUser` / `locust.events` / `locust.task` / `locust.env.Environment` below?

.github/workflows/daily-integration-check.yml
```diff
@@ -34,3 +34,24 @@ jobs:
         with:
           filename: .github/integration-check-failure-issue.md
           update_existing: true
+
+  load-testing-check:
```
richrines1, line 38: should we have this create a github issue as well?

vtomole: >  should we have this create a github issue as well?

Not yet because we don't even know what to do if it fails. For example, at the moment we have no idea why the performance is fluctuating. At the moment i view this check as a "monitor" more than a "check".

richrines1: ah ok makes sense

dev_tools/locustfile.py
vtomole: Done

.github/workflows/daily-integration-check.yml
vtomole: I see us integrating this with https://asv.readthedocs.io/en/stable/ in the far future.

richrines1: another weird question: should we set this to run at a slightly different time than the other integration tests? otherwise they'll impact the server load

vtomole: People will still be making requests to Superstaq anyway when we are load testing no? We could set a window that says "no one can make a request to Superstaq except the load tester for *this period of time" but i think that would be overkill.

.github/workflows/daily-integration-check.yml
```diff
@@ -34,3 +34,24 @@ jobs:
...
+    name: Load testing check
+    env:
+      SUPERSTAQ_API_KEY: ${{ secrets.SUPERSTAQ_API_KEY }}
+      TEST_USER_IBMQ_TOKEN: ${{ secrets.TEST_USER_IBMQ_TOKEN }}
```
dowusu-antwi, line 42: nit: can we move the `env` map outside the `jobs` section since both jobs use it? not sure if yaml syntax allows this...

dev_tools/locustfile.py
```diff
@@ -0,0 +1,36 @@
...
+"locust --config=locust.conf" in the "dev_tools" directory. Then
+navigate to http://0.0.0.0:8089/, insert the number of users and spawn rate,
+then click the "Start swarming" button.
+"""
+
+from __future__ import annotations
+
+import logging
+
+import cirq_superstaq
```
singular-value, line 12:
```suggestion
import cirq_superstaq as css
```
^ consistency with the rest of codebase (per vote last summer)

dev_tools/locustfile.py
```diff
@@ -0,0 +1,36 @@
...
+@locust.events.quitting.add_listener
+def _(environment: locust.env.Environment) -> None:
+    if environment.stats.total.fail_ratio > 0.01:
+        logging.error("Test failed due to failure ratio > 1%")
+        environment.process_exit_code = 1
+    else:
+        environment.process_exit_code = 0
+
+
+SERVICE = cirq_superstaq.Service()
```
singular-value, line 26:
```suggestion
SERVICE = css.Service()
```

singular-value: also... `service`? Again for consistency with our examples. Also, I don't think of service as a constant (which would be all caps)

dev_tools/locustfile.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class QuickstartUser(locust.HttpUser):
+    """Simulates a user during load testing"""
+
+    @locust.task
+    def get_targets(self) -> None:
+        """Load tests the get_backends endpoint"""
```
singular-value, line 34:
```suggestion
        """Load tests the get_targets endpoint"""
```

singular-value (changes requested):

dev_tools/locustfile.py
singular-value: Also fixed https://github.com/SupertechLabs/client-superstaq/pull/518

dev_tools/locustfile.py
vtomole: All caps cause it's a global var in this file.

vtomole:
@singular-value @dowusu-antwi @richrines1 @pgoiporia  I've removed the daily check and left the code for load testing in `dev_tools`. There is a larger "track performance of Superstaq" project in the future that should address the frequent load testing but now is not the time.

general-superstaq/dev-requirements.txt
```diff
@@ -3,6 +3,7 @@ black[jupyter]~=22.3.0
 flake8>=3.8.4
 flake8-type-checking>=2.1.0
 isort[colors]>=5.10.1
+locust>=2.15.1
```
richrines1, line 6: imo this shouldn't be part of gss's dev-requirements - we can just update the instructions at the top of locustfile.py to include `pip install locust`

vtomole: > imo this shouldn't be part of gss's dev-requirements

Reason? This is a "dev" requirement after all.

richrines1: not a big deal, but i guess i don't really think of it as a "requirement" in that it's not needed for any tests and most devs will never use it

also gss's dev-requirements are used used when installing every other superstaq package with [dev], so it adds a new dependency to five different packages for something that usually isn't really necessary for any of them

vtomole: Let's leave this until it starts causing problems in downstream packages.

vtomole: Done after IRL discussion: Locust has [a lot](https://github.com/locustio/locust/blob/17107e0fd45d84fcb0f3cfdab5a9728f45dae73a/pyproject.toml#L12) of dependencies.

richrines1 (approved):
lgtm!

PR #481, opened 2023-06-02
Add ibmq_token encryption to client

pgoiporia:


general-superstaq/general_superstaq/_init_vars.py
```diff
@@ -1,2 +1,12 @@
 API_URL = "https://superstaq.super.tech"
 API_VERSION = "v0.2.0"
+TOKEN_PUBLIC_KEY_E = 65537
+TOKEN_PUBLIC_KEY_N = "170085480843227966251730824714595515546847560229328345394016543390109446338"
```
vtomole, line 4: Why is it broken up like this? if it's because we need to please a linter i think this is the exception where we tell the linter to ignore this line.

richrines1: we could also use e.g.
```python
TOKEN_PUBLIC_KEY_N = int(
    "123456789123456789123456789123456789123456789123456789123456789123456789123456789"
    "123456789123456789123456789123456789123456789123456789123456789123456789123456789"
    "123456789123456789123456789123456789123456789123456789123456789123456789123456789"
    ....
)
so that it's both a single command and actually an `int`

File: general-superstaq/general_superstaq/user_config.py [hunk 2]
```diff
@@ -96,6 +98,9 @@ def ibmq_set_token(self, token: str) -> str:
         Returns:
             String containing status of update (whether or not it failed).
         """
+        public_key = rsa.key.PublicKey(int(gss.TOKEN_PUBLIC_KEY_N), gss.TOKEN_PUBLIC_KEY_E)
+        token = rsa.encrypt(token.encode(), public_key)
+        token = str(rsa.transform.bytes2int(token))
```
richrines1, line 103: why not just use `_bytes_to_str` for this? (and `_str_to_bytes` server-side)

vtomole: > why not just use _bytes_to_str for this? (and _str_to_bytes server-side)

We didn't know about this.

richrines1: > We didn't know about this.

haha yeah i'd totally forgotten about it too until like an hour ago when i ran into it working on something unrelated

richrines1: one other thought - would it make sense to move this functionality to `gss.serialization` as well? e.g.
```python
def encrypt(val: str) -> str:
    public_key = ...
    encrypted_val = rsa.encrypt(val.encode(), public_key)
    return bytes_to_str(encrypted_val)
```
so we could just call it here and also use it elsewhere if we want to start encrypting other tokens/values/etc

vtomole: I like that idea. Let's keep in our pockets for when [we will start encrypting other tokens/values ](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming))

pgoiporia:
@richrines1 @vtomole Do you have a better test in mind. Not sure how to properly test the encrypt function.

richrines1:
> Not sure how to properly test the encrypt function.

not perfect, but wdyt about something like this?
```python
def test_encrypt() -> None:
    public, private = rsa.newkeys(512)
    with mock.patch("gss.PUBLIC_KEY_E", public.e), mock.patch("gss.PUBLIC_KEY_N", public.n):
        val = "abc123"
        encrypted_val = encrypt(val)
        assert isinstance(encrypted_val, str)
        assert rsa.decrypt(rsa.str_to_bytes(encrypted_val), private).decode() == val
```

vtomole:
You don't need to test it cause we agreed that we won't add an `encrypt` function in this PR.

pgoiporia:
> You don't need to test it cause we agreed that we won't add an `encrypt` function in this PR.

Oh ok I see. I'll revert it then.

vtomole:
@PalashGoiporia 

See my earlier comment

> I like that idea. Let's keep in our pockets for when [we will start encrypting other tokens/values ](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming))

pgoiporia:
> @PalashGoiporia
> 
> See my earlier comment
> 
> > I like that idea. Let's keep in our pockets for when [we will start encrypting other tokens/values ](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming))

I see. Will revert.

vtomole (approved):
LGTM

vtomole:
Actually one more thing before the merge

@richrines1 

Why can't we have 

`TOKEN_PUBLIC_KEY_N = 17008548084322796625173082471459551554684756022932834539401654339010944633884012630425812566577967384826523512455600288060673470901154058307403946669015130858987331476342512452717278403114889330081960490338853359064317595272673698503505978525746479076673787620051191726849171552957542444802180154384185543994773754461526344709177863738429950037944448853167130387417072094145230067343804038202835456426731076499420890845328293490670702283588388753166073823091218097231586323734435678442777637262028788120451664812066695626638792656384292908770650084421320032132271166238360431469687923106481791689466937999674374196309`?

pgoiporia:
Looks like we don't need this anymore.

PR #483, opened 2023-06-05
loosen qiskit requirement

richrines1:
* updates qiskit-superstaq to support newer qiskit versions (no longer pinned to those with terra version 0.23.*)
* updates cirq-superstaq[examples] to require terra 0.24.* instead of 0.23.*


vtomole (approved):
LGTM

dowusu-antwi (approved):
Can you test this with one of the `cirq-superstaq` notebooks that use `qiskit`? Otherwise LGTM too.

bharat-thotakura (approved):
LGTM

vtomole:
> Can you test this with one of the cirq-superstaq notebooks that use qiskit

Theoretically the notebook checks should do this job for us already.

dowusu-antwi:
> Theoretically the notebook checks should do this job for us already.


Gp. The notebook checks do `cirq-superstaq/docs/ibm_cirq_superstaq.ipynb` -- but I wondered specifically about `docs/source/optimizations/ibm/ibmq_compile_css.ipynb`

richrines1:
> but I wondered specifically about `docs/source/optimizations/ibm/ibmq_compile_css.ipynb`

just confirmed it does run after installing cirq-superstaq[dev]/[examples]. I also updated the qiskit versions used in the check at the top of the notebook (and removed an equivalent check in css/examples/resource_estimate.ipynb), but that said this check is still broken - it will install the most recent version (via `cirq-superstaq[examples]`), but it won't be reloaded so the notebook will still fail.. so we should definitely rethink how we do this more generally

PR #495, opened 2023-06-08
update check requirements

richrines1:
increases the minimum versions of some of the testing packages we require (the checks won't all pass with the current minima)

i've confirmed everything now passes using the minimum and maximum versions of each req


vtomole (approved):
LGTM

bharat-thotakura (approved):
lgtm

PR #496, opened 2023-06-08
update cirq req to >=1.0.0

### LINKED ISSUE #242: Revert #237 when we update to cirq>=0.15.0
SupertechLabs/cirq-superstaq#237 is a workaround for https://github.com/quantumlib/Cirq/issues/5148 for `cirq==0.14.*`. When it's no longer necessary, `custom_gates_test.test_parallel_gates_equivalence_groups_nonadjacent` should be folded back into `test_parallel_gates_equivalence_groups` and the warning should be removed from the `ParallelGates` docstring

richrines1:
we no longer need to support cirq 0.14.*

closes #242


dowusu-antwi (approved):
LGTM.

PR #499, opened 2023-06-09
Add contact info for exceptions

stephanielee9:
Re: https://github.com/SupertechLabs/server-superstaq/issues/2319


general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -402,7 +402,7 @@ def _handle_status_codes(self, response: requests.Response) -> None:
             else:
                 message = str(response.text)
             raise gss.SuperstaQException(
-                f"Non-retriable error making request to SuperstaQ API, {message}",
+                f"Non-retriable error making request to SuperstaQ API, {message}. If you would like to contact a member of our team, email us at superstaq@infleqtion.com or join our Slack workspace with this URL: https://tinyurl.com/y7t69z9f.",
```
vtomole, line 405: FYI @dowusu-antwi @richrines1 @bharat-thotakura:

We were planning on doing this only for `Internal Server Error`s but decided to do this for all errors.

For example with this change we can get something like

`Non-retriable error making request to SuperstaQ API, the circuit you submitted contains mid-circuit measurements which is not supported on this device. If you would like to contact a member of our team, email us at superstaq@infleqtion.com or join our Slack workspace with this URL: https://tinyurl.com/y7t69z9f.`

richrines1: can we use the actual slack url instead of the tinyurl? imo it's a little sketchy

vtomole: > can we use the actual slack url instead of the tinyurl?

The slack url is long-ish. @stephanielee9 Does slack allow us to have shortened slack links?

stephanielee9: There doesn't seem to be a Slack-generated shortened link from what I can tell

richrines1: ah yeah just saw your comment [here](https://github.com/SupertechLabs/client-superstaq/pull/498#discussion_r1224613262)

fwiw it looks like qiskit just has their own url which redirects to their slack invite: https://ibm.co/joinqiskitslack
which i guess isn't that different from using a tinyurl (though imo still seems a bit more professional/less sketchy) so maybe it's ok

though i'm also ok with the long-ish url, at least it's obvious where it's going :)

general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -402,7 +402,12 @@ def _handle_status_codes(self, response: requests.Response) -> None:
             else:
                 message = str(response.text)
             raise gss.SuperstaQException(
-                f"Non-retriable error making request to SuperstaQ API, {message}",
+                f"""Non-retriable error making request to SuperstaQ API, {message}.
+
+                If you would like to contact a member of our team,
+                email us at superstaq@infleqtion.com or join our Slack workspace:
+                https://join.slack.com/t/superstaq/shared_invite/zt-1wr6eok5j-fMwB7dPEWGG~5S474xGhxw
+                """,
```
richrines1, line 410: this will have weird line breaks/indents. it's better to use single quotes to break up long strings:
```suggestion
            slack_invite_url = (
                "https://join.slack.com/t/superstaq/shared_invite/"
                "zt-1wr6eok5j-fMwB7dPEWGG~5S474xGhxw"
            )
            raise gss.SuperstaQException(
                f"Non-retriable error making request to SuperstaQ API, {message}.\n\n"
                "If you would like to contact a member of our team, email us at "
                f"superstaq@infleqtion.com or join our Slack workspace: {slack_invite_url}",
```
(the slack url could also be placed inline, but seemed nice to have it as its own variable as it's subject to change)

general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -401,8 +401,14 @@ def _handle_status_codes(self, response: requests.Response) -> None:
                 message = response.json()["message"]
             else:
                 message = str(response.text)
+            slack_invite_url = (
+                "https://join.slack.com/t/superstaq/shared_invite/"
+                "zt-1wr6eok5j-fMwB7dPEWGG~5S474xGhxw"
+            )
             raise gss.SuperstaQException(
-                f"Non-retriable error making request to SuperstaQ API, {message}",
+                f"Non-retriable error making request to SuperstaQ API, {message}.\n\n"
```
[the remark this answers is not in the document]
stephanielee9: Yeah! I think that's the purpose behind https://github.com/SupertechLabs/server-superstaq/issues/2319: giving users more helpful error messaging. Does it make sense to leave your note on mid-circuit measurement errors in that issue and create a different PR for it?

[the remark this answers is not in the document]
stephanielee9: Got it, so for this PR, is it worth it to just remove the "Non-retriable error making request to SuperstaQ API" part of the message or are there non-retriable errors we don't have specific messaging for that would be worth to keep it in this PR?

richrines1 (approved):

PR #503, opened 2023-06-12
Adding docstrings to check files

pgoiporia:


pgoiporia:
@richrines1 any tips on how to deal with *args and *kwargs in the docstrings?

richrines1:
i think it's a bit case-specific, but for these i think you can just use something like e.g.
```
    *args: Command line arguments.
```
given that the argument parser's help function already describes their usage

perlinm (approved):
Found it easier to just make nit changes than leave a flurry of comments.  Thanks for filling these in @pgoiporia!  LGTM!

PR #519, opened 2023-06-13
clean up status handling in `css.Job`

richrines1:
* consistent use of `gss.SuperstaqUnsuccessfulJobException` for failed jobs
* improve the error message in said exception
* raise TImeoutError if `job.counts()` times out but the job hasn't failed
* fix mocked statuses in css.job_test which previously weren't any of those in Job.ALL_STATES (in one case causing a test to pass which shouldn't have)

### DIFF
--- a/cirq-superstaq/cirq_superstaq/job.py
+++ b/cirq-superstaq/cirq_superstaq/job.py
@@ -150,7 +150,7 @@ class Job:
             collections.Counter that represents the results of the measurements
 
         Raises:
-            SuperstaQUnsuccessfulJobException: If the job has failed, been canceled, or deleted.
+            SuperstaQUnsuccessfulJobException: If the job failed or has been canceled or deleted.
             SuperstaQException: If unable to get the results from the API.
             TimeoutError: If no results are available in the provided timeout interval.
         """
--- a/cirq-superstaq/cirq_superstaq/job.py
+++ b/cirq-superstaq/cirq_superstaq/job.py
@@ -74,8 +74,8 @@ class Job:
             self._job = self._client.get_job(self.job_id())
 
     def _check_if_unsuccessful(self) -> None:
-        if self.status() in self.UNSUCCESSFUL_STATES:
-            status = self.status()
+        status = self.status()
+        if status in self.UNSUCCESSFUL_STATES:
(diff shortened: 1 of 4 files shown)


cirq-superstaq/cirq_superstaq/job.py
```diff
@@ -75,9 +75,12 @@ def _refresh_job(self) -> None:
 
     def _check_if_unsuccessful(self) -> None:
         if self.status() in self.UNSUCCESSFUL_STATES:
```
dowusu-antwi, line 77:
```suggestion
        if status := self.status() in self.UNSUCCESSFUL_STATES:
```

cirq-superstaq/cirq_superstaq/job.py
```diff
@@ -75,9 +75,12 @@ def _refresh_job(self) -> None:
...
-            raise gss.superstaq_exceptions.SuperstaQUnsuccessfulJobException(
-                self.job_id(), self.status()
-            )
+            status = self.status()
```
dowusu-antwi, line 78: Assuming the above:
```suggestion
```

dowusu-antwi (commented):
Should the internals in `css.Job` start with an underscore? E.g., `TERMINAL_STATES`, `NON_TERMINAL_STATES`, etc.

richrines1:
> Should the internals in `css.Job` start with an underscore? E.g., `TERMINAL_STATES`, `NON_TERMINAL_STATES`, etc.

imo those are meant in part to provide information/functionality to the user, so shouldn't really be considered "internals". they even [have docstrings](https://github.com/SupertechLabs/client-superstaq/blob/cfd1235c55a36161040a15040fb50e3b744ec269/cirq-superstaq/cirq_superstaq/job.py#L35-L40)

dowusu-antwi (approved):
> imo those are meant in part to provide information/functionality to the user, so shouldn't really be considered "internals". they even [have docstrings](https://github.com/SupertechLabs/client-superstaq/blob/cfd1235c55a36161040a15040fb50e3b744ec269/cirq-superstaq/cirq_superstaq/job.py#L35-L40)

Makes sense. LGTM.

Issue #502, opened 2023-06-12
allow custom gate definitions from `qss`


richrines1:
### What is happening?

currently we only allow users to specify custom gate defs to aqt_compile through css. We should extend this to qss

qisit gates aren't as readily serializable as cirq gates, so to start we may only support gates provided as numpy arrays

PR #529, opened 2023-06-16
support custom gate defs in qss

richrines1:
(mostly just adds a custom converter/resolver to qss.serialization because numpy arrays aren't json-serializable)

confirmed this works locally after the corresponding server-side changes

part of #502


bharat-thotakura (approved):
lgtm

PR #536, opened 2023-06-19
Fix broken link in QSCOUT tutorial

knsmith:


knsmith:
When should we use 'latest' vs. 'stable' when linking to the docs?

bharat-thotakura (approved):
Updated links lgtm

Issue #542, opened 2023-06-20
Benchmarking bot


vtomole:
### What is happening?

Add a `run_supermarq.py` to https://github.com/SupertechLabs/client-superstaq/tree/main/supermarq-benchmarks/supermarq that will look like

```python


import os
import cirq

import cirq_superstaq as css
import supermarq

targets = service.get_targets()

qpu_targets = get_qpu_targets(targets)

for target in qpu_targets:
       job_css = service.create_job(
    benchmark.circuit(), repetitions=1000, target=target, options={"tag": "supermarq bot benchmark"})
)
assert job.status() == DONE
```

And add a Github Actions to trigger it monthly.

### What are the current alternatives?

Run Supermarq manually, which takes a lot of time.

### Any additional context?

Supermarq was supposed to be run regularly to monitor hardware progress but in reality we haven't run it in 2 years!

Issue #361, opened 2022-10-26
Add daily integration test for every endpoint in qss and css


stephanielee9:

stephanielee9:
@richrines1 Can you better define this issue for June 1?

PR #544, opened 2023-06-21
Add benchmarking bot

epelaaez:
(#542)

Some things to consider:
- What exact tests do we want to run? For now I just included the ones that are run on the example notebook (pasted below)
```python
BENCHMARKS: List[Tuple[supermarq.benchmark.Benchmark, str]] = [
    (supermarq.ghz.GHZ(5), "ghz5"),
    (supermarq.hamiltonian_simulation.HamiltonianSimulation(4), "hsim4"),
    (supermarq.mermin_bell.MerminBell(3), "mb3"),
    (supermarq.bit_code.BitCode(3, 3, [1, 0, 1]), "bitcode3"),
]
```
- At the moment the jobs are being submitted with `repetitions=1000`, do we want to make this higher?


epelaaez:
Not sure how to go about fixing this type error:
```
supermarq-benchmarks/supermarq/run_supermarq.py:52: error: Argument 1 to "create_job" of "Service" has incompatible type "Union[Circuit, Sequence[Circuit]]"; expected "AbstractCircuit"  [arg-type]
```

supermarq-benchmarks/supermarq/run_supermarq.py
```diff
@@ -0,0 +1,58 @@
...
+
+for target in qpu_targets:
+    for benchmark, label in BENCHMARKS:
+        date = datetime.today().strftime("%Y-%m-%d")
+        tag = f"{date}-{label}-{target}"
+        print(tag)
+
+        # Since there is no way to determine if a target supports mid-circuit measurement purely
+        # through `service.target_info()`, we use a `try` block to catch when we submit an invalid
+        # circuit to a device that does not support it.
```
vtomole, line 49: Create an issue about this and then link to said issue here.

vtomole (commented):
Add a GitHub action here https://github.com/Infleqtion/client-superstaq/tree/main/.github/workflows that will trigger this monthly.

vtomole:
> What exact tests do we want to run?

What we have now is a good start 

> At the moment the jobs are being submitted with repetitions=1000

This is fine. I think the original Supermarq circuits were run on `1000` repetitions as well.

vtomole:
> Not sure how to go about fixing this type error:

There are cases where `benchmark.circuit()` returns a list of circuits instead of a circuit

```python
        try:
            circuit = benchmark.circuit()
            if isinstance(circuit, cirq.Circuit):
                job = service.create_job(
                circuit,
                repetitions=1000,
                target=target,
                options={"tag": tag, "qiskit_pulse": False},
            )
            else:
                for c in circuit:
                    job = service.create_job(
                        c,
                        repetitions=1000,
                        target=target,
                        options={"tag": tag, "qiskit_pulse": False},
                    )

        except Exception:
            pass
```

Should fix

supermarq-benchmarks/supermarq/run_supermarq.py
epelaaez: Created: https://github.com/Infleqtion/client-superstaq/issues/550 and linked to it

epelaaez:
I added the github action, I just have two questions:
- Currently I'm doing `- cron: "0 12 1 * *"`, which triggers every first day of the month at 12. However, I think Honeywell devices are only available during the week, so it would fail if the first of the month is on the weekend. Is there a way to specify it to run every first Monday of the month? I was trying to do that but I couldn't figure it out.
- Am I importing the secrets correctly?

.github/workflows/montly-supermarq-run.yml
```diff
@@ -0,0 +1,28 @@
...
+    - cron: "0 12 1 * *" # Every first of the month at 12
+  workflow_dispatch:  # So that we can also run manually, e.g. for testing
+
+jobs:
+  run-supermarq:
+    runs-on: ubuntu-latest
+    env:
+      SUPERSTAQ_API_KEY : ${{ secrets.SUPERSTAQ_API_KEY }}
+      IBMQ_TOKEN: ${{secrets.IBMQ_TOKEN}}
+      HONEYWELL_PASSWORD: ${{secrets.HONEYWELL_PASSWORD}}
```
vtomole, line 14: We don't use this in the client.

.github/workflows/montly-supermarq-run.yml
```diff
@@ -0,0 +1,28 @@
...
+  schedule:
+    - cron: "0 12 1 * *" # Every first of the month at 12
+  workflow_dispatch:  # So that we can also run manually, e.g. for testing
+
+jobs:
+  run-supermarq:
+    runs-on: ubuntu-latest
+    env:
+      SUPERSTAQ_API_KEY : ${{ secrets.SUPERSTAQ_API_KEY }}
+      IBMQ_TOKEN: ${{secrets.IBMQ_TOKEN}}
```
vtomole, line 13: Not needed to run Supermarq

.github/workflows/montly-supermarq-run.yml
epelaaez: So no API key is needed?

vtomole:
> Is there a way to specify it to run every first Monday of the month?
 
Is https://blog.healthchecks.io/2022/09/schedule-cron-job-the-funky-way/ useful?

.github/workflows/montly-supermarq-run.yml
epelaaez: I made the changes @vtomole

.github/workflows/montly-supermarq-run.yml
```diff
@@ -2,16 +2,12 @@ name: Monthly Supermarq run
...
   schedule:
-    - cron: "0 12 1 * *" # Every first of the month at 12
+    - cron: "0 12 */31,1-7 * 1" # Every first monday of the month at 12
   workflow_dispatch:  # So that we can also run manually, e.g. for testing
 
 jobs:
   run-supermarq:
     runs-on: ubuntu-latest
-    env:
-      SUPERSTAQ_API_KEY : ${{ secrets.SUPERSTAQ_API_KEY }}
```
vtomole: You need the `SUPERSTAQ_API_KEY`

epelaaez: Done.

vtomole (commented):
This looks good to go in but let's wait until https://github.com/Infleqtion/client-superstaq/issues/361 is done cause it's related.

supermarq-benchmarks/supermarq/run_supermarq.py
```diff
@@ -0,0 +1,71 @@
...
+        # Related issue: https://github.com/Infleqtion/client-superstaq/issues/550.
+
+        try:
+            circuit = benchmark.circuit()
+            if isinstance(circuit, cirq.Circuit):
+                job = service.create_job(
+                    circuit,
+                    repetitions=1000,
+                    target=target,
+                    options={"tag": tag, "qiskit_pulse": False, "lifespan": 3653},
```
richrines1, line 60: options should just be kwargs now:
```suggestion
                    tag=tag,
                    qiskit_pulse=False,
                    lifespan=3653,
```
also out of curiosity why aren't we use qiskit pulse for these?

vtomole: >  why aren't we use qiskit pulse for these?

Just cause the first Supermarq run didn't. Happy to revise.

epelaaez: > also out of curiosity why aren't we use qiskit pulse for these?

some of the benchmarking circuits contain resets (product of mid-circuit measurements) that are not supported by pulse (https://github.com/Infleqtion/server-superstaq/issues/2344)

richrines1: imo whenever possible we should use the default configuration of each target (i.e. what users would get with `qiskit_pulse=None`, which is uses a pulse device whenever possible because that's where our strongest optimizations are). would it make sense to check if the circuit has resets, and just set `qiskit_pulse=False` if it does?

it also might make sense to have the server do this automatically, i.e. have it automatically use the non-pulse device if qiskit_pulse is None and the circuit contains resets

epelaaez: > it also might make sense to have the server do this automatically, i.e. have it automatically use the non-pulse device if qiskit_pulse is None and the circuit contains resets

Yeah, I also think this makes sense (opened an issue for this https://github.com/Infleqtion/server-superstaq/issues/2414). Once that gets done we can remove the `qiskit_pulse` arg here.

vtomole (commented):
As discussed IRL, add function that returns all the tags for submitted jobs so that we can use it to call `fetch_jobs/tag`.

vtomole (approved):
Functions that retrieve tags will be addressed in the retrieving bot functionality.

PR #547, opened 2023-06-22
pip page for qiskit-superstaq has a broken link

1ucian0:
pip page for qiskit-superstaq points to the previous location of this source code, I think.


vtomole (approved):
Thank you :)

PR #551, opened 2023-06-22
Fix linking of Colab and Binder badges

### LINKED ISSUE #525: Open in Colab and Binder links don't work in readthedocs
### What is happening?

Open in Colab and Binder links don't work in readthedocs

E.g. the links in https://docs-superstaq.readthedocs.io/en/stable/get_started/basics.html don't work

### What should happen?

They should open correctly as they do here: https://github.com/SupertechLabs/client-superstaq/blob/main/docs/source/get_started/basics/basics_css.ipynb

### Environment

Latest

### Any additional context?

_No response_

singular-value:
Fixes #525. Thanks @stephanielee9 for the fix.

Example rendered locally:
image


docs/source/get_started/basics/basics_qss.ipynb
```diff
@@ -186,7 +172,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.10.6"
+   "version": "3.9.16"
```
dowusu-antwi, line 175: Did we at some point specify which version to use for notebooks? @bharat-thotakura @richrines1

bharat-thotakura: I don't recall a discussion specifically on the version standard to use, but I've tried to stick to version 3.9.16 ever since PR https://github.com/Infleqtion/client-superstaq/pull/490/ standardized it across the notebooks.

dowusu-antwi (approved):
LGTM.

Issue #553, opened 2023-06-22
Stray "References" in docs Table of Contents


singular-value:
### What is happening?

<img width="319" alt="image" src="https://github.com/Infleqtion/client-superstaq/assets/3976048/e9f69a75-b8bc-42cf-9df0-3de9107b415f">

The text is coming from "Superstaq for IBM"

Doesn't happen to "Superstaq for QSCOUT" which also has a References subsection

### How can we reproduce the issue?

Run `make html` from docs/ and open build/html/index.html

### What should happen?

Just like "Superstaq for QSCOUT", the References subsection should not appear in ToC

### Environment

- **SuperstaQ version**: 
- **general-superstaq version**: 
- **cirq-superstaq version**: 
- **qiskit-superstaq version**: 
- **Operating system**: 
- **Python version**: 
- **Any other relevant dependencies**:

PR #554, opened 2023-06-22
Make references in Superstaq for IBMQ a subsection

bharat-thotakura:
Fixes bug #553


dowusu-antwi (approved):
LGTM (sorry, didn't realize that happened)

bharat-thotakura:
> LGTM (sorry, didn't realize that happened)

It was easy to miss, my bad also not catching it in the review.

PR #557, opened 2023-06-23
Switch from `setup.cfg` and `setup.py` to `pyproject.toml`

### LINKED ISSUE #480: Consolidate configs in `setup.py` and `setup.cfg` to `pyproject.toml`
### What is happening?

`pyproject.toml` is trying to become the "standard" way of specifying python project metadata (see [PEP-518](https://peps.python.org/pep-0518/) and [PEP-621](https://peps.python.org/pep-0621/)).  This config file would replace `setup.cfg` and `setup.py`, thereby consolidating our configurations.

This issue supersedes #385, and as a side effect `dev-requirements.txt` would need to:
- remove `black-configparser`, and
- add `flake8-pyproject`.

Fixing this issue would probably also require some non-trivial changes to `check/configs.py`.  One possibility is to only check the contents of `pyproject.toml` after some tag such as `# Python package configurations`.

Leave things as they are.

### Any additional context?

Generally simplifying/consolidating configs seems like a good idea.  Could be a nice + simple software dev starter/intern project.

Also: [example file](https://github.com/jmlarson1/QFI-Opt/blob/1d359379856520b7456dc10848b1f1ae002ff924/pyproject.toml).

perlinm:
To procrastinate on the NSF SBIR, I've been chipping away at #480.

TLDR:
1. `setup.cfg` (which configures check-script tools) is replaced by the top-level `pyproject.toml` file in `client-superstaq`.
2. `setup.py` files inside packages are replaced by package-level `pyproject.toml` files.
3. As a nit cleanup, I prepended `python -m` to the commands run by check scripts, which ensures that these commands are being run from within the python environment.  I also removed some zombies from `__init__.py` files.

Note that once this PR gets merged and deployed, we'll need to update `server-superstaq` with a `pyproject.toml` file replacing `setup.cfg` in order to pass `check/flake8_.py` and `checks/configs.py`.  See corresponding `server-superstaq` [PR](https://github.com/Infleqtion/server-superstaq/pull/2408).

Fixes #480.


general-superstaq/dev-requirements.txt
```diff
@@ -1,6 +1,6 @@
-black-configparser~=0.2.0b2  # warning: potential issues https://github.com/danie1k/python-black-configparser/blob/5c6c8d122cdce95e2c7b4ada773cf2f4267c9189/README.md#known-issues
 black[jupyter]~=22.3.0
 flake8>=3.8.4
+flake8-pyproject
```
perlinm, line 3: Note: imo this should not have any version number attached to it, so that `pip` automatically figures out which version is compatible with `flake8`

perlinm:
The current notebook check failure is:
image
and the fix appears to be to install `pylatexenc`:
image
I'm not sure why this wasn't an error before, though.  Should I add `pylatexenc` to a `requirements.txt` or `dev-requirements.txt` file?

general-superstaq/general_superstaq/check/configs.py
```diff
@@ -79,12 +85,27 @@ def run(
     return 0
 
 
+def _trim_lines(lines: List[str], start_match: str) -> Tuple[List[str], int]:
+    """Remove package-specific configuration text, and identify the starting lines to compare."""
+    start_line = 0
+    if start_match in lines:
+        start_line = lines.index(start_match) + 2
+    return lines[start_line:], start_line
```
richrines1, line 93: nit (avoids iterating through the file twice)
```suggestion
    for start_line, line in enumerate(lines, start=2):
        if start_match in line:
            return lines[start_line:], start_line
    return lines, 0
```

also why the offset by 2? couldn't we just start comparing at the start_match line?

general-superstaq/general_superstaq/check/configs.py
```diff
@@ -12,16 +12,18 @@
 @check_utils.enable_exit_on_failure
 def run(
     *args: str,
-    config_file: str = "setup.cfg",
+    config_file: str = "pyproject.toml",
     ignore_match: str = "# REPO-SPECIFIC CONFIG",
+    start_match: str = "# Package-specific configuration (e.g., build specification) ends here.",
```
richrines1, line 17: imo it'd be cleaner/make the pyproject files more readable if we just used a line like:
```suggestion
    start_match: str = "# Check configuration:",
```

general-superstaq/general_superstaq/check/configs.py
perlinm: The +2 is because we don't want to match the `start_match` (which isn't required to be present), we want to match what's below it... so +1... and then another +1 for an empty newline between `start_match` and the content to match (which looks cleaner because the content itself is broken into blocks separated by newlines -- this way it doesn't look like `start_match` is only associated with the first block).

richrines1:
(also need to replace every instance of "**/setup.py" in .github/workflows/ci.yml)

cirq-superstaq/pyproject.toml
```diff
@@ -0,0 +1,35 @@
...
+cirq_superstaq = ["py.typed"]
+
+[tool.setuptools.dynamic.version]
+attr = "cirq_superstaq._version.__version__"
+
+[tool.setuptools.dynamic.dependencies]
+file = ["requirements.txt"]
+
+[tool.setuptools.dynamic.optional-dependencies.dev]
+file = ["dev-requirements.txt"]
```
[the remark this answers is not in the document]
perlinm: looks like it works

general-superstaq/general_superstaq/check/configs.py
perlinm: On second thought, it's simpler to just start at the `start_line` as in your example

perlinm:
@richrines1 ready for re-review.  One loose end at the moment is that I'm not sure how to fix [`dev_tools/update-pypi-version.sh`](https://github.com/Infleqtion/client-superstaq/blob/main/dev_tools/update-pypi-version.sh).  Maybe @vtomole knows?

vtomole:
Replace `python setup.py bdist_wheel` with `python3 -m build`

perlinm:
Thanks!  Done.

perlinm:
I'm really not sure what's happening with the coverage tests...

vtomole:
@KabirDubey  is running into this issue as well.

richrines1:
^ tests should be fixed by #639

cirq-superstaq/pyproject.toml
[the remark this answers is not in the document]
richrines1: nice! we shouldn't remove `optional-dependencies.examples` though - basically we just want `pip install x-superstaq[dev]` to be a strict superset of `x-superstaq[examples]`, instead of having to use `x-superstaq[dev, examples]` or duplicating requirements in example-requirements.txt and dev-requirements.txt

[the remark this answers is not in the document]
richrines1: relatedly, we should remove qiskit-terra and symengine from cirq-superstaq/dev-requirements.txt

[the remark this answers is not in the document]
perlinm: cool, done

general-superstaq/general_superstaq/check/pyproject.toml
richrines1, line 1: can we rename this something like `pyproject-template.toml` or `pyproject-shared.toml`? (while keeping the name `pyproject.toml` for all the symlinks)

if so we could also (either in this pr or another) move the contents of all the `.iswap.cfg` files into the corresponding project-specific `pyproject.toml` (the reason for the rename is that otherwise isort will see this config as overriding `general-superstaq/pyproject.toml` for just the contents of this subdirectory)

cirq-superstaq/pyproject.toml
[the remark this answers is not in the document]
richrines1: sorry one more thing: the one instance of "[dev,examples]" can now be changed to just "[dev]" in ci.yml

general-superstaq/general_superstaq/check/pyproject.toml
perlinm: Done, but we may as well remove `.isort.cfg` in this PR.

perlinm: Alright I removed `.isort.cfg` files, but now all `isort` configs have to be in every `pyproject.toml` file... which is not **horrible**, but certainly undesirable.

perlinm: Looks like specifying `isort` configs everywhere might be [unavoidable](https://pycqa.github.io/isort/docs/configuration/config_files.html):
image

perlinm: The other option is to have the common `isort` settings (namely: `profile`, `line-length`, and `color-output`) passed manually in `check/format_.py`

richrines1: i don't love the idea of hard coding them in check/format_.py, but maybe we could have it grab the settings from the top-level pyproject.toml?

but either way, this isn't any worse than what we're currently doing so i vote just having the duplicated configs for this pr

perlinm: If looks like as long as there is a `[tool.isort]` entry in project-level `pyproject.toml` files, excluding any settings (at least, the `profile`, `line-length`, and `color` settings) makes them take their default value -- even if a different value is passed explicitly as a command-line argument.  So grabbing them from project-level `pyproject.toml` does not work.

In that case I think this PR is done (but should not be merged until [ss#2408](https://github.com/Infleqtion/server-superstaq/pull/2408) is ready as well).

perlinm: @richrines1 one alternative is to check **all** `pyproject.toml` files in the repository, and enforce that for each entry X in `checks-pyproject.toml` (such as `[tool.isort]` or `[tool.black]`) if X appears in a `pyproject.toml` file, then it must be identical to that in `checks-pyproject.toml`.

The `[tool.isort.known_first_party]` argument would then be set by an exception via, say, `# REPO-SPECIFIC CONFIG`.

richrines1: that's probably overkill imo. these settings aren't likely to be changed much, and if they do it will probably be obvious anyway due to the conflict with the non-isort `black` run in format_.py

general-superstaq/dev-requirements.txt
perlinm: done

richrines1 (approved):
lgtm! thanks this is awesome :)

perlinm:
Thanks!  Now waiting on [ss#2408](https://github.com/Infleqtion/server-superstaq/pull/2408).  No rush though :slightly_smiling_face:

Issue #531, opened 2023-06-16
Update instances of SuperstaQ to Superstaq in docs


stephanielee9:
Related: https://github.com/SupertechLabs/server-superstaq/issues/2375

Also note to update in docstrings

stephanielee9:
`find . -type f -exec sed -i "s/SuperstaQ/Superstaq/g" {} \;`

Also need to update:
- [ ] SupermarQ --> Supermarq
- [ ] SupercheQ --> Supercheq

PR #559, opened 2023-06-23
Update to Superstaq naming in docs + css + qss

bharat-thotakura:
Updates naming from SuperstaQ to Superstaq as part of https://github.com/Infleqtion/client-superstaq/issues/531


richrines1:
fwiw running this from the repo root should change every instance in the repo:
```bash
sed -e 's/SuperstaQ/Superstaq/g' -i $(git ls-files $(find -type f))
```

bharat-thotakura:
> fwiw running this from the repo root should change every instance in the repo:
> 
> ```shell
> sed -e 's/SuperstaQ/Superstaq/g' -i $(git ls-files $(find -type f))
> ```

Yeah, I tried that at first but it changed SuperstaQClient -> SuperstaqClient (etc.) which caused some errors. So I reverted and did manual edits where applicable. Left some notebooks since they will be replaced with the updated tutorials

richrines1:
> changed SuperstaQClient -> SuperstaqClient (etc.)

isn't this what we want?

bharat-thotakura:
> > changed SuperstaQClient -> SuperstaqClient (etc.)
> 
> isn't this what we want?

Oh, that issue is part of #2375 and this PR was meant just for #531 (docs update). Though maybe it's better just to make it one combined PR.

richrines1:
> Oh, that issue is part of the server-side [Infleqtion/server-superstaq#2375](https://github.com/Infleqtion/server-superstaq/issues/2375) and this PR was meant just for #531 (docs update)

ahh my bad, i was just looking at the pr title

> Though maybe it's better just to make it one combined PR.

either way - whichever you think is easiest! changing the class names will also require a corresponding server-side update (which shouldn't be too bad either i think)

knsmith (approved):
Checked the changed files and everything looks good to me! Like separating docs changes from code to keep the PRs more manageable.

PR #561, opened 2023-06-23
move resource_estimate from provider to backend

richrines1:
(basically a continuation of #473)

### DIFF
--- a/qiskit-superstaq/qiskit_superstaq/superstaq_provider.py
+++ b/qiskit-superstaq/qiskit_superstaq/superstaq_provider.py
@@ -18,12 +18,11 @@ import general_superstaq as gss
 import numpy as np
 import numpy.typing as npt
 import qiskit
-from general_superstaq import ResourceEstimate, superstaq_client, user_config
 
 import qiskit_superstaq as qss
 
 
-class SuperstaQProvider(qiskit.providers.ProviderV1, user_config.UserConfig):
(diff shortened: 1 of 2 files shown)


qiskit-superstaq/qiskit_superstaq/superstaq_provider.py
```diff
@@ -117,7 +116,7 @@ def backends(self) -> List[qss.SuperstaQBackend]:
 
     def resource_estimate(
         self, circuits: Union[qiskit.QuantumCircuit, List[qiskit.QuantumCircuit]], target: str
-    ) -> Union[ResourceEstimate, List[ResourceEstimate]]:
+    ) -> Union[gss.ResourceEstimate, List[gss.ResourceEstimate]]:
```
bharat-thotakura, line 119: Maybe I am mistaken, but would this still be? 
```suggestion
    ) -> Union[ResourceEstimate, List[ResourceEstimate]]:
```

richrines1: i removed the `from general_superstaq import ..` line because we already import `gss` and it's only a few more characters to write it out here ([see also](https://github.com/Infleqtion/server-superstaq/blob/main/docs/source/developer_guide.rst#imports))

bharat-thotakura: Oh yeah, noted for other instances.

bharat-thotakura (approved):
LGTM

PR #564, opened 2023-06-26
standardize validation

richrines1:
as much as possible, makes validation consistent across endpoints & clients
* ensures existing validators are called everywhere they're relevant
* makes error messaging more consistent
* moves common validators into `gss.validation` to reduce duplication
* removes some redundancies (e.g. by calling `validate_target` in `gss._SuperstaqClient.create_job` instead of client-specific job creators)


qiskit-superstaq/qiskit_superstaq/superstaq_backend.py
```diff
@@ -210,20 +210,22 @@ def aqt_compile(
...
+            options["num_equivalent_circuits"] = int(num_equivalent_circuits)
         if random_seed is not None:
-            options["random_seed"] = random_seed
+            gss.validation.validate_integer_param(random_seed)
+            options["random_seed"] = int(random_seed)
         if atol is not None:
-            options["atol"] = atol
+            options["atol"] = float(atol)
         if gate_defs is not None:
             options["gate_defs"] = gate_defs
```
paaige, line 226: Check if options is None before converting to json here?

qiskit-superstaq/qiskit_superstaq/superstaq_provider.py
```diff
@@ -305,7 +304,7 @@ def cq_compile(
             ValueError: If `target` is not a valid CQ target.
         """
         if not target.startswith("cq_"):
-            raise ValueError(f"{target} is not a CQ target")
+            raise ValueError(f"{target!r} is not a valid CQ target.")
```
paaige, line 307: Reason for !r in string formatting here vs. elsewhere?

qiskit-superstaq/qiskit_superstaq/validation.py
```diff
@@ -1,4 +1,3 @@
-import re
```
paaige, line 1: Seems sort of weird to have a lone function in this file and not move all of the validation code to gss. I assume this is because we don't want to have qiskit specific functions in gss?

paaige (approved):
Just added some questions comments but otherwise lgtm!

qiskit-superstaq/qiskit_superstaq/superstaq_provider.py
richrines1: i tried to do this everywhere a target appears in an error message - imo it's more clear because it clearly shows the target string as it was passed by the user, e.g.
```
ValueError: 'ibmq_qasm_simulator' is not a valid CQ target.
```
instead of just
```
ValueError: ibmq_qasm_simulator is not a valid CQ target.
```

qiskit-superstaq/qiskit_superstaq/validation.py
richrines1: yep - it requires qiskit, which we don't want to add as a dependency for gss

i agree it's a little weird, but i couldn't come up with a less weird place to put it - we use it in more than one place in qss so it doesn't really make sense in any of the existing files, and it requires qiskit which we don't want to add as a dependency for gss. i eventually justified this to myself by assuming we'll probably have other qiskit-specific validators in the future

qiskit-superstaq/qiskit_superstaq/superstaq_backend.py
richrines1: done (updated in `_get_compile_request_json()`)

PR #566, opened 2023-06-26
standardize aqt_compile arguments

richrines1:
* use `num_eca_circuits` instead of `num_equivalent_circuits` in aqt_compile (this is what the server is looking for, and makes more sense for functions that aren't always used for ECA)
* deprecates `aqt_compile_eca()` (in favor of `aqt_compile()` with `num_eca_circuits=` argument)
* updates function args and docstrings


cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -502,13 +504,14 @@ def _aqt_compile(
...
         options_dict = {**kwargs}
 
-        if num_equivalent_circuits is not None:
-            _validate_integer_param(num_equivalent_circuits)
-            options_dict["num_eca_circuits"] = num_equivalent_circuits
+        if num_eca_circuits is not None:
+            _validate_integer_param(num_eca_circuits)
+            options_dict["num_eca_circuits"] = num_eca_circuits
         if random_seed is not None:
+            _validate_integer_param(num_eca_circuits)
```
bharat-thotakura, line 514: Should this be
```suggestion
            _validate_integer_param(random_seed)
```
?

cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -390,41 +391,6 @@ def resource_estimate(
...
-            sequences for each compiled circuit, and its .pulse_list(s) attribute contains the
-            corresponding list(s) of cycles.
-        """
-
-        _validate_cirq_circuits(circuits)
-        return self._aqt_compile(circuits, target=target, atol=atol, gate_defs=gate_defs, **kwargs)
-
     def aqt_compile_eca(
         self,
         circuits: Union[cirq.Circuit, Sequence[cirq.Circuit]],
```
bharat-thotakura, line 396: I think the type for `circuits` here should also be updated to: 
```suggestion
        circuits: Union[cirq.Circuit, List[cirq.Circuit]],
```
following line 456

cirq-superstaq/cirq_superstaq/service.py
richrines1: oops thanks, good catch

cirq-superstaq/cirq_superstaq/service.py
richrines1: good catch, updated for consistency (it looks like we use `List[Circuit]` in input annotations everywhere in qss/css)

however - i think really we should be using `Sequence` for all of these input annotations. in general input annotations should be as general as possible and output annotations should be as specific as possible, and these functions should work with any `Sequence` type (e.g. tuples) instead of just lists. will open a follow up pr to change this

bharat-thotakura (approved):
LGTM

PR #567, opened 2023-06-27
remove deprecated calls from daily integration checks

richrines1:
follow up to #566 - removes deprecated aqt_compile_eca calls from daily integration checks (instead uses `aqt_compile(..., num_eca_circuits=n)`)

also changes `num_equivalent_circuits` -> `num_eca_circuits` in `qss.compiler_output` for consistency with css


bharat-thotakura (approved):
LGTM after format update

PR #568, opened 2023-06-27
Added GIF code block to readme

stephanielee9:


pgoiporia (approved):
lgtm. Looks super cool!

bharat-thotakura:
I noticed that the GIF uses SuperstaQProvider instead of the recently updated SupertaqProvider, so if it’s not too much extra work, maybe worth updating that. 

(Also, a nice bonus later on could be to have the GIF loop alternative between cirq-superstaq & qiskit-superstaq)

stephanielee9:
> I noticed that the GIF uses SuperstaQProvider instead of the recently updated SupertaqProvider, so if it’s not too much extra work, maybe worth updating that.
> 
> (Also, a nice bonus later on could be to have the GIF loop alternative between cirq-superstaq & qiskit-superstaq)

@bharat-thotakura Oh good catch - Just opened a new PR for the Q --> q change

PR #570, opened 2023-06-27
generalize input annotations from `List[<circuit_type>]` to `Sequence[<circuit_type>]`

richrines1:
in general input annotations should be as general as possible (especially for public-facing functions). these methods don't actually require that circuits be passed as a proper `list` instance (tuples of circuits work just as well, for example)


bharat-thotakura (approved):
LGTM

PR #573, opened 2023-06-27
use `**kwargs` instead of `options=` when creating jobs

richrines1:
for consistency with compile endpoints (we're making all sorts of breaking changes so might as well do this now too)


dowusu-antwi (approved):
LGTM.

bharat-thotakura (approved):

PR #575, opened 2023-06-27
drop support for python 3.7

richrines1:
today is its official eol :) https://devguide.python.org/versions/


bharat-thotakura (approved):
LGTM

PR #578, opened 2023-06-27
Single `# pylint`  for tests

bharat-thotakura:
Removes all the pylint comments for tests and replaces them with a single comment at the top of the test files (excluding gss checks -- to be done by https://github.com/Infleqtion/client-superstaq/pull/503).


vtomole (changes requested):
Now that the test files default to the pragms on top of the files. The individual pragmas in the test can be removed. 

```
general-superstaq/general_superstaq/qubo_test.py:1:# pylint: disable=missing-function-docstring
general-superstaq/general_superstaq/serialization_test.py:1:# pylint: disable=missing-function-docstring
general-superstaq/general_superstaq/superstaq_client_test.py:13:# pylint: disable=missing-function-docstring
```

for example.

bharat-thotakura:
@vtomole I think I'm misunderstanding something, there should now only be a single pragma at the top of the test files. Do you mean some of those should also be removed? For instance, there's only one `# pylint` in `general-superstaq/general_superstaq/qubo_test.py` at line 1,  and removing that will yield a docstring error.

vtomole:
`qubo_test.py` is an exception. A better example would be `general-superstaq/general_superstaq/user_config_test.py` which has 9 pragmas instead of one.

vtomole (approved):
LGTM

Issue #580, opened 2023-06-29
Make `RGate.__init__` follow our docstring standard


vtomole:
`RGate.__init__` should still be fixed, otherwise lgtm!

_Originally posted by @richrines1 in https://github.com/Infleqtion/client-superstaq/pull/513#pullrequestreview-1504656446_

PR #585, opened 2023-06-29
Further docstrings for `qubit_gates.py`

bharat-thotakura:
Addresses #580 + miscellaneous typo fixes in qss.


cirq-superstaq/cirq_superstaq/ops/qubit_gates.py
```diff
@@ -227,16 +229,21 @@ class AceCR(cirq.Gate):
...
     ) -> None:
+        """Initializes an AceCR gate.
+
+        Args:
+            rads: Angle of rotation for CR gate (i.e., twice the angle for each echoed half-CR).
+            sandwich_rx_rads: Angle of rotation for an rx gate applied to the "X side"
+                simultaneously with the X gate on the "Z side".
+
+        Raises:
+            ValueError: If the polarity of `rads` is not '+-' or '-+'.
```
richrines1, line 245:
```suggestion
            ValueError: If the polarity of `rads` is a string other than '+-' or '-+'.
```
(non-string values are ok)

cirq-superstaq/cirq_superstaq/ops/qubit_gates.py
```diff
@@ -663,6 +677,14 @@ class ParallelRGate(cirq.ParallelGate, cirq.InterchangeableQubitsGate):
     """Wrapper class to define a ParallelGate of identical RGate gates."""
 
     def __init__(self, theta: cirq.TParamVal, phi: cirq.TParamVal, num_copies: int) -> None:
+        """Initializes the `ParallelRGate` class.
+
+        Args:
+            theta: The RGate angle (in radians) by which to rotate.
+            phi: The RGate angle (in radians) defining the axis of rotation in the `X`-`Y` plane:
+             `cos(phi) X + sin(phi) Y` (i.e. `phi` radians from `X` to `Y`).
```
richrines1, line 685: nit
```suggestion
                `cos(phi) X + sin(phi) Y` (i.e. `phi` radians from `X` to `Y`).
```
(also same question re: `:math:`)

cirq-superstaq/cirq_superstaq/ops/qubit_gates.py
```diff
@@ -594,14 +607,15 @@ def parallel_gates_operation(*ops: cirq.Operation) -> cirq.Operation:
...
-        """Args:
-        phi: Angle (in radians) defining the axis of rotation in the `X`-`Y` plane:
-             `cos(phi) X + sin(phi) Y` (i.e. `phi` radians from `X` to `Y`).
+        """Initializes an RGate.
 
-        theta: Angle (in radians) by which to rotate.
+        Args:
+            theta: The angle (in radians) by which to rotate.
+            phi: The angle (in radians) defining the axis of rotation in the `X`-`Y` plane:
+             `cos(phi) X + sin(phi) Y` (i.e. `phi` radians from `X` to `Y`).
```
richrines1, line 618: should we be using `:math:` here? (or not using it elsewhere?) (not sure what the our protocol is for this, tbh i don't think i've ever put `:math:` in a docstring)

cirq-superstaq/cirq_superstaq/ops/qubit_gates.py
bharat-thotakura: Corrected.

cirq-superstaq/cirq_superstaq/ops/qubit_gates.py
bharat-thotakura: I've used `:math:` in `qiskit_superstaq/custom_gates.py` for example, mainly for things like `:math: \pi` and matrices. I didn't use it here since `phi` was the name of one of the input args (unlike, say, \pi). Thought about using `:math:` for cosine, sine, X, and Y if rendering in docs was a nit.

dowusu-antwi: Is there anything wrong with copy-pasting the unicode π? My gut reaction is "yes", but I don't know.

If not, I'd argue for that. Otherwise, using `:math:` makes sense to me -- I would say it's low-priority, so we can file an issue for it.

bharat-thotakura: I don't think there's anything wrong with using the Unicode π (I've seen it used in one of the files actually). Qiskit seems to use `:math:` for their docstrings and Cirq seems to just use LateX mode with `$$ $$` afaict. For me, I guess I wanted to stick to one standard rendering mode for all math symbols but most docstrings don't have many explicit math expressions/formulas to begin with so not really an issue to use Unicode characters once off.

richrines1: agree re: choosing something consistent. there also might be accessibility-related reasons to prefer one or another, which we should look into

i think all else being equal i generally prefer unicode characters, especially if it's just for e.g. a single greek letter (usually i'm reading docstrings in a terminal/text editor, which will render unicode `π` but not e.g. ``:math:`-\pi\`:``. but for this pr i think whatever seems most consistent with what we have is fine :)

richrines1: (otherwise this pr looks good to me, will approve)

richrines1 (approved):

PR #558, opened 2023-06-23
update_submit_qubo

dhirpalshah:
sent qubo to toshiba qpu if method is dry-run, still need to integrate it with correct qubo obj

PR #587, opened 2023-06-29
reorganized docs superstaq and deleted unnecessary application examples

dhirpalshah:
reorganizing the files in our apps directory


vtomole (approved):

vtomole:
This file simply moves things around in anticipation of https://github.com/Infleqtion/client-superstaq/pull/558

dhirpalshah:
There will be an odd instance when building the html where when clicking into the max sharpe ratio example, the page seems to expand a little bit. I will be editing that file in #558 to alleviate this issue.

Issue #563, opened 2023-06-26
Update docstrings in `client-superstaq`


dowusu-antwi:
### What is happening?

From Open-Beta week: "Update docstrings in client --> add docstrings to methods and classes (make sure there are no pragmas for disabling docstrings in clients)" ~(also do this server-side, unless that's already done?)~

- Remove `#pylint` from test functions and add a single `#pylint` at the top of each test file
- Update old docstrings
- Review @pgoiporia's docstring spreadsheet and confirm changes

### What are the current alternatives?

None.

### Any additional context?

None.

vtomole:
Server-side updates are big enough to be it's own issue. I've re-titled this one to be more specific.

bharat-thotakura:
@dowusu-antwi 

I think PR #565 should address the first bullet point and I am currently looking at files that require docstrings (I knew already for example that `ops/qubit_gates.py` requires it). Should we make a checklist here or do you have a suggestion on how we should proceed?

dowusu-antwi:
Sounds good, thanks!

> Should we make a checklist here or do you have a suggestion on how we should proceed?

I think we can use the spreadsheet checklist Palash made (he mentioned it probably need updates).

bharat-thotakura:
@dowusu-antwi, going through the Excel file, all docstrings look to be fairly complete after merging #503 & #513. Further updates imo would mostly be of the 'nit kind' and increased conformity to the dev guide at the moment.

bharat-thotakura:
To keep track: 

> Remove #pylint from test functions and add a single #pylint at the top of each test file

Completed by https://github.com/Infleqtion/client-superstaq/pull/565 & https://github.com/Infleqtion/client-superstaq/pull/578

> Update old docstrings

Addressed by #585, #588, #589, #590, & #591.

> Review @pgoiporia's docstring spreadsheet and confirm changes

Done (spreadsheet also updated now). 

This issue can be closed imo from merging the PRs listed above.

PR #588, opened 2023-06-30
Revised docstrings in `supermarq`

bharat-thotakura:
Part of the 'Update old docstrings' point #563.


supermarq-benchmarks/supermarq/plotting.py
```diff
@@ -237,15 +237,14 @@ def annotate_heatmap(
...
+        data: Optional data used to annotate. If None, the image's data is used.
+        valfmt: An optional format of the annotations inside the heatmap.  This should either
+            use the string format method, e.g. "$ {x:.2f}", or be a `matplotlib.ticker.Formatter`.
         textcolors: A pair of colors.  The first is used for values below a threshold, the second
-            for those above.  Optional.
-        threshold: Value in data units according to which the colors from textcolors are
-            applied. If None (the default) uses the middle of the colormap as
-            separation. Optional.
+            for those above. Defaults to black and white respectively.
+        threshold: An optional value in data units according to which the colors from textcolors
```
dowusu-antwi, line 246:
```suggestion
        threshold: An optional value in data units according to which of the colors from textcolors
```

supermarq-benchmarks/supermarq/plotting.py
```diff
@@ -284,9 +283,9 @@ def annotate_heatmap(
 
 
 def radar_factory(num_vars: int) -> npt.NDArray[np.float_]:
-    """(https://matplotlib.org/stable/gallery/specialty_plots/radar_chart.html)
+    """Create a radar chart with `num_vars` axes.
 
-    Create a radar chart with `num_vars` axes.
+    (https://matplotlib.org/stable/gallery/specialty_plots/radar_chart.html)
 
     This function creates a RadarAxes projection and registers it.
```
dowusu-antwi, line 290:
```suggestion
    (https://matplotlib.org/stable/gallery/specialty_plots/radar_chart.html) This function creates a
    RadarAxes projection and registers it.
```

dowusu-antwi (approved):
LGTM, with comments.

supermarq-benchmarks/supermarq/converters.py
```diff
@@ -20,7 +20,8 @@ def cirq_to_qiskit(circuit: cirq.Circuit) -> qiskit.circuit.QuantumCircuit:
 def compute_communication_with_qiskit(circuit: qiskit.circuit.QuantumCircuit) -> float:
     """Compute the program communication of the given quantum circuit.
 
-    Program communication = circuit's average qubit degree / degree of a complete graph
+    Note:
+        Program communication = circuit's average qubit degree / degree of a complete graph.
```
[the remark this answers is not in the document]
bharat-thotakura: I thought that formatting was more standard in Google Style Python Docstrings. In one of the examples,

```ruby
def example_method(self, param1, param2):
        """Class methods are similar to regular functions.

        Note:
            Do not include the `self` parameter in the ``Args`` section.

        Args:
            param1: The first parameter.
            param2: The second parameter.

        Returns:
            True if successful, False otherwise.

        """
        return True
```

supermarq-benchmarks/supermarq/plotting.py
bharat-thotakura: Done.

supermarq-benchmarks/supermarq/plotting.py
bharat-thotakura: Good call, updated.

supermarq-benchmarks/supermarq/converters.py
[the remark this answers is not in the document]
dowusu-antwi: Oh interesting. Basically a nit, but I think we decided to go right into text, without the `Note:`. See [here](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) (this was linked in the server-side dev guide section on docstrings). @richrines1 was this choice deliberate?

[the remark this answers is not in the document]
bharat-thotakura: I don't have a strong preference either way, my thought process in this instance was that using `Note:` would emphasize/highlight the equation specifically (amongst the text). But will revert to follow our dev guide more.

PR #589, opened 2023-06-30
Revised `cirq-superstaq` docstrings

bharat-thotakura:
Part of the 'Update old docstrings' bullet point of https://github.com/Infleqtion/client-superstaq/issues/563 in css.


cirq-superstaq/cirq_superstaq/compiler_output.py
```diff
@@ -99,10 +123,11 @@ def __init__(
         self.seq = seq
 
     def has_multiple_circuits(self) -> bool:
-        """Returns True if this object represents multiple circuits.
+        """Checks if an object has .circuits and .pulse_lists attributes. Otherwise, the
+        object represents a single circuit, and has .circuit and .pulse_list attributes.
```
dowusu-antwi, line 127: Shouldn't this be split into a single line + additional text after a newline?

cirq-superstaq/cirq_superstaq/compiler_output.py
```diff
@@ -124,13 +149,15 @@ def read_json(json_dict: Dict[str, Any], circuits_is_list: bool) -> CompilerOutp
     """Reads out returned JSON from Superstaq API's IBMQ compilation endpoint.
 
     Args:
-        json_dict: a JSON dictionary matching the format returned by /ibmq_compile endpoint
-        circuits_is_list: bool flag that controls whether the returned object has a .circuits
+        json_dict: A JSON dictionary matching the format returned by /ibmq_compile endpoint
+        circuits_is_list: A bool flag that controls whether the returned object has a .circuits
             attribute (if True) or a .circuit attribute (False)
```
dowusu-antwi, line 154:
```suggestion
            attribute (if True) or a .circuit attribute (False).
```

cirq-superstaq/cirq_superstaq/compiler_output.py
```diff
@@ -256,12 +284,13 @@ def read_json_qscout(json_dict: Dict[str, Any], circuits_is_list: bool) -> Compi
...
-        circuits_is_list: bool flag that controls whether the returned object has a .circuits
-            attribute (if True) or a .circuit attribute (False)
+        json_dict: A JSON dictionary matching the format returned by /qscout_compile endpoint.
+        circuits_is_list: A bool flag that controls whether the returned object has a .circuits
+            attribute (if True) or a .circuit attribute (False).
+
     Returns:
-        a CompilerOutput object with the compiled circuit(s) and a list jaqal programs
-        represented as strings
+        A `CompilerOutput` object with the compiled circuit(s) and a list jaqal programs
```
dowusu-antwi, line 292:
```suggestion
        A `CompilerOutput` object with the compiled circuit(s) and a list of jaqal programs
```

cirq-superstaq/cirq_superstaq/sampler.py
```diff
@@ -50,15 +50,12 @@ def __init__(
         service: css.service.Service,
         target: str,
     ) -> None:
-        """Constructs the sampler. Uers should get a sampler from the `sampler` method on
+        """Constructs the sampler. Users should get a sampler from the `sampler` method on
         `css.Service`.
```
dowusu-antwi, line 54: Should be one line, e.g., 
```suggestion
        """Constructs the sampler, accessed from the `sampler` method on `css.Service`.
```

dowusu-antwi (commented):
I'm unsure how strict to be about keeping the initial docstring description to a single line. I think the goal is to:
- (a) avoid making functions/classes more complex than necessary, and
- (b) condense documentation so it's user-friendly

A longer description after a carriage-return should support this. Even if it includes redundant information, it's just a docstring so that's fine -- in such case, I'd think of the first line as a tl;dr.

cirq-superstaq/cirq_superstaq/compiler_output.py
bharat-thotakura: You're right, fixed now.

cirq-superstaq/cirq_superstaq/compiler_output.py
bharat-thotakura: Thanks for the catch, fixed.

cirq-superstaq/cirq_superstaq/compiler_output.py
bharat-thotakura: Done.

cirq-superstaq/cirq_superstaq/sampler.py
bharat-thotakura: Done.

dowusu-antwi (approved):
LGTM.

bharat-thotakura:
^ I would agree that moving forward we should try to stick to a single-line initial docstring. I wasn't as strict in this PR, but do let me know if you catch other instances (or maybe another PR can address all such instances separately).

PR #607, opened 2023-07-07
Revert "rename all instances of `_SuperstaqClient` to `SuperstaqClient`"

dhirpalshah:
Reverts Infleqtion/client-superstaq#603


bharat-thotakura (approved):
lgtm

PR #608, opened 2023-07-07
Rename `userconfig` to `service` in `general_superstaq` and write `submit_qubo` in `service.py`

dhirpalshah:


dhirpalshah:
the file previously named `userconfig` is now `service`, and all the appropriate changes have been made throughout the repository to address this change.

dhirpalshah:
this is all in effort to keep the user from accessing private methods like '_superstaqclient` as mentioned in this thread of PRs: https://github.com/Infleqtion/client-superstaq/pull/603.

cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -99,7 +99,7 @@ def counts_to_results(
     return result
 
 
-class Service(user_config.UserConfig):
+class Service(service.Service):
```
vtomole, line 102: This can be a bit confusing as to which `Service` is being imported. Let's do 
```suggestion
class Service(gss.service.Service):
```

general-superstaq/general_superstaq/service.py
```diff
@@ -1,14 +1,17 @@
...
+    from general_superstaq import superstaq_client
 
-class UserConfig:
+
+class Service:
     """This class contains all the user configurations that are used to operate Superstaq."""
 
-    def __init__(self, client: gss.superstaq_client._SuperstaqClient):
-        """Initializes the `UserConfig` class.
+    def __init__(self, client: "superstaq_client._SuperstaqClient"):
```
vtomole, line 13:
```suggestion
    def __init__(self, client: gss.superstaq_client._SuperstaqClient):
```

Works for me. Could be something wrong with your local machine.

general-superstaq/general_superstaq/service.py
```diff
@@ -1,14 +1,17 @@
 import os
-from typing import Any, Dict, Optional, Tuple, Union
+from typing import TYPE_CHECKING, Any, Dict, Optional, Tuple, Union
 
-import general_superstaq as gss
+import qubovert as qv
 
+if TYPE_CHECKING:
```
vtomole, line 6:
```suggestion
import general_superstaq as gss
```

vtomole (changes requested):
Did you run the MIS notebook with this code to make sure things still work?

general-superstaq/general_superstaq/service.py
dhirpalshah: I thought the plan was to deploy this and then make the changes to the MIS notebook since the kernel in the jupyter notebook requires the change to be made in the main branch. What do you think?

vtomole: I'm only talking about testing here. You can run the MIS notebook from this branch but don't commit the notebook.

general-superstaq/general_superstaq/service.py
```diff
@@ -1,14 +1,15 @@
...
+from typing import TYPE_CHECKING, Any, Dict, Optional, Tuple, Union
 
-import general_superstaq as gss
+import qubovert as qv
 
 
-class UserConfig:
+
+class Service:
     """This class contains all the user configurations that are used to operate Superstaq."""
```
pgoiporia, line 9: This probably needs an updated docstring. I'm assuming `solve_qubo` falls outside the umbrella of simply setting using configurations.

general-superstaq/general_superstaq/service.py
```diff
@@ -94,6 +95,26 @@ def update_user_role(self, email: str, role: int) -> str:
             }
         )
 
+    def solve_qubo(
```
pgoiporia, line 98: I see this is being added here, but where is it being deleted from? Or did this function not exist before this PR?

general-superstaq/general_superstaq/service.py
dhirpalshah: Seems to be a local issue - here is the following error I get when I use that line:

`AttributeError: partially initialized module 'general_superstaq' has no attribute 'superstaq_client' (most likely due to a circular import)`

pgoiporia (changes requested):
Very brief review. Everything else looks good! I'm a bit surprised that changing UserConfig -> Service was this easy though. Did nothing else really break?

general-superstaq/general_superstaq/service.py
dhirpalshah: Sounds good, we will address this in the next PR.

general-superstaq/general_superstaq/service.py
dhirpalshah: A lot of stuff broke, but we were able to fix the issues. Docstrings are updated (minor tweak). Done.

general-superstaq/general_superstaq/service.py
dhirpalshah: Correct, this method did not exist prior to this PR. This method simply calls the `submit_qubo` that is in `superstaq_client.py` (so it was a minor edit).

BTW, I have renamed `solve_qubo` to `submit_qubo`, as that is the name of the function our notebook calls.

dhirpalshah:
This PR will be followed by two other PRs. The first will be to change the notebook to correctly call `submit_qubo` as mentioned here:
> in the next PR, we will actually integrate it into `max_sharpe_ratio_optimization.ipynb`.
Then, the second PR will be to refactor `_SuperstaqClient` such that the clients are inheriting from it and not from `Service` (formerly `UserConfig`) as @richrines1 and @vtomole have discussed, so that we reduce redundancy in not needing every endpoint be a method in both `_SuperstaqClient` and `UserConfig`.

pgoiporia (approved):
lgtm. nice work:)

dhirpalshah:
> @dhirpalshah @vtomole Should this be 2 separate PRs?

It makes sense that this should be one PR because the reason we're doing all this renaming is so we can include the new function `submit_qubo` that is in `service.py`.

vtomole (approved):
LGTM

PR #609, opened 2023-07-10
Add limit to user balance

paaige:


paaige:
@vtomole we need to add a limit for internal users (us), right? What did we agree on that being?

general-superstaq/general_superstaq/user_config_test.py
```diff
@@ -69,6 +69,15 @@ def test_update_user_balance(
     )
 
 
+def test_update_user_balance_limit() -> None:  # pylint: disable=missing-function-docstring
```
vtomole, line 72: I believe we already ignore these pragmas for the test files.
```suggestion
def test_update_user_balance_limit() -> None:
```

vtomole (commented):
This change needs to be done server side because I can still use the Swagger UI to go above this limit.

vtomole:
@paaige  limit will be the same for both internal and external users.

paaige:
> This change needs to be done server side because I can still use the Swagger UI to go above this limit.

Yeah, I think we should do both, because most users are using the API programmatically and on the client side it seems better to fail first before sending a request

vtomole:
> I think we should do both

One thing i'm worried about doing this twice is that it will slip past review in the future. If this limit is ever changed, the reviewer will have to remember that it needs to changed in 2 places.

paaige:
Sister PR: [#2470](https://github.com/Infleqtion/server-superstaq/pull/2470)

general-superstaq/general_superstaq/service_test.py
```diff
@@ -58,6 +58,12 @@ def test_update_user_balance(
     )
 
 
+def test_update_user_balance_limit() -> None:  # pylint: disable=missing-function-docstring
```
vtomole, line 61: I think this pragma can be removed cause we already ignore docstring for test files.
```suggestion
def test_update_user_balance_limit() -> None:
```

vtomole (approved):
LGTM

PR #616, opened 2023-07-11
needed to rename userconfig to service and remove finance / logistics in `general_superstaq.rst`

dhirpalshah:
edits that needed to be made in the `general_superstaq.rst` file for renaming / removal purposes.


vtomole (approved):
LGTM

Issue #583, opened 2023-06-29
Add maxout for submit_qubo


dhirpalshah:
Users may want to have control over the number of solutions that `submit_qubo` outputs. In the `/server-superstaq/superstaq/request_converter.py`, the function `solve_qubo` calls `solver = SimulatedBifurcationMachine(workspace, loops=1000, maxout=1)`. This only outputs the first and best solution. We can try to include a parameter for clients to specify the number of solutions (maxout value) they want.

dhirpalshah:
also change loop # to 1 million, since timeout will then be the limiting factor

dhirpalshah:
addressed in https://github.com/Infleqtion/server-superstaq/pull/2475 (server-side), and also in https://github.com/Infleqtion/client-superstaq/pull/621

PR #621, opened 2023-07-12
added `maxout` parameters in `service.py` and `superstaq_client.py`

dhirpalshah:
Closes https://github.com/Infleqtion/client-superstaq/issues/583.

still have yet to guarantee that it works on the notebooks. this is the next step.


dhirpalshah:
i ran the server on my local host and the changes are reflected correctly - we can now add in a `maxout` parameter to the `submit_qubo` function from the client to specific number of solutions outputted before the job terminates.

vtomole (approved):
LGTM

Issue #268, opened 2022-06-13
`examples_test.py`


vtomole:
Similar to the notebooks we don't run our examples in the CI.

stephanielee9:
Close by deleting `.py` example files

PR #624, opened 2023-07-13
Remove `.py` files from examples

vtomole:
We've decided that examples should only be notebooks.

Fixes https://github.com/Infleqtion/client-superstaq/issues/268


epelaaez (approved):
lgtm!

PR #626, opened 2023-07-14
use minimal cirq/qiskit versions in standalone checks

### LINKED ISSUE #287: create github action using older cirq version
we assert that cirq-superstaq supports cirq>=0.14.0; we should add a test to make sure pytest actually passes with the older cirq version

richrines1:
for every other test we use the most recent versions of cirq/qiskit, so this makes sure the tests still pass with the lowest versions allowed by our requirements.txts

(turns out one change was required to support cirq==1.0.0 - `cirq.MatrixGate.with_name()` was added in 1.1.0)

closes #287

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -147,12 +147,12 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: 3.8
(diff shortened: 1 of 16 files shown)


epelaaez (approved):
lgtm

PR #631, opened 2023-07-17
fixed docstrings for `maxout` parameter in `submit_qubo` method

### LINKED ISSUE #627: Fix `maxout` parameter docstrings
Lies in the `submit_qubo` function in `superstaq_client`. In reference to the maxout parameter created in #583.

dhirpalshah:
Closes #627.


dhirpalshah:
> would it make sense to also change the variable name to something more descriptive?

changed `maxout` to `max_solutions` to be a little more descriptive. NOTE: the change is only in `service.py` as opposed to in `superstaq_client.py` as well, because azure's function still searches for `"maxout"` as the argument. So, the change is reflected for the users in `service`, but not in the backend.

general-superstaq/general_superstaq/service.py
```diff
@@ -124,21 +124,25 @@ def submit_qubo(
...
-        """Solves the QUBO given via the submit_qubo function in superstaq_client
+        """Solves the QUBO given via the submit_qubo function in superstaq_client, and returns any
+        number of specified dictionaries that seek the minimum of the energy landscape from the
+        given objective function known as output solutions.
 
         Args:
             qubo: A `qv.QUBO` object.
             target: The target to submit the qubo.
-            repetitions: Number of repetitions to use. Defaults to 1000.
+            repetitions: Optional number of times that the execution is repeated before stopping.
```
vtomole, line 136: Why are we saying stuff is `optional` and what it defaults to? These should be self-documenting from the type annotations.

dhirpalshah: seemed like in the previous documentation it mentioned whether stuff was optional or not, but you make a good point. I've made the changes - done.

richrines1: > seemed like in the previous documentation it mentioned whether stuff was optional or not

i think in some cases mentioning that it's optional/describing the default behavior is useful, e.g.
```
    num_eca_circuits: Optional number...
```
makes sense imo because it tells the compiler to return something different than it otherwise would (ECA circuits). Or for this function, it's not obvious from the annotation what the default `method=None` does, so it might be useful to describe the default behavior instead of saying it "defaults to None"

general-superstaq/general_superstaq/service.py
```diff
@@ -124,21 +124,24 @@ def submit_qubo(
         target: str,
         repetitions: int = 1000,
         method: Optional[str] = None,
-        maxout: Optional[int] = 1000,
+        max_solutions: Optional[int] = 1000,
```
richrines1, line 127: `Optional[<x>]` is just an alias for `Union[<x>, None]` - this parameter should always be an integer so it's not necessary here 
```suggestion
        max_solutions: int = 1000,
```

general-superstaq/general_superstaq/service.py
```diff
@@ -124,21 +124,24 @@ def submit_qubo(
...
+        given objective function known as output solutions.
 
         Args:
             qubo: A `qv.QUBO` object.
             target: The target to submit the qubo.
-            repetitions: Number of repetitions to use. Defaults to 1000.
-            method: An optional method parameter. Defaults to None.
-            maxout: An optional maxout paramter. Defaults to 1000.
+            repetitions: Number of times that the execution is repeated before stopping.
+            method: The parameter specifying method of QUBO solving execution. Defaults to none.
```
richrines1, line 137: probably better to say what the default method is - it's clear from the annotation that it defaults to None, but not clear what this means

general-superstaq/general_superstaq/service.py
```diff
@@ -124,21 +124,24 @@ def submit_qubo(
...
+            max_solutions: A parameter that specifies the max number of output solutions.
+            Defaults to 1000.
```
richrines1, line 139: per @vtomole's comment this isn't necessary i think
```suggestion
```

general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -336,9 +336,10 @@ def submit_qubo(
         Args:
             qubo: A `qv.QUBO` object.
             target: The target to submit the qubo.
-            repetitions: Number of repetitions to use. Defaults to 1000.
-            method: An optional method parameter. Defaults to None.
-            maxout: An optional maxout paramter. Defaults to 1000.
+            repetitions: Number of times that the execution is repeated before stopping.
+            method: The parameter specifying method of QUBO solving execution. Defaults to none.
+            maxout: A parameter that specifies the max number of output solutions.
+            Defaults to 1000.
```
richrines1, line 342:
```suggestion
```

general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -336,9 +336,10 @@ def submit_qubo(
         Args:
             qubo: A `qv.QUBO` object.
             target: The target to submit the qubo.
-            repetitions: Number of repetitions to use. Defaults to 1000.
-            method: An optional method parameter. Defaults to None.
-            maxout: An optional maxout paramter. Defaults to 1000.
+            repetitions: Number of times that the execution is repeated before stopping.
+            method: The parameter specifying method of QUBO solving execution. Defaults to none.
+            maxout: A parameter that specifies the max number of output solutions.
```
richrines1, line 341: we should add `gss.validation.validate_integer_param(maxout)` alongside the other validators below

richrines1 (commented):
> NOTE: the change is only in `service.py` as opposed to in `superstaq_client.py` as well, because azure's function still searches for `"maxout"` as the argument. So, the change is reflected for the users in `service`, but not in the backend.

might be worth updating the server to look for "max_solutions" instead of "maxout" make this consistent now, rather than living with the inconsistency or having to make a breaking change later

general-superstaq/general_superstaq/service.py
dhirpalshah: done

general-superstaq/general_superstaq/service.py
dhirpalshah: done

general-superstaq/general_superstaq/superstaq_client.py
dhirpalshah: done.

general-superstaq/general_superstaq/superstaq_client.py
dhirpalshah: done.

general-superstaq/general_superstaq/service.py
dhirpalshah: done.

general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -327,30 +327,33 @@ def submit_qubo(
...
         gss.validation.validate_integer_param(repetitions)
+        gss.validation.validate_integer_param(max_solutions)
 
         json_dict = {
             "qubo": gss.qubo.convert_qubo_to_model(qubo),
             "target": target,
             "shots": int(repetitions),
             "method": method,
-            "maxout": maxout,
+            "maxout": max_solutions,
```
richrines1, line 356:
```suggestion
            "max_solutions": max_solutions,
```

vtomole (approved):
@richrines1's comments addressed

PR #632, opened 2023-07-17
Disable coverage/pytest warnings

epelaaez:
fixes https://github.com/Infleqtion/server-superstaq/issues/2406

### DIFF
--- a/general-superstaq/general_superstaq/check/coverage_.py
+++ b/general-superstaq/general_superstaq/check/coverage_.py
@@ -51,6 +51,7 @@ def run(
 
     if not parsed_args.enable_socket:
         pytest_args.append("--disable-socket")
+    pytest_args.append("--disable-warnings")
 
     coverage_arg = "--include=" + ",".join(files)
     test_returncode = subprocess.call(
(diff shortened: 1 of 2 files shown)


vtomole (changes requested):
We want to disable specific warnings. Can we only disable the warnings coming from `qiskit`? That's what's flooding our terminal. Even better: is it easy to fix those qiskit warnings?

epelaaez:
> Can we only disable the warnings coming from qiskit? 

I think we can do something like: https://stackoverflow.com/a/53218641/13579299.

richrines1:
> We want to disable specific warnings. Can we only disable the warnings coming from `qiskit`?

we already do this for some qtrl warnings server-side: https://github.com/Infleqtion/server-superstaq/blob/main/pytest.ini

richrines1:
> Even better: is it easy to fix those qiskit warnings?

looked into this briefly, basically all the warnings appear as soon as `qiskit_ibm_runtime` is imported in a pytest - they still appear even if you just save a file like
```python
### foo.py

import qiskit_ibm_runtime

def test_foo():
    pass
```
and run
```
pytest foo.py
```
so afaik the only way to avoid suppressing the warnings is to file an issue on qiskit_ibm_runtime

epelaaez:
> we already do this for some qtrl warnings server-side

so this can be done server-side without changing anything here?

epelaaez:
closed as superseded by https://github.com/Infleqtion/server-superstaq/pull/2491

Issue #545, opened 2023-06-21
Retrieve partial job results from an unsuccessful batch job


bharat-thotakura:
### What is happening?

Following PR #468, results from a successful sub-job in qss are stored and can be retrieved even if the overarching job was unsuccessful. It would be a helpful feature to more easily retrieve successful sub-job results immediately (even if some sub-jobs are running) by passing some method-type argument or flag (see https://github.com/SupertechLabs/client-superstaq/pull/468#discussion_r1237388397) to `result()`. 

### What are the current alternatives?

With #468, calling `job.result().get_counts(n)` for a successful sub-job-n could give you partial job results.

### Any additional context?

See discussion in PR #468.

bharat-thotakura:
Related: https://github.com/Infleqtion/client-superstaq/pull/691#discussion_r1321885197

vtomole:
@cdbf1 Is this done by job refactor?

PR #636, opened 2023-07-24
Match qss counts keys to circuit classical register bits

### LINKED ISSUE #405: `job.result().get_counts()`s keys don't match `qiskit's
### What is happening?

Printing the counts from `qss` doesn't look like what you get when you print them from `qiskit` which can be baffling for a user.

```python
import qiskit

qc = qiskit.QuantumCircuit(4, 4)

qc.h(2)
qc.cx(2, 3)
qc.measure(2, 2)
qc.measure(3, 3)

token = "token"

superstaq = qss.superstaq_provider.SuperstaQProvider(token)
backend = superstaq.get_backend("ibmq_qasm_simulator")
simulator = qiskit.Aer.get_backend('aer_simulator')

# Run and get counts
result = simulator.run(qc, shots=10).result()
counts = result.get_counts(qc)

print(counts)
# prints
# {'1100': 7, '0000': 3}

job = backend.run(qc, shots=10)
print(job.result().get_counts())
# prints
# {'00': 4, '11': 6}
```

### What should happen?

The keys on the second call to `job.result().get_counts())` should also be {`1100': x, '0000':y}`

### Environment

- **SuperstaQ version**: `0.3.28`
- **general-superstaq version**:  `0.3.21`
- **cirq-superstaq version**:  `0.3.21`
- **qiskit-superstaq version**:  `0.3.21`
- **Operating system**:  Ubuntu
- **Python version**: `3.10`

### Any additional context?

_No response_

bharat-thotakura:
Fixes #405. For example, 

counts_fixed


qiskit-superstaq/qiskit_superstaq/superstaq_backend.py
```diff
@@ -28,13 +28,14 @@ class SuperstaqBackend(qiskit.providers.BackendV1):
     """This class represents a Superstaq backend."""
 
     def __init__(self, provider: qss.SuperstaqProvider, target: str) -> None:
-        """Initializes a SuperstaqBackend.
+        """Initializes a `SuperstaqBackend`.
```
pgoiporia, line 31: Is this something we're meant to do? If so, let's add it to the dev guide. I myself have not been code-styling objects.

bharat-thotakura: We definitely use backticks for code snippets like `qiskit.QuantumCircuit` for example. But as the docstrings have been updated, the standard seems to be that `args` and `objects` should be code-styled. Alternatively, we could use `:class:` and `:attr:`, but I haven't seen any instance of that. 

Though maybe @vtomole @richrines1 can clarify what guideline should be added to the dev guide.

pgoiporia:
@bharat-thotakura this honestly looks good to me, but there are probably changes to be suggested. Could you request a review from someone else? I unfortunately don't have much input to offer.

bharat-thotakura:
@pgoiporia, will request review from @richrines1 also since he was involved in the corresponding issue.

qiskit-superstaq/qiskit_superstaq/superstaq_job.py
```diff
@@ -57,9 +58,47 @@ def _wait_for_results(self, timeout: float, wait: float = 5) -> List[Dict[str, D
...
+        Returns:
+            The number of classical bits in the circuit.
+        """
+        if self._circ_meas_bit_indices == []:
+            self._circ_meas_bit_indices = [
+                qss.compiler_output.measured_qubit_indices(circuit)
+                for circuit in self.compiled_circuits()
+            ]
+
+        return self._backend._num_clbits_in_circ[index]
```
richrines1, line 100: nit: this seems to be doing too different things (the returned value is unrelated to the fetched `_circ_meas_bit_indices`). imo this should either be separated into two functions (or, `self._backend._num_clbits_in_circ[index]` could just be used inline in `result()`)

also do we need to make `_circ_meas_bit_indices` an attribute of the job? it seems like it only needs to be computed once in `result()`, in which case this should just return it as a value

qiskit-superstaq/qiskit_superstaq/superstaq_job.py
```diff
@@ -57,9 +58,47 @@ def _wait_for_results(self, timeout: float, wait: float = 5) -> List[Dict[str, D
...
+    def _fetch_meas_info(self, index: int) -> int:
+        """Helper method to update measurement indices and get number of classical bits.
+
+        Args:
+            index: The index of the circuit of interest.
+
+        Returns:
+            The number of classical bits in the circuit.
+        """
+        if self._circ_meas_bit_indices == []:
```
richrines1, line 94: nit
```suggestion
        if not self._circ_meas_bit_indices:
```

richrines1 (commented):
does this still work if e.g. one job out of many failed?

qiskit-superstaq/qiskit_superstaq/superstaq_job.py
bharat-thotakura: I originally had everything inline in `result()` but somewhere along the line, I changed it because I didn't want to re-populate `_circ_meas_bit_indices` every time the user called `result()` or something. I've updated the code based on your suggestion now.

bharat-thotakura:
> does this still work if e.g. one job out of many failed?

Well, if a job failed, then presumably there would be no counts returned, so `if counts:` would not be triggered. If you're asking if the user tried to retrieve partial counts from a batch job (with a couple of failed jobs in between), that may be related to https://github.com/Infleqtion/client-superstaq/issues/545.

qiskit-superstaq/qiskit_superstaq/superstaq_job.py
```diff
@@ -74,13 +112,19 @@ def result(self, timeout: Optional[float] = None, wait: float = 5) -> qiskit.res
...
         results = self._wait_for_results(timeout, wait)
+        circ_meas_list: List[List[int]] = []
 
         # create list of result dictionaries
         results_list = []
-        for result in results:
+        for index, result in enumerate(results):
             counts = result["samples"]
             if counts:  # change endianess to match Qiskit
+                num_clbits = self._backend._num_clbits_in_circ[index]
```
richrines1, line 122: one problem i just realized: it's not necessarily the case that `self._backend` is the same backend instance that created the job; for example, as long as i save my job_id, the following should work even is i restart my python shell after submitting the job:
```python
SuperstaqJob(provider.get_backend(target), job_id)`
```
(relatedly, does anything need to change if i create multiple jobs using the same backend before retrieving any of their results?)


given this, afaict it seems like we need to make one of two server-side changes to make this work (unless you see another workaround):
1. we could have the server return the input circuits alongside the compiled ones as part of the job info, so they're always available regardless of how the job was constructed)
2. we could handle more of this server-side by making sure the compiled circuit's clbits are always the same as the input circuits'

imo we should eventually aim to do both of these, though (2) might be a bigger change so it seems reasonable to start with just (1) if that makes sense to you

bharat-thotakura: So yes that's a good catch, this implementation would fail if you were to instantiate `provider.get_backend()` at a later point. And to your other question, I think I overlooked the scenario where multiple jobs are created but the results are retrieved at a later time so that is also something to fix. 

As a client-side fix, we could store `num_clbits` within the circuit metadata itself during the serialization process (which is unaffected during server-side compilation) and then can always be retrieved within `SuperstaqJob` with the `compiled_circuits()` method. And then, after deserialization, this information could be stripped from the circuit metadata into a job attribute so that the altered circuit metadata does not bewilder the user. Other than that workaround, nothing strikes me at the moment.

Though, in the beginning, I did start out wanting to do the (1) suggestion, but stopped after realizing the further server-side changes (and thinking client-side might prove sufficient). However, if it's something that would be implemented eventually anyways, I could definitely start with (1).

richrines1: > As a client-side fix, we could store `num_clbits` within the circuit metadata...

oh cool, as a temporary fix that's pretty nice

> ...if it's something that would be implemented eventually anyways, I could definitely start with (1).

i definitely think it's something we want, but depending on your bandwidth/how much you want to prioritize this imo the fix you outlined above is pretty sufficient for the short term

bharat-thotakura: With the inclusion of the `_get_circuits("input_circuit")` method, `num_clbits_in_circ` is no longer associated with a `SuperstaqBackend` and can be retrieved within a `SuperstaqJob` itself.

bharat-thotakura:
A side note: Should `to_dict()` show the counts with the updated keys?

qiskit-superstaq/qiskit_superstaq/superstaq_backend.py
```diff
@@ -378,7 +378,7 @@ def cq_compile(
             kwargs: Other desired compile options.
 
         Returns:
-            An CQ CompilerOutput object.
+            An CQ `CompilerOutput` object.
```
paaige, line 381: ``A CQ `CompilerOutput` object.``

qiskit-superstaq/qiskit_superstaq/superstaq_job.py
```diff
@@ -56,9 +56,51 @@ def _wait_for_results(self, timeout: float, wait: float = 5) -> List[Dict[str, D
         """
 
         self.wait_for_final_state(timeout, wait)  # should call self.status()
-
         return [self._job_info[job_id] for job_id in self._job_id.split(",")]
 
+    def _arrange_counts(
```
paaige, line 61: we should probably have tests for these helper methods (at the very least a test for _arrange_counts)

qiskit-superstaq/qiskit_superstaq/superstaq_backend.py
bharat-thotakura: Corrected.

qiskit-superstaq/qiskit_superstaq/superstaq_job.py
bharat-thotakura: Added some more tests with [474d162](https://github.com/Infleqtion/client-superstaq/pull/636/commits/474d1628ed419ca79f838ee083b0926aff9f9979)

bharat-thotakura:
One issue I just realized is that the current implementation's use of the `num_clbits` attribute can cause an unexpected discrepancy if there is more than 1 classical register. That is, Qiskit separates the bits between classical registers with a space (see https://github.com/Qiskit/qiskit-terra/blob/e6c431e30682fe7112e2a7079e520383baa982ec/qiskit/result/result.py#L252C1-L257C72), so we may have to account for that by identifying which register the measurements are stored in.

qiskit-superstaq/qiskit_superstaq/superstaq_job.py
```diff
@@ -54,10 +54,55 @@ def _wait_for_results(self, timeout: float, wait: float = 5) -> List[Dict[str, D
...
+
+        Args:
+            index: The index of the circuit to get indices from.
+
+        Returns:
+            The specific measurement indices of the circuit with label `index` in
+            the job.
+        """
+        compiled_circuit = self._get_circuits("compiled_circuit", index=index)
+        return qss.compiler_output.measured_qubit_indices(compiled_circuit)
```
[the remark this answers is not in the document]
bharat-thotakura: Thanks for the catch, you're right this doesn't match the keys when `len(qiskit.ClassicalRegister())` > `len(qiskit.QuantumRegister())` . Will correct this by adding another function to `compiler_output`.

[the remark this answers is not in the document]
bharat-thotakura: Update: I've added the suggested changes and added the example circuit you mentioned as a test in [b29965b](https://github.com/Infleqtion/client-superstaq/pull/636/commits/b29965bb596384763ebeb5f4ca161bf2e4de293d)

qiskit-superstaq/qiskit_superstaq/superstaq_job.py
```diff
@@ -54,10 +54,55 @@ def _wait_for_results(self, timeout: float, wait: float = 5) -> List[Dict[str, D
...
+        arranged_counts = {}
+        for key in counts.keys():
+            updated_key = "0" * num_clbits
+            for counter, index in enumerate(circ_meas_bit_indices):
+                updated_key = updated_key[:index] + key[counter] + updated_key[index + 1 :]
+            arranged_counts[updated_key] = counts[key]
+
+        return arranged_counts
+
+    def _get_meas_bit_info(self, index: int) -> List[int]:
```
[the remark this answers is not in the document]
bharat-thotakura: Done with [694f0b9](https://github.com/Infleqtion/client-superstaq/pull/636/commits/694f0b9e7cb26aea921573d4c2b88635de18dc12)

qiskit-superstaq/qiskit_superstaq/superstaq_job.py
```diff
@@ -54,10 +54,55 @@ def _wait_for_results(self, timeout: float, wait: float = 5) -> List[Dict[str, D
...
+        """Helper method to update the measurement indices from the compiled circuit.
+
+        Args:
+            index: The index of the compiled circuit to get indices from.
+
+        Returns:
+            The specific measurement indices of the circuit with label `index` in
+            the job.
+        """
+        input_circuit = self._get_circuits("input_circuit", index=index)
```
richrines1, line 92: could this just be
```suggestion
        input_circuit = self.input_circuits(index)
```
?

qiskit-superstaq/qiskit_superstaq/superstaq_job.py
```diff
@@ -54,10 +54,55 @@ def _wait_for_results(self, timeout: float, wait: float = 5) -> List[Dict[str, D
...
+    def _get_num_clbits(self, index: int) -> int:
+        """Helper to get number of classical bits in the classical register of the input circuit.
+
+        Args:
+            index: The index of the circuit to get the classical bits from.
+
+        Returns:
+            The number of classical bits for the circuit in the job.
+        """
+        return self._get_circuits("input_circuit", index=index).num_clbits
```
richrines1, line 104: ditto
```suggestion
        return self.input_circuits(index).num_clbits
```

qiskit-superstaq/qiskit_superstaq/superstaq_job.py
```diff
@@ -54,10 +54,55 @@ def _wait_for_results(self, timeout: float, wait: float = 5) -> List[Dict[str, D
...
+        Args:
+            counts: The raw counts from a job result.
+            circ_meas_bit_indices: The indices of the measured qubits.
+            num_clbits: The number of classical bits for the corresponding job circuit.
+
+        Returns:
+            A dictionary with the updated counts keys.
+        """
+        arranged_counts = {}
+        for key in counts.keys():
```
richrines1, line 74: nit

```suggestion
        for key in counts:
```

qiskit-superstaq/qiskit_superstaq/compiler_output.py
```diff
@@ -59,6 +59,27 @@ def measured_qubit_indices(circuit: qiskit.QuantumCircuit) -> List[int]:
     return sorted(circuit.find_bit(qubit).index for qubit in measured_qubits)
 
 
+def measured_clbit_indices(circuit: qiskit.QuantumCircuit) -> List[int]:
```
richrines1, line 62: i think this should recurse into .definitions like measured_qubit_indices does - input circuits could contain custom instructions which include measurements

bharat-thotakura: Thanks, added in [9356a04](https://github.com/Infleqtion/client-superstaq/pull/636/commits/9356a041571c7605f4d9292404ed0a5d07e7cca5) with a basic test for it in [735bb5d](https://github.com/Infleqtion/client-superstaq/pull/636/commits/735bb5dab9ada125d3b6fa5e7e702728a4d4bee0)

richrines1 (commented):
this lgtm, but #846 probably needs to be fixed before merging

bharat-thotakura:
@richrines1, #846 has been resolved by https://github.com/Infleqtion/server-superstaq/pull/2994. Let me know if this PR is good to merge now

PR #638, opened 2023-07-25
Add direct fidelity estimation functions

epelaaez:
to-do:
- [x] add function to general superstaq
- [x] add function to cirq superstaq 
- [x] add function to qiskit superstaq
- [x] add integration test


epelaaez:
cirq superstaq:

```python
import cirq_superstaq as css
import cirq

service = css.Service(remote_host="https://127.0.0.1:5000/")
circuit = cirq.Circuit(cirq.H(cirq.q(0)))
target = "ss_unconstrained_simulator"

ids = service.submit_dfe(
    (circuit, target),
    (circuit, target),
    20,
    1000,
)
result = service.process_dfe(ids)

print(result)
> 0.9984440443287513
```

epelaaez:
qiskit superstaq:
```python
import qiskit

provider = qss.SuperstaqProvider(remote_host="https://127.0.0.1:5000")
qc = qiskit.QuantumCircuit(1)
qc.h(0)
target = "ss_unconstrained_simulator"

ids = provider.submit_dfe(
    (qc, target),
    (qc, target),
    5,
    1000,
)
result = provider.process_dfe(ids)

print(result)
> 0.9987790258104869
```

epelaaez:
submitted dfe job to `ibmq_belem` to make sure this also works with real devices, will update when jobs come back.

cirq-superstaq/cirq_superstaq/daily_integration_test.py
```diff
@@ -239,6 +239,21 @@ def test_supercheq(service: css.Service) -> None:
     assert fidelities.shape == (32, 32)
 
 
+def test_dfe(service: css.Service) -> None:
+    circuit = cirq.Circuit(cirq.H(cirq.q(0)))
+    target = "ss_unconstrained_simulator"
+    ids = service.submit_dfe(
+        rho_1=(circuit, target),
+        rho_2=(circuit, target),
+        m=5,
```
vtomole, line 248: A bit hard to know what this is based on arg name. How about
```suggestion
        num_of random_bases=5,
```
?

cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -688,6 +688,93 @@ def supercheq(
         fidelities = gss.serialization.deserialize(json_dict["fidelities"])
         return circuits, fidelities
 
+    def submit_dfe(
```
dowusu-antwi, line 691: Maybe for another PR, but could we consider having this return a single Superstaq job object that contains:
- both `job_id` values (maybe a tuple return from `job.job_id`)
- a result object that is populated with the results of `process_dfe`?
Not sure how feasible this is.

epelaaez: To initialize a `css.job.Job` we need to pass a `job_id` that is then referenced in `status()`, `counts()`, etc. We could pass some placeholder string to this and store the actual job ids in another field, but then calls to methods like `counts()` would fail because `job_id` would not be an actual job id. 

I feel like it would make more sense to return a tuple of `css.job.Job`, but users would still not be able to call `counts()` because server-side each job stored with these two job ids is a `MetaJob`, which contains references to the actual jobs returned. 

However, if we allow multiple circuits to be stored by a single job (see: https://github.com/Infleqtion/server-superstaq/pull/2477), then returning a tuple of `css.job.Job` would make the most sense, as the linked job in the database would return the counts for all the circuits it contains when `counts()` is called.

cirq-superstaq/cirq_superstaq/daily_integration_test.py
epelaaez: Done.

bharat-thotakura (approved):
LGTM, but will leave final approval to @vtomole @dowusu-antwi

vtomole (approved):
LGTM

Issue #560, opened 2023-06-23
Standardize notebook comments 


stephanielee9:
Regarding capitalization and ending punctuation in `# comments`

stephanielee9:
@stephanielee9 review the PR and add to dev guide

PR #640, opened 2023-07-25
Standardize notebook comments 

bharat-thotakura:
Addresses #560

### DIFF
--- a/cirq-superstaq/docs/getting_started_cirq_superstaq.ipynb
+++ b/cirq-superstaq/docs/getting_started_cirq_superstaq.ipynb
@@ -186,14 +186,6 @@
     "print(\"Job status: \", job.status())\n",
     "print(job.counts())"
    ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "id": "845f5925",
-   "metadata": {},
-   "outputs": [],
-   "source": []
   }
  ],
  "metadata": {
--- a/cirq-superstaq/docs/ibm_cirq_superstaq.ipynb
+++ b/cirq-superstaq/docs/ibm_cirq_superstaq.ipynb
@@ -205,14 +205,6 @@
     "print(\"Job status: \", job.status())\n",
     "print(job.counts())"
    ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "id": "dcd21c6e",
-   "metadata": {},
-   "outputs": [],
-   "source": []
   }
  ],
  "metadata": {
--- a/docs/source/apps/max_sharpe_ratio_optimization.ipynb
+++ b/docs/source/apps/max_sharpe_ratio_optimization.ipynb
@@ -985,7 +985,7 @@
     "\n",
     "obj_QUBO = (\n",
     "    obj.to_qubo()\n",
-    ")  # Converts our objective function into a QUBOWe then set up the client connection in the following cell."
+    ")  # Converts our objective function into a QUBO. We then set up the client connection in the following cell."
    ]
   },
   {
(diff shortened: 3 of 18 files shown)


docs/source/optimizations/aqt/aqt_qss.ipynb
epelaaez, line 1: I think this comment is supposed to be commented out code, so it should still be lowercase (`import matplotlib`) or fully deleted if that code is not being used

docs/source/optimizations/ibm/ibmq_compile_css.ipynb
epelaaez, line 1: isn't this the same sentence but wrapped because of the 100-character limit? then shouldn't it still be lowercase? (same for `ibmq_compile_qss`)

cirq-superstaq/docs/getting_started_cirq_superstaq.ipynb
```diff
@@ -186,14 +186,6 @@
...
     "print(job.counts())"
    ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "id": "845f5925",
-   "metadata": {},
-   "outputs": [],
-   "source": []
```
dowusu-antwi, line 196: Maybe a nit, but does this need to be deleted?

cirq-superstaq/docs/ibm_cirq_superstaq.ipynb
```diff
@@ -205,14 +205,6 @@
...
     "print(job.counts())"
    ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "id": "dcd21c6e",
-   "metadata": {},
-   "outputs": [],
-   "source": []
```
dowusu-antwi, line 215: Ditto the above.

docs/source/optimizations/aqt/aqt_qss.ipynb
bharat-thotakura: Good catch, fixed!

docs/source/optimizations/ibm/ibmq_compile_css.ipynb
bharat-thotakura: Corrected now.

dowusu-antwi (commented):
Should we always end with a period as well?

bharat-thotakura:
> Should we always end with a period as well?

Looking around, there doesn't seem to be a standard on that. Sometimes if it's a long enough sentence a period is used, but not otherwise (especially not for short phrases). So I left it to the individual notebook to use a colon, period, or just multiple short phrases.

cirq-superstaq/docs/getting_started_cirq_superstaq.ipynb
bharat-thotakura: It was a blank cell at the end which I assumed didn't serve a specific purpose. The other notebooks don't have empty cells like that, so I deleted them to be more standard.

docs/source/apps/max_sharpe_ratio_optimization.ipynb
```diff
@@ -985,7 +985,7 @@
     "\n",
     "obj_QUBO = (\n",
     "    obj.to_qubo()\n",
-    ")  # converts our objective function into a QUBOWe then set up the client connection in the following cell."
+    ")  # Converts our objective function into a QUBOWe then set up the client connection in the following cell."
```
epelaaez, line 988: (precedes this PR) is this supposed to be:
```suggestion
    ")  # Converts our objective function into a QUBO. We then set up the client connection in the following cell."
```
?

bharat-thotakura: Yes, fixed now.

PR #650, opened 2023-07-27
add checks-pyproject.toml to gss package_data

richrines1:
otherwise check/configs_.py will fail if gss was pulled from pypi


perlinm:
Out of curiosity, what does `package-data` actually do?

perlinm (approved):

richrines1:
> Out of curiosity, what does `package-data` actually do?

by default only python files will be copied into the wheel, so e.g. `checks-pyproject.toml` won't be copied into `$PYTHONPATH/site-packages/general_superstaq` when gss is installed from pypi. adding files to package-data tells the build script to consider them as part of the package as well

Issue #644, opened 2023-07-26
Job info populates after querying status


ColinCampbellCQ:
### What is happening?

Jobs that have run on Hilbert appear to have no job info unless you query the job status with something like .result() or .status()

### How can we reproduce the issue?

`job = cq_hilbert_qpu.retrieve_job(job_id=job_id)`
`print(job._job_info)`
>>empty dictionary

`job = cq_hilbert_qpu.retrieve_job(job_id=job_id)`
`job.status()`
`print(job._job_info)`
>>populated dictionary

### What should happen?

The first dictionary should be populated

### Environment

- **Superstaq version**: 
- **general-superstaq version**: 
- **cirq-superstaq version**: 
- **qiskit-superstaq version**: 0.4.6
- **Operating system**: 
- **Python version**: 
- **Any other relevant dependencies**: 

### Any additional context?

This is not a problematic issue or a real bug. This is just something I noticed after being confused when looking at jobs after the fact.

epelaaez:
@vtomole since `_job_info` is intended to be private, would the best solution for this be to add a `get_job_info` (or similar) method that looks into `_job_info` and if it is populated returns it and if not, it populates it and then returns it?

vtomole:
@ColinCampbellCQ Thoughts on @epelaaez 's suggestion?

ColinCampbellCQ:
I agree with @epelaaez. The best solution appears to be a simple `get_job_info` method that checks `_job_info`, populates it if needed, then returns `_job_info`.

epelaaez:
I can work on implementing this tomorrow

bharat-thotakura:
I think most of the information present in `_job_info` should already be accessible with some of the methods. For example, with `job.job_id`, `job.result()`, `job.compiled_circuits()`. Is there particular information that is present in `_job_info()` that is not accessible otherwise? I think we could then implement a specialized method for that info

ColinCampbellCQ:
I guess not. The main motivator for my using _job_info was to access the circuit, but that was included in a recent release. There is pretty much only one other reason, which is that I think when I was using my cq_cloud token for submission, ._job_info would contain both my superstaq job id as well as my coldquanta cloud id. Having both of these is incredibly useful when talking with the hardware team because I need cloud job ids in order to work with them.

bharat-thotakura:
[...]

PR #651, opened 2023-07-27
Add `to_dict` to job classes

epelaaez:
fix https://github.com/Infleqtion/client-superstaq/issues/644

### DIFF
--- a/cirq-superstaq/cirq_superstaq/job.py
+++ b/cirq-superstaq/cirq_superstaq/job.py
@@ -196,9 +196,9 @@ class Job:
         """Refreshes and returns job information.
 
         Note:
-            Whenever possible, users should use the specific methods to retrieve the desired
-            information instead of this method. The contents of `_job` are likely to change pretty
-            frequently and are not guaranteed to be consistent or reliable.
+            The contents of this dictionary are not guaranteed to be consistent over time. Whenever
+            possible, users should use the specific `SuperstaqJob` methods to retrieve the desired
+            job information instead of relying on particular entries in the output of this method.
 
         Returns:
             A dictionary containing updated job information.
--- a/cirq-superstaq/cirq_superstaq/job.py
+++ b/cirq-superstaq/cirq_superstaq/job.py
@@ -197,8 +197,8 @@ class Job:
 
         Note:
             The contents of this dictionary are not guaranteed to be consistent over time. Whenever
-            possible, users should use the specific `SuperstaqJob` methods to retrieve the desired
-            job information instead of relying on particular entries in the output of this method.
+            possible, users should use the specific `Job` methods to retrieve the desired job
+            information instead of relying on particular entries in the output of this method.
 
         Returns:
             A dictionary containing updated job information.
--- a/cirq-superstaq/cirq_superstaq/job.py
+++ b/cirq-superstaq/cirq_superstaq/job.py
@@ -192,6 +192,15 @@ class Job:
         self._check_if_unsuccessful()
         return self._job["samples"]
 
+    def to_dict(self) -> Dict[str, Any]:
+        """Refreshes and returns job information.
+
+        Returns:
+            A dictionary containing updated job information.
+        """
+        self._refresh_job()
+        return self._job
+
     def __str__(self) -> str:
         return f"Job with job_id={self.job_id()}"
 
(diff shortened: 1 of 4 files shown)


qiskit-superstaq/qiskit_superstaq/superstaq_job.py
```diff
@@ -196,3 +196,13 @@ def submit(self) -> None:
...
+
+    def to_dict(self) -> Dict[str, Any]:
+        """Refreshes and returns job information.
+
+        Returns:
+            A dictionary containing updated job information.
+        """
+        if self._overall_status not in self.TERMINAL_STATES:
+            self._refresh_job()
+        return self._job_info
```
bharat-thotakura, line 208: The `to_dict` method will show the compiled circuits in their serialized form in qss, whereas the css version is deserialized (though it's a list description of the moments, etc) -- which I think the user would not be interested in.

I wonder if we should parse the dictionary first, and return a new dictionary with user-relevant info (to avoid things like the 'data' and 'samples' keys repeating the same dictionary). Open to other thoughts here.

epelaaez: > I wonder if we should parse the dictionary first, and return a new dictionary with user-relevant info

I like this idea. Should we not return circuits at all? Then, should the returned info be:

```python
{
    "job_id": job_id,
    "num_qubits": 2,
    "samples": {"11": 1},
    "shots": 1,
    "status": "Done",
    "target": "ss_unconstrained_simulator",
}
```
? Or what else should we include/remove?

richrines1: i feel like the Job object itself is the "parsed" dictionary - this is kind of a workaround for stuff we don't have a specific function to access yet

i definitely see the argument for deserializing the circuits specifically, though on the other hand it also might be useful to keep as is so that the user could save the dictionary to a text file and then load it again to recover the job info (we might even want to add a `from_dict()` classmethod in the future)

bharat-thotakura: If there is some benefit for the user to save the dictionary (with the serialized circuits), and further if we plan to implement a `from_dict` sometime later, then I think the current implementation is fine.

bharat-thotakura (approved):
LGTM (w/ discussion)

qiskit-superstaq/qiskit_superstaq/superstaq_job.py
richrines1: (fwiw i'm also hesitant to make this function very user-friendly - we want users to use the actual methods instead of this dictionary whenever possible. the contents of `_job_info` itself will likely to change pretty frequently, and i don't think we should do anything to imply the output of this function will be any more consistent/reliable)

epelaaez: > and i don't think we should do anything to imply the output of this function will be any more consistent/reliable

should we add a note in the docstrings saying this?

richrines1: ah yeah great idea

epelaaez: added note, let me know if it looks good

cirq-superstaq/cirq_superstaq/job.py
```diff
@@ -192,6 +192,20 @@ def counts(self, timeout_seconds: int = 7200, polling_seconds: float = 1.0) -> D
         self._check_if_unsuccessful()
         return self._job["samples"]
 
+    def to_dict(self) -> Dict[str, Any]:
+        """Refreshes and returns job information.
+
+        Note:
+            Whenever possible, users should use the specific methods to retrieve the desired
+            information instead of this method. The contents of `_job` are likely to change pretty
+            frequently and are not guaranteed to be consistent or reliable.
```
richrines1, line 201: nit rewording
```suggestion
            The contents of this dictionary are not guaranteed to be consistent over time. Whenever
            possible, users should use the specific `SuperstaqJob` methods to retrieve the desired
            job information instead of relying on particular entries in the output of this method.
```

richrines1 (approved):
nit rewording suggestion, otherwise lgtm!

qiskit-superstaq/qiskit_superstaq/superstaq_job.py
richrines1: (we can also revisit this discussion if/when a user requests something more specific)

PR #652, opened 2023-07-27
pass config file path to black

richrines1:
by default black will look for config files in the common parent directory of all the files it's checking. this causes `check/format_.py <files>` (or `check/format_.py -i`) to break if all the (modified) files were in the same subdirectory, because e.g. it will use `cirq-superstaq/pyproject.toml` instead of `./pyproject.toml`


vtomole (approved):
LGTM

Issue #649, opened 2023-07-27
Client Superstaq animation refers to `result` object that doesn't exist.


pgoiporia:
### What is happening?

The last line of the animation is `print(result.get_counts())`. However, there is no result object. 

### How can we reproduce the issue?

n/a

### What should happen?

The following would fix it.

```
result = job.result()
print(result.get_counts())
```
or
```
print(job.result().get_counts())
```

### Environment

Any

PR #659, opened 2023-07-28
Update readme animation

stephanielee9:
Closes https://github.com/Infleqtion/client-superstaq/issues/649

### DIFF
--- a/docs/source/_static/svg/code.svg
+++ b/docs/source/_static/svg/code.svg
@@ -148,7 +148,7 @@
 
       @keyframes type-line-11 {
         from {width: 0;}
-        to {width: 518px;}
+        to {width: 600px;}
       }
 
       .line-11 {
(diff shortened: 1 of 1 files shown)


vtomole (approved):

richrines1:
out of curiosity is there a reason we don't show the output of the last line?

vtomole:
>  i guess it would be a bit misleading because it would show the results coming back nearly instantly..

This is fine cause the passed in backend is a simulator. I always thought we don't do that cause we don't want to have too much info in the gif.

stephanielee9:
@vtomole @richrines1 just let me know if you prefer it - i can add it as a comment in the bottom line

vtomole:
This is a huge nit to me so I'm fine with what we have.

richrines1:
> I always thought we don't do that cause we don't want to have too much info in the gif.

i don't think it's too much to show something like `{"00": 47, "11": 53}`, and imo it seems a little incomplete without it (tbh i thought my browser was hesitating or something when i first watched it)

(not necessarily related to this pr though)

stephanielee9:
@vtomole @richrines1 added the counts as a last line in the comments

vtomole (approved):
Noice  :+1:

richrines1 (approved):
awesome :) thanks!

stephanielee9:
Yeah of course! Also not sure where this comment should really go but the files changed makes it look like I'm making a new image (e.g., like I'm making something in Photoshop and dropping in into the folder), but I'm just updating a few lines of code in the svg file in case people want to make updates in the future

Issue #342, opened 2022-09-23
Don't hardcode `backend.configuration()`


vtomole:
Related: https://github.com/SupertechLabs/superstaq-client/issues/150 https://github.com/SupertechLabs/superstaq-client/issues/151

dowusu-antwi:
IIRC, this issue is about parity with `qiskit` so `qiskit`-familiar users like Colin won't experience whiplash, i.e.,
```python
provider = qss.SuperstaqProvider()
hilbert = provider.get_backend("cq_hilbert_qpu")
hilbert.configuration().n_qubits  # gets number of qubits
hilbert.configuration().max_experiments  # gets maximum number of experiments
```
The server-side [`backend_info` issue](https://github.com/Infleqtion/server-superstaq/issues/1116) arose independently to assemble useful device info, possibly including info `qiskit_backend.configuration` does not (e.g., perhaps from `qiskit_backend.properties`)  -- but it looks like that won't be the case, so repurposing `target_info` to populate a client-side version of [`BackendConfiguration`](https://qiskit.org/documentation/stubs/qiskit.providers.models.BackendConfiguration.html) should fix #342.

In retrospect, also having `target_info` makes sense b/c it's a much cleaner representation (particularly for non-qiskit-familiar users, but also in general).

stephanielee9:
Related: https://github.com/Infleqtion/client-superstaq/issues/151, https://github.com/Infleqtion/client-superstaq/issues/150

vtomole:
Started a draft PR in https://github.com/Infleqtion/client-superstaq/pull/666 but it's requiring too many mocks. Thinking about how to accomplish this with less lines of code.

vtomole:
Done by https://github.com/Infleqtion/client-superstaq/pull/666

PR #666, opened 2023-07-29
Populate `backend.configuration_dict` from `target_info` endpoint

vtomole:
Part of https://github.com/Infleqtion/client-superstaq/issues/342


qiskit-superstaq/qiskit_superstaq/superstaq_provider.py
```diff
@@ -462,17 +462,6 @@ def process_dfe(self, ids: List[str]) -> float:
         """
         return self._client.process_dfe(ids)
 
-    def target_info(self, target: str) -> Dict[str, Any]:
```
vtomole, line 465: This is deleted cause it's not how qiskit users are supposed to be getting target info. They should be using `backend.configuration`

qiskit-superstaq/qiskit_superstaq/superstaq_backend_test.py
```diff
@@ -9,21 +12,17 @@
 
 import qiskit_superstaq as qss
 
+if TYPE_CHECKING:
+    from qiskit_superstaq.conftest import MockSuperstaqProvider
 
-def test_default_options() -> None:
```
vtomole, line 13: Unnecessary test: Don't affect coverage with it's removal and doesn't look like a regression test.

vtomole:
Notebooks check won't pass until a new version of server is released.

qiskit-superstaq/qiskit_superstaq/superstaq_backend.py
```diff
@@ -35,21 +35,24 @@ def __init__(self, provider: qss.SuperstaqProvider, target: str) -> None:
...
             "backend_version": "n/a",
-            "n_qubits": -1,
-            "basis_gates": None,
+            "n_qubits": target_info["num_qubits"],
+            "basis_gates": target_info["native_gate_set"],
             "gates": [],
             "local": False,
             "simulator": False,
             "conditional": False,
             "open_pulse": False,
```
bharat-thotakura, line 49: Should something like this also be added?
```suggestion
            "conditional": True if target_info["conditional"] else False ,
            "open_pulse": True if target_info["open_pulse"] else False ,
```

bharat-thotakura (commented):
Partial review

qiskit-superstaq/qiskit_superstaq/superstaq_backend.py
paaige: I think these checks are already taken care of when the target info is initialized in the client

vtomole: Oh i see what's going on

@bharat-thotakura  So the reason to do something like this is when the keys the target info endpoint if returning is different from the ones that the configuration dict wants. For example `"n_qubits": target_info["num_qubits"],`

In the case of your suggestions, `target_info` endpoint already returns the `condition` and `open_pulse` keys which are then updated in `self.configuration_dict.update(target_info)`

vtomole:
@bharat-thotakura I did need to apply your suggestion for the other keys though: https://github.com/Infleqtion/client-superstaq/issues/678. Just pushed a commit to handle undefined keys. Please take a look.

bharat-thotakura (commented):
I tested the changes locally and it seems fine to me. Though, I did notice that `qss.SuperstaqProvider().get_backend(<target>).configuration_dict` yields a couple of duplicate information. Namely, `'backend_name'` <-> `'target'`  and `'n_qubits'` <-> `'num_qubits'` list the same thing. I think this is because of ` self.configuration_dict.update(target_info)`, so it may be worth dropping the duplicate key info.

vtomole:
Good catch. I've removed them from the final dict.

qiskit-superstaq/qiskit_superstaq/superstaq_backend.py
```diff
@@ -35,21 +35,26 @@ def __init__(self, provider: qss.SuperstaqProvider, target: str) -> None:
...
             "memory": False,
-            "max_shots": -1,
+            "max_shots": None,
             "coupling_map": None,
         }
+        target_info.pop("num_qubits", None)
+        target_info.pop("native_gate_set", None)
```
bharat-thotakura, line 55: Should there also be: 
``` 
target_info.pop("target", None)
```
given that `"backend_name": target` is present in the dict, or do we later want to update and list both the 'superstaq name' and 'provider name' -- e.g. "qtm_h1-1_qpu" and "Quantinuum H1-1"

vtomole: Oops. I forgot to pop this one. Done.

bharat-thotakura:
@vtomole No other changes from me :+1:

Issue #641, opened 2023-07-25
Supercheq tutorial missing link to environment variable information


epelaaez:
### What is happening?

Documentation for Supercheq (https://superstaq.readthedocs.io/en/latest/apps/supercheq/Supercheq_css.html; notebook: https://github.com/Infleqtion/client-superstaq/blob/main/docs/source/apps/supercheq/Supercheq_css.ipynb; and same for qiskit-superstaq notebook) says:

> We then supply a Superstaq API token (or key) by either providing the API token as an argument of css.Service() or by setting it as an environment variable (see more details here).

But "here" at the end of this sentence does not link to anywhere.

### How can we reproduce the issue?

Look at documentation site or notebooks in this repo that generate it.

### What should happen?

"here" should link to somewhere, I'm not sure where.

### Environment

Latest.

bharat-thotakura:
The hyperlink in ‘see more details [here](https://docs-superstaq.readthedocs.io/en/latest/get_started/basics.html#Set-up-access-to-Superstaq%E2%80%99s-API)’ for the optimization notebooks seems to be broken as well

bharat-thotakura:
It should link here for the qiskit notebooks for example: https://superstaq.readthedocs.io/en/latest/get_started/basics/basics_qss.html#Set-up-access-to-Superstaq%E2%80%99s-API

PR #667, opened 2023-07-31
Broken 'here' link fix for docs

bharat-thotakura:
Fixes https://github.com/Infleqtion/client-superstaq/issues/641


docs/source/apps/max_sharpe_ratio_optimization.ipynb
```diff
@@ -1271,9 +1271,9 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3 (ipykernel)",
+   "display_name": "Python 3",
    "language": "python",
-   "name": "python"
+   "name": "python3"
```
epelaaez, line 1276: does this affect anything?

epelaaez (approved):
lgtm with question

docs/source/apps/max_sharpe_ratio_optimization.ipynb
bharat-thotakura: Not that I'm aware of. The standard/default metadata name for all the other notebooks is "python3" hence why I updated it for consistency. Shouldn't change much as this will get rewritten with your specific kernel name when running the notebook locally.

Issue #657, opened 2023-07-28
Edit benchmark bot to check for mid-circuit measurement support


epelaaez:
Currently, the benchmarking bot uses a `try; except` block to handle benchmarks with mid-circuit measurement. Now that https://github.com/Infleqtion/client-superstaq/issues/550 was addressed, we can avoid this and check if a device supports mid-circuit measurement before submitting to it.

PR #672, opened 2023-08-02
Check target info for mid-circuit measurement support in benchmark runs

epelaaez:
close https://github.com/Infleqtion/client-superstaq/issues/657

### DIFF
--- a/supermarq-benchmarks/supermarq/run_benchmarks.py
+++ b/supermarq-benchmarks/supermarq/run_benchmarks.py
@@ -45,21 +45,24 @@ for target in qpu_targets:
         if label == "bitcode3" and not target_info.get("supports_midcircuit_measurement"):
             continue
 
-        circuit = benchmark.circuit()
-        if isinstance(circuit, cirq.Circuit):
-            job = service.create_job(
-                circuit,
-                repetitions=1000,
-                target=target,
-                tag=tag,
-                lifespan=3653,
-            )
-        else:
-            for idx, c in enumerate(circuit):
+        try:
+            circuit = benchmark.circuit()
+            if isinstance(circuit, cirq.Circuit):
                 job = service.create_job(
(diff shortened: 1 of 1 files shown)


supermarq-benchmarks/supermarq/run_benchmarks.py
```diff
@@ -44,29 +41,25 @@ def get_qpu_targets(targets: Dict[str, List[str]]) -> List[str]:
         date = datetime.today().strftime("%Y-%m-%d")
         tag = f"{date}-{label}-{target}"
 
-        # Since there is no way to determine if a target supports mid-circuit measurement purely
-        # through `service.target_info()`, we use a `try` block to catch when we submit an invalid
-        # circuit to a device that does not support it.
-        # Related issue: https://github.com/Infleqtion/client-superstaq/issues/550.
+        target_info = service.target_info(target)
+        if label == "bitcode3" and not target_info.get("supports_midcircuit_measurement", False):
```
richrines1, line 45: nit
```suggestion
        if label == "bitcode3" and not target_info.get("supports_midcircuit_measurement"):
```

supermarq-benchmarks/supermarq/run_benchmarks.py
```diff
@@ -44,29 +41,25 @@ def get_qpu_targets(targets: Dict[str, List[str]]) -> List[str]:
...
+            continue
 
-        try:
```
richrines1, line 52: should we still wrap this in a `try..except` block, so if it fails for a target because e.g. some provider is down the rest of the benchmarks would still run?

epelaaez: ah yes, that makes sense, will add that back in.

richrines1 (approved):
lgtm

Issue #675, opened 2023-08-03
Add build docs check


stephanielee9:
`make html SPHINXOPTS="-W"` to treat warnings as errors

stephanielee9:
adding the build_docs check from https://github.com/Infleqtion/server-superstaq/blob/main/.github/workflows/ci.yml to client

epelaaez:
Added check, but not sure if this should be closed as warnings are not being treated as errors yet

PR #679, opened 2023-08-07
Add build docs check

epelaaez:
close https://github.com/Infleqtion/client-superstaq/issues/675


epelaaez:
There are a lot of warnings that pop up with this check, and if we set the warnings to be treated as errors (https://github.com/Infleqtion/client-superstaq/issues/675#issue-1835308520) it would fail. Should fixing these warnings be part of this PR?

.github/workflows/ci.yml
```diff
@@ -13,6 +13,29 @@ on:
...
+        with:
+          python-version: '3.11'
+      - uses: actions/cache@v3
+        with:
+          path: ${{ env.pythonLocation }}
+          key: ${{ env.pythonLocation }}-${{ hashFiles('**/pyproject.toml', '**/*requirements.txt') }}
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r ./docs/requirements.txt
```
richrines1, line 33: i think we need to make sure the local versions of the *staq packages are installed instead of the ones on pypi (might be a more efficient way to do this)
```suggestion
          pip install -e ./general-superstaq[dev] -e ./cirq-superstaq[dev] -e ./qiskit-superstaq[dev] -e ./supermarq-benchmarks[dev]
          pip install -r ./docs/requirements.txt
```

richrines1: also might be worth updating docs/requirements.txt - it seems like a few of the requirements are out of date/redundant/conflict with requirements we use elsewhere. definitely double-check, but i think this is sufficient:
```
cirq-superstaq[dev]~=0.4.10
ipywidgets>=8.0.0
nbsphinx>=0.9.0
qiskit-superstaq[dev]~=0.4.10
sphinx-rtd-theme>=1.0.0
supermarq[dev]~=0.4.10
```

.github/workflows/ci.yml
```diff
@@ -13,6 +13,29 @@ on:
   workflow_call: null
 
 jobs:
+  docs:
```
richrines1, line 16: nit: can we put this at the bottom?

.github/workflows/ci.yml
```diff
@@ -13,6 +13,29 @@ on:
...
+          python-version: '3.11'
+      - uses: actions/cache@v3
+        with:
+          path: ${{ env.pythonLocation }}
+          key: ${{ env.pythonLocation }}-${{ hashFiles('**/pyproject.toml', '**/*requirements.txt') }}
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r ./docs/requirements.txt
+          sudo apt-get install pandoc
```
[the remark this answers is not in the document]
epelaaez: Not sure, I got this from the server-side equivalent check ([here](https://github.com/Infleqtion/server-superstaq/blob/main/.github/workflows/ci.yml#L37))

.github/workflows/ci.yml
```diff
@@ -259,3 +259,27 @@ jobs:
...
+  docs:
+    name: Build docs
+    runs-on: ubuntu-latest
+    timeout-minutes: 7
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python 3.11
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.11'
```
epelaaez, line 272: Should this check use 3.11 like the server-side equivalent, or 3.8 like the other checks in this file?

.github/workflows/ci.yml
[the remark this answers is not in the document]
richrines1: can we try it without the sudo just to find out?

.github/workflows/ci.yml
vtomole: `3.11`

.github/workflows/ci.yml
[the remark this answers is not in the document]
epelaaez: didn't work: 

```
WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

E: Could not open lock file /var/lib/dpkg/lock-frontend - open (13: Permission denied)
E: Unable to acquire the dpkg frontend lock (/var/lib/dpkg/lock-frontend), are you root?
```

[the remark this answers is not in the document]
richrines1: ah damn. thanks for checking!

richrines1 (approved):
lgtm!

PR #681, opened 2023-08-08
Finding lighter-weight way to send QUBO to server

### LINKED ISSUE #143: Find a simpler way to send QUBO to server
https://github.com/SupertechLabs/applications-superstaq/blob/374db4248901c6ebef3773a388d00bf833bb869b/applications_superstaq/qubo.py#L19

dhirpalshah:
Closes #143 

The goal of this PR is to find a way to serialize QUBOs so that we don't need to be calling `convert_qubo_to_model` in `superstaq_client.py`. Instead, we can just call `list(qubo.items())` to convert the given qubo to a list, then in `server-superstaq/superstaq/request_converter.py`, iterate through the list and convert it to a dict that can be used.


dhirpalshah:
I think I've finally been able to get it working by calling `list(qubo.items())`. However, I'm not too sure if this is a considerably lighter-weight way of doing it compared to calling `gss.qubo.convert_qubo_to_model(qubo)`. @vtomole @richrines1; What do you guys think?

dhirpalshah:
I also need to make sure the tests are passing correctly, so I'll do that right now.

dhirpalshah:
Since we have no current use for `gss.qubo.convert_qubo_to_model(qubo)` in `qubo.py`, should I just delete the function and its corresponding test, or is there no harm in leaving it in the repo?

vtomole (commented):
Did you test locally and make sure it works?

paaige:
Doesn't the endpoint also have to be updated since what we're sending over is now different from what the endpoint expects?

paaige:
> Since we have no current use for `gss.qubo.convert_qubo_to_model(qubo)` in `qubo.py`, should I just delete the function and its corresponding test, or is there no harm in leaving it in the repo?

Remove if this works (and after testing whether `read_json_qubo_result` returns an `npt.NDArray[np.int_]` after deserializing as @vtomole said.

dhirpalshah:
> Did you test locally and make sure it works?

I tried running both the dry run and the toshiba target, and they both work in the `max_sharpe_ratio_optimization` notebook.

> Since we have no current use for `gss.qubo.convert_qubo_to_model(qubo)` in `qubo.py`, should I just delete the function and its corresponding test, or is there no harm in leaving it in the repo?

There is the corresponding PR in server!

> Remove if this works (and after testing whether `read_json_qubo_result` returns an `npt.NDArray[np.int_]` after deserializing as @vtomole said.

Currently, `read_json_qubo_result` returns type `list`, so that's something I'll have to go fix.

paaige:
> > Did you test locally and make sure it works?
> 
> I tried running both the dry run and the toshiba target, and they both work in the `max_sharpe_ratio_optimization` notebook.
> 
> > Since we have no current use for `gss.qubo.convert_qubo_to_model(qubo)` in `qubo.py`, should I just delete the function and its corresponding test, or is there no harm in leaving it in the repo?
> 
> There is the corresponding PR in server!
> 
> > Remove if this works (and after testing whether `read_json_qubo_result` returns an `npt.NDArray[np.int_]` after deserializing as @vtomole said.
> 
> Currently, `read_json_qubo_result` returns type `list`, so that's something I'll have to go fix.

Just to summarize: read_json_qubo_result is outdated (from when we were using qubovert's solver which returns an NDArray) so there's nothing wrong with your code. Just update it to return the type the endpoint returns after deserialization

pnoell:
> > > Did you test locally and make sure it works?
> > 
> > 
> > I tried running both the dry run and the toshiba target, and they both work in the `max_sharpe_ratio_optimization` notebook.
> > > Since we have no current use for `gss.qubo.convert_qubo_to_model(qubo)` in `qubo.py`, should I just delete the function and its corresponding test, or is there no harm in leaving it in the repo?
> > 
> > 
> > There is the corresponding PR in server!
> > > Remove if this works (and after testing whether `read_json_qubo_result` returns an `npt.NDArray[np.int_]` after deserializing as @vtomole said.
> > 
> > 
> > Currently, `read_json_qubo_result` returns type `list`, so that's something I'll have to go fix.
> 
> Just to summarize: read_json_qubo_result is outdated (from when we were using qubovert's solver which returns an NDArray) so there's nothing wrong with your code. Just update it to return the type the endpoint returns after deserialization. We could also potentially get rid of this function entirely but I'm on the fence about exposing the serialization/deserialization methods to users @pnoell (tagging Peter to weigh in)

I might be misunderstanding but this is my take: 

`read_json_qubo_result` currently returns an NDArray after deserialization. However, I would expect `List[Dict[Tuple[int], int]]`. Is that the decision in question? I can't really speak to the implications of exposing serialization methods to users but I would expect that `List` behavior instead of an NDArray

dhirpalshah:
> > > > Did you test locally and make sure it works?
> > > 
> > > 
> > > I tried running both the dry run and the toshiba target, and they both work in the `max_sharpe_ratio_optimization` notebook.
> > > > Since we have no current use for `gss.qubo.convert_qubo_to_model(qubo)` in `qubo.py`, should I just delete the function and its corresponding test, or is there no harm in leaving it in the repo?
> > > 
> > > 
> > > There is the corresponding PR in server!
> > > > Remove if this works (and after testing whether `read_json_qubo_result` returns an `npt.NDArray[np.int_]` after deserializing as @vtomole said.
> > > 
> > > 
> > > Currently, `read_json_qubo_result` returns type `list`, so that's something I'll have to go fix.
> > 
> > 
> > Just to summarize: read_json_qubo_result is outdated (from when we were using qubovert's solver which returns an NDArray) so there's nothing wrong with your code. Just update it to return the type the endpoint returns after deserialization. We could also potentially get rid of this function entirely but I'm on the fence about exposing the serialization/deserialization methods to users @pnoell (tagging Peter to weigh in)
> 
> I might be misunderstanding but this is my take:
> 
> `read_json_qubo_result` currently returns an NDArray after deserialization. However, I would expect `List[Dict[Tuple[int], int]]`. Is that the decision in question? I can't really speak to the implications of exposing serialization methods to users but I would expect that `List` behavior instead of an NDArray

Sounds good - I'll change the expected return type to this: `List[Dict[Tuple[int], int]]`

vtomole (commented):
Deploy the server and then run your notebooks against this branch to make sure things still work.

vtomole (approved):
LGTM

general-superstaq/general_superstaq/qubo.py
```diff
@@ -1,13 +1,11 @@
...
 
-import numpy as np
-import numpy.typing as npt
 import qubovert as qv
 
 import general_superstaq as gss
 
 
-def read_json_qubo_result(json_dict: Dict[str, str]) -> npt.NDArray[np.int_]:
+def read_json_qubo_result(json_dict: Dict[str, str]) -> list[Dict[tuple[int], int]]:
```
paaige, line 8: A lot of your tests are failing because `list[Dict[tuple[int], int]]` should be `List[Dict[Tuple[int], int]]` but you should verify that this is the type returned by the deserialized json dict returned from the server.

dhirpalshah: I made the fix to make it `List` instead of `list` by importing `from typing import List`, but I have yet to make the assertion in the test. Will do that ASAP

dhirpalshah: Done!

PR #687, opened 2023-08-10
Add ACES endpoints

epelaaez:
(related: https://github.com/Infleqtion/server-superstaq/pull/2612)

### DIFF
--- a/qiskit-superstaq/qiskit_superstaq/superstaq_provider.py
+++ b/qiskit-superstaq/qiskit_superstaq/superstaq_provider.py
@@ -469,59 +469,6 @@ class SuperstaqProvider(qiskit.providers.ProviderV1, gss.service.Service):
         """
         return self._client.process_dfe(ids)
 
-    def submit_aces(
-        self,
-        target: str,
-        qubits: List[int],
-        shots: int,
-        num_circuits: int,
-        mirror_depth: int,
-        extra_depth: int,
-        **kwargs: Any,
-    ) -> str:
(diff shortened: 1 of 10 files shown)


qiskit-superstaq/qiskit_superstaq/superstaq_provider.py
```diff
@@ -469,6 +469,59 @@ def process_dfe(self, ids: List[str]) -> float:
         """
         return self._client.process_dfe(ids)
 
+    def submit_aces(
```
richrines1, line 472: it looks like this is the same in css.Service and qss.SuperstaqProvider - should we just move it to `gss.service.Service` instead?

also we should add it to `qss.SuperstaqBackend` (which would just be a wrapper for `self._provider(target=self.name(), ...)`)

epelaaez: > it looks like this is the same in css.Service and qss.SuperstaqProvider - should we just move it to gss.service.Service instead?

sounds good

> also we should add it to qss.SuperstaqBackend (which would just be a wrapper for self._provider(target=self.name(), ...))

will do

epelaaez: done!

richrines1 (approved):
lgtm!

epelaaez:
> lgtm!

cool! will merge once https://github.com/Infleqtion/server-superstaq/pull/2612 goes in

PR #689, opened 2023-08-11
Update DD tutorial notebook 

paaige:
Adds more examples (using various strategies), examples on more complex circuits.


knsmith:
This is looking really great!

This probably doesn't matter, but should we use lower case for strategies? 

When we add results, a nice thing to add would be a little closing paragraph describing the ??x benefit compared to the qiskit baseline!

knsmith:
Todos - 
1) make sure numbering looks nice
2) draw `pulse_gate_circuits`
3) change x axis on plots from ints to bin strings

qiskit-superstaq/requirements.txt
```diff
@@ -1,3 +1,4 @@
 general-superstaq~=0.4.19
 matplotlib~=3.0
 qiskit>=0.43.3
+qiskit-ibm-provider
```
singular-value, line 4: ```suggestion```

singular-value, line 4:
```suggestion
qiskit>=0.43.3
```

singular-value (approved):

PR #690, opened 2023-08-11
Delete `convert_qubo_to_model` in `qubo.py`

dhirpalshah:
In `qubo.py`, the `convert_qubo_to_model` function previously was used to serialize a qubo before sending it in as a request. However, after finishing #681, there is no need for this function anymore - removed for cleanliness.


bharat-thotakura (approved):
LGTM

PR #693, opened 2023-08-12
allow newest black version

richrines1:
(newer versions don't allow empty lines at the top of function definitions, so various files had to be reformatted)


epelaaez (approved):
lgtm!

Issue #523, opened 2023-06-14
Document keyword arguments that matter


stephanielee9:
Document options_dict

stephanielee9:
Document `cq_token`

stephanielee9:
`cz_radius` and maybe also `cz_rads` and maybe rename @SalahedeenIssa

stephanielee9:
make `cz_rads` `stripped_cz_rads`

PR #699, opened 2023-08-15
Add noise options to ACES

epelaaez:
Adds the ability to pass noise models to the ACES endpoints to allow for more interesting simulations.

(related: https://github.com/Infleqtion/server-superstaq/pull/2634)


general-superstaq/general_superstaq/service.py
```diff
@@ -318,6 +318,15 @@ def submit_aces(
                 - lifespan: How long to store the jobs submitted for in days (only works with right
                 permissions).
                 - method: Which type of method to execute the circuits with.
+                - noise: A tuple indicating a noise model to simulate the run with. The tuple will
+                    have the form ("channel_name", prob_of_error). The permitted channels are:
+                    "symmetric_depolarize", "phase_flip", "bit_flip" and "asymmetric_depolarize".
+
+                    For "asymmetric_depolarize", `prob_of_error` will be a three-tuple with the
+                    error rates for the X, Y, Z gates in that order. So, a valid argument would be
+                    `noise = ("asymmetric_depolarize", (0.1, 0.1, 0.1))`. Notice that this values
```
bharat-thotakura, line 327: nit
```suggestion
                    `noise = ("asymmetric_depolarize", (0.1, 0.1, 0.1))`. Notice that these values
```
and the two other occurrences

epelaaez: done!

general-superstaq/general_superstaq/service.py
```diff
@@ -318,6 +318,15 @@ def submit_aces(
                 - lifespan: How long to store the jobs submitted for in days (only works with right
                 permissions).
                 - method: Which type of method to execute the circuits with.
+                - noise: A tuple indicating a noise model to simulate the run with. The tuple will
+                    have the form ("channel_name", prob_of_error). The permitted channels are:
```
richrines1, line 322: * imo the noise model and error probability should be two separate arguments, instead of having to be passed as a tuple
* `cirq.NoiseModel`s are json-serializable - would it make sense to allow them to be passed as well?

also - predates this pr, but all the parameters we're documenting here they should really be made into optional arguments instead of using kwargs

vtomole: > cirq.NoiseModels are json-serializable - would it make sense to allow them to be passed as well?

Makes sense. This will have to be hacky on the qiskit side though.

epelaaez: > imo the noise model and error probability should be two separate arguments, instead of having to be passed as a tuple

Sounds good.

> cirq.NoiseModels are json-serializable - would it make sense to allow them to be passed as well?

I was thinking about doing this, but I don't think there's an exact equivalence between Qiskit's and Cirq's models, so it isn't as easy to convert back and forth. Or we could allow `cirq.NoiseModel` coming from css and keep it as it is for qss.

> predates this pr, but all the parameters we're documenting here they should really be made into optional arguments instead of using kwargs

I'll change them to optional parameters.

bharat-thotakura: > Or we could allow cirq.NoiseModel coming from css and keep it as it is for qss.

Can we not use `NoiseModel.to_dict()` and `from_dict()` for qss?

epelaaez: > Can we not use NoiseModel.to_dict() and from_dict() for qss?

Ah yeah, I think this could work.

epelaaez: For reference, I think the best way to translate from Qiskit's to Cirq's noise models is to use [`Kraus(qiskit_model)`](https://qiskit.org/documentation/stubs/qiskit.quantum_info.Kraus.html) and then use [`cirq.KrausChannel`](https://quantumai.google/reference/python/cirq/KrausChannel) to go back and forth. Let me know if you think there is a better way to do this.

richrines1: > (even tag and lifespan? these two are also optional when creating jobs)

imo they should be optional arguments in `create_job` too, if they are intended for external use - if nothing else it's a form of documentation (part of #523)

epelaaez: > imo they should be optional arguments in create_job too

sounds good! I'll make the change for this function for now.

richrines1: > use [`cirq.KrausChannel`](https://quantumai.google/reference/python/cirq/KrausChannel) to go back and forth

`qss` shouldn't be given a `cirq` dependency, unfortunately

i kinda think it's okay for parity to be imperfect here though (or if nothing else it seems preferable to keeping a useful functionality out of one or the other client just for the sake of parity). We'll still be able to specify a noise model in `qss` using the string method. I imagine that we'll also be making our error models more general/customizable, which would hopefully expand this to cover a lot more use cases

we can also add an issue to revisit this in the future

epelaaez: > qss shouldn't be given a cirq dependency, unfortunately

Oh, I meant server-side, sorry I wasn't specific. In the client I think it's possible to let the users send a serialized `cirq.NoiseModel` and a `qiskit_aer.noise.QuantumError`.

epelaaez: It looks like [`qiskit_aer.noise.NoiseModel.from_dict`](https://qiskit.org/ecosystem/aer/_modules/qiskit_aer/noise/noise_model.html#NoiseModel.from_dict) is being deprecated, and I don't think there's another way to reconstruct a Qiskit noise model back in the server.

epelaaez:
I added support to send a `cirq.NoiseModel` through `css.Service.submit_aces`. I wanted to do the same for `qss.SuperstaqBackend`, but the methods in Qiskit to initialize a noise model from a dictionary (which we need to construct the model back in the server) are being deprecated. For now, I've only left the original noise implementation in qss.

Related, now that the functionality is different between css and qss, should `qss.SuperstaqBackend.submit_aces` move to `qss.SuperstaqProvider`. Since, if I understood correctly, the rationale under putting it in `SuperstaqBackend` was that css and qss did the same (original discussion: https://github.com/Infleqtion/client-superstaq/pull/687#discussion_r1291559327, cc: @richrines1).

richrines1:
> I added support to send a `cirq.NoiseModel` through `css.Service.submit_aces`. I wanted to do the same for `qss.SuperstaqBackend`, but the methods in Qiskit to initialize a noise model from a dictionary (which we need to construct the model back in the server) are being deprecated. For now, I've only left the original noise implementation in qss.

ok sounds good. we'll definitely want to generalize this more in the future anyway (e.g. for standard simulations with noise)
 
> Related, now that the functionality is different between css and qss, should `qss.SuperstaqBackend.submit_aces` move to `qss.SuperstaqProvider`

yep i think so

cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -792,6 +792,85 @@ def process_dfe(self, ids: List[str]) -> float:
...
+            ValueError: If the target is not valid.
+            ValueError: If the noise model passed is not valid.
+            SuperstaqServerException: if the request fails.
+        """
+        noise_dict: Dict[str, object] = {}
+        if isinstance(noise, str):
+            noise_dict["type"] = noise
+            noise_dict["params"] = error_prob if isinstance(error_prob, tuple) else (error_prob,)
+        elif isinstance(noise, cirq.NoiseModel):
+            noise_dict["cirq_noise_model"] = cirq.to_json(noise)
```
epelaaez, line 848: So, apparently not **all** `cirq.NoiseModel`s are serializable (see: https://github.com/quantumlib/Cirq/issues/6249). Should I add a `try` statement here to catch when a given noise model is not serializable? Or is the error thrown by Cirq enough?

Error thrown by Cirq:
```
    160 def default(self, o):
    161     """Implement this method in a subclass such that it returns
    162     a serializable object for ``o``, or calls the base implementation
    163     (to raise a ``TypeError``).
   (...)
    177 
    178     """
--> 179     raise TypeError(f'Object of type {o.__class__.__name__} '
    180                     f'is not JSON serializable')

 

TypeError: Object of type InsertionNoiseModel is not JSON serializable

```

qiskit-superstaq/qiskit_superstaq/superstaq_backend.py
```diff
@@ -422,54 +422,3 @@ def resource_estimate(
         if circuits_is_list:
             return resource_estimates
         return resource_estimates[0]
-
-    def submit_aces(
```
richrines1, line 426: why was this removed from `SuperstaqBackend`?

epelaaez: I thought we wanted to move it from `SuperstaqBackend` to `SuperstaqProvider` (https://github.com/Infleqtion/client-superstaq/pull/699#issuecomment-1679753907).

richrines1: ohhh im sorry i totally misread that question.

basically i think we want the methods to appear in \*both\* `qss.SuperstaqProvider` and `qss.SuperstaqBackend`. The reason we only had to put them in `qss.SuperstaqBackend` in #687 was that `SuperstaqProvider` was already inheriting them from `gss.Service`

thinking about this a bit more though - it seems like the `qss` version of this is still independent of qiskit? in which case we might not want to remove it from `gss.Service`; instead we can just:
* leave `submit_aces` in `gss.Service` and `qss.SuperstaqBackend`, but update both include the new/updated arguments
* override `submit_aces` in `css.Service` to add support for cirq noise models
* leave `process_aces` alone

stephanielee9:
copy of https://github.com/Infleqtion/client-superstaq/pull/713

epelaaez:
@vtomole @richrines1 This PR now has the changes that were made in the PR form my fork.

vtomole (approved):
LGTM assuming this has the same contents as https://github.com/Infleqtion/client-superstaq/pull/713

epelaaez:
I'm going to re-check, but I think they do in fact have the same changes. I will also wait for the server-side changes since otherwise the integration test will fail.

qiskit-superstaq/qiskit_superstaq/daily_integration_test.py
```diff
@@ -234,6 +234,22 @@ def test_dfe(provider: qss.superstaq_provider.SuperstaqProvider) -> None:
     assert isinstance(result, float)
 
 
+def test_aces(provider: qss.superstaq_provider.SuperstaqProvider) -> None:
+    backend = provider.get_backend("ss_unconstrained_simulator")
+    job_id = backend.submit_aces(
+        qubits=qiskit.QuantumRegister(1),
```
vtomole, line 240: This throws `FAILED qiskit-superstaq/qiskit_superstaq/daily_integration_test.py::test_aces - TypeError: Object of type QuantumRegister is not JSON serializable` when i run it on the local server. Why is it not 
```suggestion
        qubits=[0],
```

Like the `test_aces` in `cirq-superstaq?

epelaaez: Yes this should be `qubits=[0]`, I didn't update this test, sorry about that. I changed the `qubits` argument from accepting a `qiskit.QuantumRegister` to a `Sequence[int]` since it is harder to specify non-linear qubits with `qiskit.QuantumRegister`.

vtomole:
@epelaaez The server PR that this corresponds to has been merged so you can merge this in as well.

epelaaez:
I don't think I have write access to merge this in, so feel free to do so @vtomole !

PR #701, opened 2023-08-17
Add option to skip checks in check/all_.py

perlinm:
I use the check scripts a lot in `research-superstaq`, but that repo doesn't have the `requirements` or `build_docs` checks, which breaks `check/all_.py`.  As a consequence, I have been running individual checks manually.  I would like the option to just tell `check/all_.py` to skip certain checks.

This PR is a proposed solution to my problem.


general-superstaq/general_superstaq/check/all_.py
```diff
@@ -16,13 +16,29 @@
...
+CHECK_TYPE = Literal[
+    "configs",
+    "format",
+    "flake8",
+    "pylint",
+    "mypy",
+    "coverage",
+    "requirements",
+    "build_docs",
+]
```
perlinm, line 28: Not sure whether this is better to check at "compile time" (mypy) or runtime.  I could just instead define `CHECKS = ["configs", ...]` up here, accept a `skip: Optional[List[str]]` argument in `run` and then throw an error if `not all(entry not in CHECKS for entry in skip)`

richrines1: could we just add `parser.add_argument("--skip", choices=[...], ...)` to the arg parser?

perlinm: Ah nice yeah I'll do that!

perlinm: done

general-superstaq/general_superstaq/check/all_.py
```diff
@@ -50,6 +61,13 @@ def run(*args: str, sphinx_paths: Optional[List[str]] = None) -> int:
         dest="force_all",
         help="'Hard force' ~ continue past (i.e. do not exit after) all failing checks.",
     )
+    parser.add_argument(
+        "--skip",
+        choices=CHECK_LIST,
+        nargs="*",
```
richrines1, line 67:
```suggestion
        nargs="+",
        action="extend",
```

richrines1 (approved):
one more suggestion then lgtm!

general-superstaq/general_superstaq/check/all_.py
perlinm: `action="extend"` makes sense to me, but I think we want [`nargs="*"`](https://docs.python.org/3/library/argparse.html#nargs):
image

perlinm: Oh: trying out `nargs="+"` works just fine though.  What am I not getting?

perlinm: Ah gotcha.  Thanks!

richrines1:
looking into that failing test btw - i'm guessing it has something to do with the version of qiskit that was released an hour ago

richrines1:
> looking into that failing test btw - i'm guessing it has something to do with the version of qiskit that was released an hour ago

should be fixed by #704

perlinm:
Awesome thanks @richrines1!

perlinm:
Oops.  Realized post-merge that this does not work due to the way that we [pass arguments downstream](https://github.com/Infleqtion/client-superstaq/blob/cd783c673a800971ac2b544c6087c5acf43be2d6/general-superstaq/general_superstaq/check/all_.py#L81)...

richrines1:
ah right my bad, should've caught that.

not perfect, but one option is to move the argument declaration to `check_utils.get_file_parser()` so it gets stripped out in all downstream checks too (and then maybe adding something in its `help=".."` indicating it's only relevant to all_.py)

perlinm:
Another thought: `args_to_pass` is currently passed to _**all**_ of `format_`, `flake8_`, `pylint_`, `mypy_`, and `coverage_`.  It seems... unrealistic to expect that they all recognize the same arguments.  Maybe `check/all_.py` shouldn't pass any arguments at all?

richrines1:
would something like this work?

in check/all_.py:
```python
def run(*args, ...):
    ...
    parsed_args, unknown_args = parser.parse_known_args(args)

    check.some_check(*unknown_args, ..., parsed_args=parsed_args)
    check.some_other_check(*unknown_args, ..., parsed_args=parsed_args)
```
in check/some[_other]_check.py:
```python
def run(*args, ..., parsed_args=None):
    ...
    parsed_args = parser.parse_args(args, namespace=parsed_args)
    ...
```

perlinm:
I like the idea of passing `parsed_args` downstream.  Seems like if we do that then we don't need to pass `unknown_args` then, because `parsed_args` should capture all arguments that are common to all checks.  So it would just be (in `check/all_.py`):
```python
def run(*args, ...):
    ...
    parsed_args, _ = parser.parse_known_intermixed_args(args)

    check.some_check(..., parsed_args=parsed_args)
    check.some_other_check(..., parsed_args=parsed_args)
```

PR #704, opened 2023-08-17
use unicode π in more qss.custom_gates strings

richrines1:
(fixes tests when upgrading to qiskit 0.44.1)


bharat-thotakura (approved):
LGTM

PR #709, opened 2023-08-22
pass parsed arguments from `check/all_.py` to individual checks

perlinm:
Follow-up on #701

### DIFF
--- a/general-superstaq/general_superstaq/check/all_.py
+++ b/general-superstaq/general_superstaq/check/all_.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python3
-
 import sys
 import textwrap
 from typing import List, Optional
@@ -78,31 +77,30 @@ def run(*args: str, sphinx_paths: Optional[List[str]] = None) -> int:
     default_mode = not parsed_args.files and parsed_args.revisions is None
     checks_failed = 0
 
-    args_to_pass = [arg for arg in args if arg not in ("-f", "-F", "--force")]
-
     # run formatting checks
     # silence most checks to avoid printing duplicate info about incremental files
     # silencing does not affect warnings and errors
     exit_on_failure = not (parsed_args.force_formats or parsed_args.force_all)
+    common_kwargs = dict(namespace=parsed_args, exit_on_failure=exit_on_failure, silent=True)
     if "configs" not in parsed_args.skip:
-        checks_failed |= configs.run(exit_on_failure=exit_on_failure, silent=True)
+        checks_failed |= configs.run(**common_kwargs)
     if "format" not in parsed_args.skip:
-        checks_failed |= format_.run(*args_to_pass, exit_on_failure=exit_on_failure, silent=True)
+        checks_failed |= format_.run(**common_kwargs)
     if "flake8" not in parsed_args.skip:
-        checks_failed |= flake8_.run(*args_to_pass, exit_on_failure=exit_on_failure, silent=True)
+        checks_failed |= flake8_.run(**common_kwargs)
     if "pylint" not in parsed_args.skip:
-        checks_failed |= pylint_.run(*args_to_pass, exit_on_failure=exit_on_failure, silent=True)
+        checks_failed |= pylint_.run(**common_kwargs)
 
     # run typing and coverage checks
     exit_on_failure = not parsed_args.force_all
     if "mypy" not in parsed_args.skip:
-        checks_failed |= mypy_.run(*args_to_pass, exit_on_failure=exit_on_failure, silent=True)
+        checks_failed |= mypy_.run(**common_kwargs)
     if "coverage" not in parsed_args.skip:
-        checks_failed |= coverage_.run(*args_to_pass, exit_on_failure=exit_on_failure, silent=True)
+        checks_failed |= coverage_.run(**common_kwargs)
 
     # check that all pip requirements files are in order
     if "requirements" not in parsed_args.skip:
-        checks_failed |= requirements.run(exit_on_failure=exit_on_failure)
+        checks_failed |= requirements.run(namespace=parsed_args, exit_on_failure=exit_on_failure)
 
     if default_mode and "build_docs" not in parsed_args.skip:
         # checks that the docs build
--- a/general-superstaq/general_superstaq/check/check_utils.py
+++ b/general-superstaq/general_superstaq/check/check_utils.py
@@ -1,8 +1,4 @@
-#!/usr/bin/env python3
-
-"""Dumping ground for check script utilities.
-"""
-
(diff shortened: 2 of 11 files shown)


general-superstaq/general_superstaq/check/check_utils.py
```diff
@@ -298,7 +294,7 @@ def get_file_parser() -> argparse.ArgumentParser:
     )
 
     help_text = "The files to check. If not passed any files, inspects the entire repo."
-    parser.add_argument("files", nargs="*", help=help_text)
+    parser.add_argument("--files", nargs="*", default=[], help=help_text)
```
perlinm, line 297: This appeared to be necessary to get rid of warnings like this:
image

richrines1: this changes our usage pretty drastically - to pass files we'll have to use e.g. `./check/pytest_.py --files x.py ...` instead of  `./check/pytest_.py x.py ...`

it seems like this warning only comes up when using `parse_known_intermixed_args` - can we just use `parse_known_args` everywhere instead? (and also add `action="extend"` in each case, to prevent sub-parsers from clobbering `parsed_args.files`)

richrines1 (commented):
just occurred to me there's a somewhat simpler workaround for this - we could have just added the `--skip` argument in `check_utils.get_file_parser` (so it's present in all file parsers), and then e.g. in `gss.check.coverage_.py` add a line like
```python
if "coverage" in parsed_args.skip:
    return 0
```
(and revert the rest of the changes here and in #701)

richrines1:
also out of curiosity (definitely outside the scope of this pr) - would it be unreasonable to include the superset of arguments from \*all\* checks in `check_utils.get_file_parser()`, and just ignore arguments when they aren't necessary for a particular check? For example it seems pretty reasonable for `check/all_.py --enable-socket` to run everything normally, with `parsed_args.enable_socket` only changing the behavior of coverage. Then we'd never have to worry about conflicts between the parsers or stripping arguments out of `args` before passing them downstream. We could even handle `--force_*` in the individual checks instead of in `all_` - e.g. `flake8_.py` could use `parsed_args.force_formats` to determine whether to return or exit on failure

general-superstaq/general_superstaq/check/check_utils.py
perlinm: Oh right duh.  Sorry that was dumb :sweat_smile:

perlinm: I like adding `--skip` to the common parser.

perlinm:
Great ideas @richrines1!

Based on the description/explanation [here](https://docs.python.org/3/library/argparse.html#intermixed-parsing), I think I like using `parse_known_intermixed_args`, but I definitely like making `--skip` common to everything.

With regards to using a superset of common arguments, that (i) seems to not make sense with e.g. the `-a/--all` argument of `pylint_.py`, or with the `--notebook` and `--integration` arguments of `pytest_.py`, and (ii) it would make help text (`-h/--help`) less helpful.

general-superstaq/general_superstaq/check/coverage_.py
```diff
@@ -13,6 +13,7 @@ def run(
     *args: str,
     include: Union[str, Iterable[str]] = "*.py",
     exclude: Union[str, Iterable[str]] = "*_integration_test.py",
+    namespace: argparse.Namespace = argparse.Namespace(),
```
richrines1, line 16: is there any reason to pass this around anymore?

if so, this should just be
```suggestion
    namespace: Optional[argparse.Namespace] = None,
```

perlinm: Oh nope, my mistake.

perlinm: removed

general-superstaq/general_superstaq/check/check_utils.py
```diff
@@ -283,12 +279,31 @@ def get_test_files(
...
-def get_file_parser() -> argparse.ArgumentParser:
-    """Construct an console argument parser that identifies files to check.
+    This parser collects a list of files to check, with "no files" == "all files".
 
-    The parser also has a flag to do incremental cheks, i.e., on the subset files that have changed
-    since a specified git revision, as well as a flag to exclude files matching a specified glob.
+    In addition, this parser has flags/arguments to:
+    - skip certain (specified) checks.
+    - run incremental checks, i.e., on the files that have changed since a specified git revision,
+    - exclude files matching a specified glob,
```
richrines1, line 303:
```suggestion
    - exclude files matching a specified glob.
```

general-superstaq/general_superstaq/check/check_utils.py
```diff
@@ -283,12 +279,31 @@ def get_test_files(
...
+
+    Args:
+        no_files: ignore file-related arguments.
```
richrines1, line 306:
```suggestion
        no_files: Ignore file-related arguments.
```

general-superstaq/general_superstaq/check/all_.py
```diff
@@ -50,6 +38,7 @@ def run(*args: str, sphinx_paths: Optional[List[str]] = None) -> int:
 
     parser.add_argument(
         "-f",
+        "--soft-force",
         action="store_true",
         dest="force_formats",
```
richrines1, line 43: nit: for consistency can we get rid of this line, and either use `"--force-formats"` above or `parsed_args.soft_force` below?
```suggestion
```

richrines1 (commented):
a few nits then lgtm!

general-superstaq/general_superstaq/check/all_.py
perlinm: I made it `--force-formats` and `--force-all`.

general-superstaq/general_superstaq/check/all_.py
```diff
@@ -78,35 +58,30 @@ def run(*args: str, sphinx_paths: Optional[List[str]] = None) -> int:
     default_mode = not parsed_args.files and parsed_args.revisions is None
     checks_failed = 0
 
-    args_to_pass = [arg for arg in args if arg not in ("-f", "-F", "--force")]
+    args_to_pass = [arg for arg in args if arg not in ("-f", "--soft-force", "-F", "--force")]
```
richrines1, line 61:
```suggestion
    args_to_pass = [arg for arg in args if arg not in ("-f", "--force-formats", "-F", "--force")]
```

richrines1 (approved):
lgtm % one more correction:

perlinm:
@richrines1 thanks for the careful review!  And sorry for my sloppiness haha

PR #710, opened 2023-08-23
diagnostic qubits

SalahedeenIssa:
Gives client users the option to submit diagnostic qubits with their circuits, once it is merged in server side


general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -715,3 +715,44 @@ def find_api_key() -> str:
...
+        return "".join([bitstring[index] for index in indices])
+
+    for key, val in counts_dict.items():
+        target_key = meas_substring(key, target_qubit_indices)
+        circuit_key = meas_substring(key, circuit_qubit_indices)
+        for new_key, new_dict in zip((target_key, circuit_key), (target_dict, circuit_dict)):
+            if new_key in new_dict.keys():
+                new_dict[new_key] += val
+            else:
+                new_dict[new_key] = val
```
perlinm, line 756: If you initialize above with
```python
    target_dict = collections.defaultdict(int)
    circuit_dict = collections.defaultdict(int)
```
then here you can just to
```suggestion
            new_dict[new_key] += val
```

general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -715,3 +715,44 @@ def find_api_key() -> str:
         "Try passing an 'api_key' variable, or setting your API key in the command line "
         "with SUPERSTAQ_API_KEY=..."
     )
+
+
+def get_counts_on_qubits(
```
perlinm, line 720: At a high level, my biggest thoughts here are that this method should either
1. only return the counts on the requested qubits, or
2. be called something like `split_counts`.

I'm slightly leaning in favor of (1), because imo there isn't much gained from getting both sets of counts at the same time.

qiskit-superstaq/qiskit_superstaq/superstaq_job.py
```diff
@@ -101,6 +101,17 @@ def result(self, timeout: Optional[float] = None, wait: float = 5) -> qiskit.res
             }
         )
 
+    def get_counts_on_qubits(
```
perlinm, line 104: My first thought:

The way to get counts in `qss` is through `job.result()`, so let's make this part of that method (right above).  So something like
```python
    def result(self, timeout: Optional[float] = None, wait: float = 5, qubits: Sequence[int]) -> qiskit.result.Result:
```

My second thought:

Can't a `SuperstaqJob` object hold multiple circuits?  What do you do if those circuits have different qubit numbers?  Maybe we actually need to pass a `Union[Sequence[int], Sequence[Sequence[int]]]` to cover the possibility of getting different qubits from different circuits...

cirq-superstaq/cirq_superstaq/job.py
```diff
@@ -201,7 +205,20 @@ def counts(self, timeout_seconds: int = 7200, polling_seconds: float = 1.0) -> D
             time_waited_seconds += polling_seconds
 
         self._check_if_unsuccessful()
-        return self._job["samples"]
+        counts_dictionary = self._job["samples"]
+
+        return counts_dictionary
+
+    def get_counts_on_qubits(
```
perlinm, line 212: This should be folded into the `def counts(...` above, not its own method.

general-superstaq/general_superstaq/superstaq_client.py
SalahedeenIssa: So should I change it such that it fits the definition set by 1?

cirq-superstaq/cirq_superstaq/job.py
```diff
@@ -201,7 +207,14 @@ def counts(self, timeout_seconds: int = 7200, polling_seconds: float = 1.0) -> D
...
         self._check_if_unsuccessful()
-        return self._job["samples"]
+        counts_dictionary = self._job["samples"]
+
+        if qubit_indices:
+            counts_dictionary = gss.superstaq_client.get_counts_on_qubits(
+                counts_dictionary, qubit_indices
+            )
+
+        return counts_dictionary
```
perlinm, line 217: imo typing info shouldn't be put into the name of an object.  It's kind of like saying `apple_fruit`.
```suggestion
        counts = self._job["samples"]
        if qubit_indices:
            counts = gss.superstaq_client.get_counts_on_qubits(
                counts, qubit_indices
            )
        return counts
```

cirq-superstaq/cirq_superstaq/job_test.py
```diff
@@ -203,6 +203,7 @@ def test_job_counts(job: css.job.Job) -> None:
     }
     with mocked_get_job_requests(job_dict):
         assert job.counts() == {"11": 1}
+        assert job.counts(qubit_indices=[0]) == ({"1": 1})
```
perlinm, line 206: nit: let's change the example above so that it's clear which qubit is "[0]".

That is, something like:
```python
def test_job_counts(job: css.job.Job) -> None:
    job_dict = {
        "data": {"histogram": {"10": 1}},
        "num_qubits": 2,
        "samples": {"10": 1},
        "shots": 1,
        "status": "Done",
        "target": "ss_unconstrained_simulator",
    }
    with mocked_get_job_requests(job_dict):
        assert job.counts() == {"10": 1}
        assert job.counts(qubit_indices=[0]) == ({"1": 1})
```

cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -246,6 +246,7 @@ def create_job(
         repetitions: int = 1000,
         target: Optional[str] = None,
         method: Optional[str] = None,
+        diagnostic_qubits: Optional[List[int]] = None,
```
perlinm, line 249:
```suggestion
        diagnostic_qubits: Optional[Sequence[int]] = None,
```

qiskit-superstaq/qiskit_superstaq/superstaq_job_test.py
```diff
@@ -68,15 +68,19 @@ def test_result(backend: qss.SuperstaqBackend) -> None:
...
     with mock.patch(
         "general_superstaq.superstaq_client._SuperstaqClient.get_job",
         return_value=mock_response("Done"),
     ):
-        ans = job.result()
+        ans = job.result(
+            qubit_indices=[
+                1,
+            ]
+        )
```
perlinm, line 79: nit
```suggestion
        ans = job.result(qubit_indices=[1])
```

qiskit-superstaq/qiskit_superstaq/superstaq_job.py
```diff
@@ -80,6 +86,9 @@ def result(self, timeout: Optional[float] = None, wait: float = 5) -> qiskit.res
             counts = result["samples"]
             if counts:  # change endianess to match Qiskit
                 counts = dict((key[::-1], value) for (key, value) in counts.items())
+
+            if qubit_indices:
+                counts = gss.superstaq_client.get_counts_on_qubits(counts, qubit_indices)
```
perlinm, line 91: nit (predates this PR): fix the misspelling of "endianness" above

perlinm (commented):
Looks good!  I'll wait to approve until the server-side PR is ready and we test offline that everything works :slightly_smiling_face:

perlinm (commented):
Looks good to me!  But I don't understand the clients much and don't trust that I won't make some error in how they're supposed to be used.

@vtomole @richrines1 @bharat-thotakura can I get a quick review on this?

TLDR it's just adding the capability to return counts/results from a job on specific qubits.

cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -250,6 +250,7 @@ def create_job(
         repetitions: int = 1000,
         target: Optional[str] = None,
         method: Optional[str] = None,
+        diagnostic_qubits: Optional[Sequence[int]] = None,
```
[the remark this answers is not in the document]
perlinm: Also, this is only being added to `css` services.  We need it to work with both `css` and `qss`.

bharat-thotakura:
@SalahedeenIssa @perlinm 

Perhaps I'm misunderstanding the purpose, but as far as `qss` is concerned, couldn't you use qiskit's [marginal counts](https://qiskit.org/documentation/locale/de_DE/stubs/qiskit.result.marginal_counts.html) to achieve the intent behind `get_counts_on_qubits()`?

perlinm:
@bharat-thotakura @SalahedeenIssa

Yep, looks like we should be using that method in `qss`.  In that case we only need to add a custom marginalize counts method to `css` rather than `gss`.

perlinm (approved):
LGTM

cirq-superstaq/cirq_superstaq/job.py
```diff
@@ -225,3 +235,20 @@ def __repr__(self) -> str:
...
+        indices: The indices of the bits on which to accumulate counts.
+
+    Returns:
+        A dictionary of counts on the target indices.
+    """
+    target_counts: Dict[str, int] = collections.defaultdict(int)
+    for bitstring, count in counts.items():
+        target_key = "".join([bitstring[index] for index in indices])
+        target_counts[target_key] += count
+    return target_counts
```
richrines1, line 254:
```suggestion
    return dict(target_counts)
```
(so that the returned value is not a defaultdict)

qiskit-superstaq/qiskit_superstaq/superstaq_job.py
```diff
@@ -59,14 +59,20 @@ def _wait_for_results(self, timeout: float, wait: float = 5) -> List[Dict[str, D
 
         return [self._job_info[job_id] for job_id in self._job_id.split(",")]
 
-    def result(self, timeout: Optional[float] = None, wait: float = 5) -> qiskit.result.Result:
+    def result(
+        self,
+        timeout: Optional[float] = None,
+        qubit_indices: Optional[Sequence[int]] = None,
+        wait: float = 5,
```
richrines1, line 66:
```suggestion
        wait: float = 5,
        qubit_indices: Optional[Sequence[int]] = None,
```

cirq-superstaq/cirq_superstaq/job.py
```diff
@@ -175,12 +176,18 @@ def input_circuit(self) -> cirq.Circuit:
...
+        timeout_seconds: int = 7200,
+        polling_seconds: float = 1.0,
+        qubit_indices: Optional[Sequence[int]] = None,
+    ) -> Dict[str, int]:
         """Polls the Superstaq API for counts results (frequency of each measurement outcome).
 
         Args:
             timeout_seconds: The total number of seconds to poll for.
             polling_seconds: The interval with which to poll.
+            qubit_indices: The qubit indices to return the results of individually.
```
richrines1, line 190: this is a little unclear imo. maybe something like

SalahedeenIssa: @richrines1 Something like what ?

richrines1: haha sorry got caught up in a conversation. i was thinking something like
```suggestion
            qubit_indices: If provided, only include measurements of these qubits in the counts dictionary.
```

cirq-superstaq/cirq_superstaq/job.py
```diff
@@ -225,3 +235,20 @@ def __repr__(self) -> str:
...
+        indices: The indices of the bits on which to accumulate counts.
+
+    Returns:
+        A dictionary of counts on the target indices.
+    """
+    target_counts: Dict[str, int] = collections.defaultdict(int)
+    for bitstring, count in counts.items():
+        target_key = "".join([bitstring[index] for index in indices])
+        target_counts[target_key] += count
+    return dict(target_counts)
```
perlinm, line 254: Leave as a counter (which is a subclass of `dict`)
```suggestion
    return target_counts
```
or, if you insist on an ordinary `dict`, don't make a counter to begin with, and accumulate counts with something like
```python
try:
    target_counts[target_key] += count
except KeyError:
    target_counts[target_key] = count
```

richrines1: i suggested this cast in order to make the return type of `job.counts()` consistent.

imo wrapping the output in `dict()` is way cleaner than adding a try..except block, and the cost of the extra copy should be negligible. but if you really want to avoid it we could also use something like
```python
        target_counts[target_key] = target_counts.get(target_key, 0) + count
```
or
```python
        if target_key not in target_counts:
            target_counts[target_key] = 0
        target_counts[target_key] += count
```
which imo are somewhat preferable to try..except

perlinm: Gotcha.  In that case @SalahedeenIssa you can go ahead and merge

PR #723, opened 2023-09-05
Update `build-system.requirements` to enforce newer versions of `setuptools` and `pip`

### LINKED ISSUE #714: Non editable project
### What is happening?

The changes I was making locally to cirq_superstaq were not being reflected when I was testing them locally. This turns out to be because I installed it using an outdated version of pip (<23). When I installed it again using a version of pip that was >23, it seemed to work fine. 

Install css or qss using pip<23

### What should happen?

You should be able to see local changes on local branches reflected when running locally. 

### Environment

Pip < 23

### Any additional context?

_No response_

perlinm:
fixes #714

...in theory.  Does not work at the moment, for reasons unknown.  `build-system.requires` seems to be ignored.

### DIFF
--- a/cirq-superstaq/pyproject.toml
+++ b/cirq-superstaq/pyproject.toml
@@ -13,7 +13,7 @@ dynamic = ["version", "dependencies", "optional-dependencies"]
 homepage = "https://github.com/Infleqtion/client-superstaq"
 
 [build-system]
-requires = ["setuptools"]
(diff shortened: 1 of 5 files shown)


stephanielee9:
@perlinm: @SalahedeenIssa is going to update the dev guide to update to the latest version of `pip` so this PR should not be needed

PR #740, opened 2023-09-09
Remove token setting methods from the client interface

cjguttormsson:
These don't actually do anything, and can be safely removed.


vtomole (approved):
LGTM

PR #741, opened 2023-09-11
Adding support for email validation

pgoiporia:


pgoiporia:
@cjguttormsson pinging for review

general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -615,6 +615,13 @@ def _handle_status_codes(self, response: requests.Response) -> None:
             ):
                 self._prompt_accept_terms_of_use()
                 return
+
+            elif response.json() == ("You must validate your registered email."):
```
cjguttormsson, line 619: Is this meant to be a tuple it's checking against? If not, I think the parentheses can be removed here.

cjguttormsson (approved):
One nit, not a big deal whether it changes or not.

PR #745, opened 2023-09-11
change package install order in ci.yml

richrines1:
qss always has to come before css[dev] to prevent it from trying to install qss from pypi


vtomole (approved):
LGTM

PR #748, opened 2023-09-12
Add support for DD kwarg in qss

paaige:


knsmith (approved):

qiskit-superstaq/qiskit_superstaq/superstaq_backend.py
```diff
@@ -144,12 +144,14 @@ def retrieve_job(self, job_id: str) -> qss.SuperstaqJob:
     def compile(
         self,
         circuits: Union[qiskit.QuantumCircuit, Sequence[qiskit.QuantumCircuit]],
+        dynamical_decoupling: bool = False,
```
singular-value, line 147: do we want it to default to True?

qiskit-superstaq/qiskit_superstaq/superstaq_backend.py
```diff
@@ -179,6 +181,7 @@ def compile(
     def _get_compile_request_json(
         self,
         circuits: Union[qiskit.QuantumCircuit, Sequence[qiskit.QuantumCircuit]],
+        dynamical_decoupling: bool = False,
```
richrines1, line 184: this shouldn't need to be added here, it'd be better to put this in the options dict

knsmith (approved):

richrines1 (approved):
lgtm! thanks for the quick fix :) we should probably think of a good way to test this but for now let's just get it in

also definitely not for this pr but we should probably add dynamical_decoupling as an argument to qss.superstaq_provider.ibmq_compile and css.service.ibmq_compile as well

qiskit-superstaq/qiskit_superstaq/superstaq_backend.py
```diff
@@ -150,6 +150,7 @@ def compile(
 
         Args:
             circuits: The qiskit QuantumCircuit(s) to compile.
+            dynamical_decoupling: Setting to add dynamical decoupling to the circuit.
```
richrines1, line 153:
```suggestion
```

PR #759, opened 2023-09-12
add missing requirements

perlinm:
versions copied from `server-superstaq`

also increased time limit for `build-docs`


vtomole (commented):
Match the versions in https://github.com/Infleqtion/client-superstaq/blob/main/docs/requirements.txt

perlinm:
@vtomole done.  I opted to relax requirements to `>=older.version` as in https://github.com/Infleqtion/client-superstaq/blob/main/docs/requirements.txt

vtomole (approved):
LGTM

PR #770, opened 2023-09-14
Update DD notebook to include strategies

paaige:


paaige:
I'm still seeing an issue on all of the static methods (they are applying DD after measurement).

BV_2_static

knsmith:
> I'm still seeing an issue on all of the static methods (they are applying DD after measurement). This is the static method on a 2 qubit BV circuit, seeing similar on SCA for larger circuits.
> 
> BV_2_static

Summary of dev meeting discussion: DD is scheduled on all qubits, and it's appearing on q1 while q0 is being measured. Unfortunately, drawing the circuit using qiskit doesn't give context wrt the duration of the measurement. Timeline drawer would be a better tool to get a sense of timing.

Perhaps a solution that would create a nicer circuit would be to use a different circuit (GHZ was also a nice one - I think we had this in an earlier version of this notebook) or to place a measurement operation on all qubits?

knsmith:
IMO we should have 2 notebooks: one DD user test and a second notebook that goes into greater detail about strategies. 

It would also be good to emphasize that Superstaq compilation inserts DD by default. We state that dynamic is the default technique but it would be good for users to know that DD will always be invoked unless `dynamical_decoupling = False`

docs/source/optimizations/ibm/ibmq_dd_advanced.ipynb
```diff
@@ -0,0 +1,376 @@
+{
```
knsmith, line 1: <p>Maybe we could make this more specific? Such as "IBM Dynamical Decoupling Strategies with Superstaq"?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Infleqtion/client-superstaq/pull/770/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

knsmith, line 1: <p>We could probably shorten this a bit. I'd suggest cutting out everything before "There are many possible ways to determine..." and adding a sentence at the beginning to introduce the notebook's topics. The sentence could be something simple like:</p><p> </p><p>"In this notebook, we will examine different ways Superstaq places dynamical decoupling (DD) sequences within circuit idle windows.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Infleqtion/client-superstaq/pull/770/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

knsmith, line 1: <p>We could remove this cell since we aren't submitting circuits to IBM QCs</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Infleqtion/client-superstaq/pull/770/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

paaige: Need to update this but I think this PR is good to go after @knsmith

paaige:
Waiting on #824 to merge before this goes in

knsmith:
This looks pretty much good to go! A few more minor comments:

- Small nit: superstaq -> Superstaq in the intro paragraph.

- Optional change: "All of the above strategies, by default, use the CPMG sequence. " -> "All of the above strategies, by default, use the CPMG, or XX, sequence."

- Don't forget to remove `remote_host="https://127.0.0.1:5000"`

- This will be fixed when the notebook is re-run, but right now, it looks like the last two circuits produce ghz states of different sizes

knsmith (changes requested):
just a few minor nits.

knsmith:
The second to last cell might need to be rerun since the default `dd_strategy` is now `static_context_aware` - I guess the intro paragraph should be updated as well...

Other than that, it looks good to go!

knsmith (approved):
LG2M !(once we determine why notebook checks are failing...)

paaige:
> LG2M !(once we determine why notebook checks are failing...)

I had to switch from using guadalupe to perth

PR #771, opened 2023-09-14
Remove `supermarq` from readthedocs

vtomole:
Supermarq is not a product so it's a bit misleading to have it's docs sit alongside Superstaq.


bharat-thotakura (approved):
LGTM

PR #773, opened 2023-09-15
Add pulse_gate_circuit support in css

### LINKED ISSUE #768: Have client-superstaq Jobs return pulse_gate_circuits
### What is happening?

Right now, we only return the compiled circuit which is a cirq.Circuit. This does not retain timing information (so if someone wanted to analyze dynamical decoupling applied to a circuit, they would not be able to). 

### Any additional context?

The solution is for users to be able to call job.pulse_gate_circuit after creating a job for a pulse device backend.

paaige:
Closes #768


cirq-superstaq/cirq_superstaq/job.py
```diff
@@ -154,7 +156,7 @@ def repetitions(self) -> int:
         return self._job["shots"]
 
     def compiled_circuit(self) -> cirq.Circuit:
-        """Gets the compiled circuit that was submitted for this job.
+        """Gets the compiled circuit returned by this job.
```
paaige, line 159: `'for`

paaige:
will add a test (currently testing on a real device)

cirq-superstaq/requirements.txt
```diff
@@ -1,3 +1,4 @@
 cirq>=1.0.0
 general-superstaq~=0.4.26
+qiskit>=0.43.3
```
richrines1, line 3: we don't want css to require qiskit by default. it's already required in `example-requirements` so we should have it available for tests anyway - see similar tests in compiler_output_test.py

cirq-superstaq/cirq_superstaq/job_test.py
```diff
@@ -134,6 +136,22 @@ def test_compiled_circuit(job: css.job.Job) -> None:
     assert job.compiled_circuits(index=0) == compiled_circuit
 
 
+def test_pulse_gate_circuit(job: css.job.Job) -> None:
+    pulse_gate_circuit = qiskit.QuantumCircuit(1, 1)
+    pulse_gate = qiskit.circuit.Gate("test_pulse_gate", 1, [3.14, 1])
+    pulse_gate_circuit.append(pulse_gate, [0])
+    pulse_gate_circuit.measure(0, 0)
+    job_dict = {"status": "Done", "pulse_gate_circuits": qss.serialize_circuits(pulse_gate_circuit)}
```
richrines1, line 144: we already have a function in compiler_output.py (`_deserialize_qiskit_circuits`) to deserialize qiskit circuits, with special handling for qiskit not being installed/etc. - let's move to `css.serialization.deserialize_qiskit_circuits` so we can use it here too

cirq-superstaq/cirq_superstaq/job.py
```diff
@@ -319,6 +320,37 @@ def input_circuits(
...
+            deserialized_circuits = []
+            for serialized_circuit in serialized_circuits:
+                deserialized_circuit = css.serialization.deserialize_qiskit_circuits(
+                    serialized_circuit, False
+                )
+                if deserialized_circuit is not None:
+                    deserialized_circuits.append(deserialized_circuit[0])
+            return deserialized_circuits
+
+        error = f"Job: {self._job_id} was not submitted to a pulse-enabled device"
```
richrines1, line 351: wdyt about changing this to something like
```suggestion
        error = f"Target '{self.target()}' does not use pulse gate circuits."
```
? some targets might be pulse-enabled (e.g. via braket) but not use qiskit pulse gate circuits

cirq-superstaq/cirq_superstaq/job.py
```diff
@@ -319,6 +320,44 @@ def input_circuits(
...
+                serialized_circuits = [
+                    self._job[job_id]["pulse_gate_circuits"] for job_id in job_ids
+                ]
+                deserialized_circuits = []
+                for serialized_circuit in serialized_circuits:
+                    deserialized_circuit = css.serialization.deserialize_qiskit_circuits(
+                        serialized_circuit, False
+                    )
+                    if deserialized_circuit is not None:
+                        deserialized_circuits.append(deserialized_circuit[0])
```
richrines1, line 351: if one circuit fails to deserialize, they probably all will so imo it's better to bail out instead of printing the same warning over and over (and otherwise they can always uses `index` to try to deserialize one circuit at a time, which is probably more useful than getting an unspecified subset of the circuits)
```suggestion
                    if deserialized_circuit is None:
                        raise ValueError("Some circuits could not be deserialized.")
                    deserialized_circuits.append(deserialized_circuit[0])
```

richrines1 (commented):
one suggestion but otherwise this looks great :)

richrines1:
(lg % coverage check)

cirq-superstaq/cirq_superstaq/job.py
```diff
@@ -235,13 +235,14 @@ def _get_circuits(self, circuit_type: str, index: int) -> cirq.Circuit:
...
-        cirq.Circuit, List[cirq.Circuit]
+        cirq.Circuit,
+        List[cirq.Circuit],
     ]:  # Change return to `List[cirq.Circuit]` after deprecation
         ...
 
     def _get_circuits(
         self, circuit_type: str, index: Optional[int] = None
-    ) -> Union[cirq.Circuit, List[cirq.Circuit]]:
+    ) -> Union[cirq.Circuit, List[cirq.Circuit],]:
```
richrines1, line 245:
```suggestion
    ) -> Union[cirq.Circuit, List[cirq.Circuit]]:
```

richrines1 (approved):
lgtm (% typo)!

PR #780, opened 2023-09-19
revert addition of "device" attribute to serialized circuits

richrines1:
reverts https://github.com/Infleqtion/cirq-superstaq/pull/241 (which at the time was necessary because we supported `cirq<=0.14.*`)


vtomole (approved):
LGTM

PR #781, opened 2023-09-19
Stricter assert for Hilbert integration test

vtomole:
Remove simple test and refine the `assert of https://github.com/Infleqtion/client-superstaq/pull/777

### DIFF
--- a/cirq-superstaq/cirq_superstaq/daily_integration_test.py
+++ b/cirq-superstaq/cirq_superstaq/daily_integration_test.py
@@ -3,6 +3,7 @@
 
 import cirq
 import general_superstaq as gss
+import numpy as np
 import pytest
 from general_superstaq import ResourceEstimate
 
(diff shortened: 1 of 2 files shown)


cirq-superstaq/cirq_superstaq/daily_integration_test.py
```diff
@@ -286,13 +287,24 @@ def test_submit_to_provider_simulators(target: str, service: css.Service) -> Non
...
+        css.ParallelRGate(np.pi / 2, 0.0, 24).on(*qubits),
+        cirq.rz(np.pi).on(qubits[2]),
+        css.ParallelRGate(-np.pi / 2, 0.0, 24).on(*qubits),
+        cirq.measure(*qubits),
+    )
 
-    job = service.create_job(circuit=circuit, repetitions=100, target="cq_hilbert_qpu")
-    assert sum(job.counts().values()) == 100
+    job = service.create_job(
+        circuit=circuit, repetitions=400, verbatim=True, route=False, target=target
```
richrines1, line 303: do we need this many shots? it seems like 100 should still be enough for this to pass consistently (though also not sure if there's any reason to use fewer)

richrines1 (approved):
one question but lgtm!

cirq-superstaq/cirq_superstaq/daily_integration_test.py
vtomole: Changed to `100`

vtomole:
[Passes after 10 minutes](https://github.com/Infleqtion/client-superstaq/actions/runs/6242256478) when not reserved.

PR #783, opened 2023-09-20
Handle float counts + warn users

### LINKED ISSUE #782: `css.Service.run` is broken for ibmq targets
### What is happening?

counts returned from ibmq targets are usually floats because we enable measurement error mitigation by default. This breaks `service.run()` because we try to use the counts to construct a `cirq.ResultDict`, which only makes sense if they're integers

(originally uncovered by @cjguttormsson)

```python
service = css.Service()
circuit = cirq.Circuit(
    cirq.H(cirq.q(0)),
    cirq.measure(cirq.q(0)),
)
_ = service.run(circuit, target='ibmq_qasm_simulator', repetitions=10)
```
which raises
```
TypeError: 'float' object cannot be interpreted as an integer
```

### What should happen?

`run()` should return a `cirq.ResultDict`

### Environment

- **Superstaq version**: 0.4.38
- **client-superstaq version**: 0.4.22

### Any additional context?

_No response_

dowusu-antwi:
Fixes #782.

This primarily handles getting counts like `{"01": 500.0, "10": 500.0}`, where they're already whole numbers. For `float` results more generally, I claim it's enough to warn users since we also offer `service.get_counts` as a public method, but I'm open to ideas about how else to address this.

(Seems like we can't get around `cirq.ResultDict` needing a list/array of results for the `measurements` param, and we can't append half or a quarter of an array element...)

### DIFF
--- a/cirq-superstaq/cirq_superstaq/service.py
+++ b/cirq-superstaq/cirq_superstaq/service.py
@@ -58,7 +58,9 @@ def _to_matrix_gate(matrix: npt.ArrayLike) -> cirq.MatrixGate:
 
 
 def counts_to_results(
-    counter: Dict[str, int], circuit: cirq.AbstractCircuit, param_resolver: cirq.ParamResolver
+    counter: Union[Dict[str, int], Dict[str, float]],
+    circuit: cirq.AbstractCircuit,
+    param_resolver: cirq.ParamResolver,
 ) -> cirq.ResultDict:
     """Converts a `collections.Counter` to a `cirq.ResultDict`.
 
@@ -91,6 +93,17 @@ def counts_to_results(
         # Appends all the keys onto 'samples' list number-of-counts-in-the-key times
         # If collections.Counter({"01": 48, "11": 52}), [0, 1] is appended to 'samples` 48 times and
         # [1, 1] is appended to 'samples' 52 times
+        if int(counts_of_key) != counts_of_key:
+            print(
+                warnings.warn(
+                    "The raw counts are fractional due to measurement error mitigation; please use "
+                    "service.get_counts to see raw results.",
+                    UserWarning,
+                    stacklevel=2,
+                )
+            )
+
+        counts_of_key = int(counts_of_key)
         for _ in range(counts_of_key):
             samples.append(keys_as_list)
 
--- a/cirq-superstaq/cirq_superstaq/service.py
+++ b/cirq-superstaq/cirq_superstaq/service.py
@@ -58,7 +58,7 @@ def _to_matrix_gate(matrix: npt.ArrayLike) -> cirq.MatrixGate:
 
 
 def counts_to_results(
-    counter: Union[Dict[str, int], Dict[str, float]],
+    counter: Mapping[str, float],
     circuit: cirq.AbstractCircuit,
     param_resolver: cirq.ParamResolver,
 ) -> cirq.ResultDict:
--- a/cirq-superstaq/cirq_superstaq/service.py
+++ b/cirq-superstaq/cirq_superstaq/service.py
@@ -91,13 +91,10 @@ def counts_to_results(
         # If collections.Counter({"01": 48, "11": 52}), [0, 1] is appended to 'samples` 48 times and
         # [1, 1] is appended to 'samples' 52 times
         if int(counts_of_key) != counts_of_key:
-            print(
-                warnings.warn(
-                    "The raw counts are fractional due to measurement error mitigation; please use "
-                    "service.get_counts to see raw results.",
-                    UserWarning,
-                    stacklevel=2,
-                )
+            warnings.warn(
+                "The raw counts are fractional due to measurement error mitigation; please use "
+                "service.get_counts to see raw results.",
+                stacklevel=2,
             )
 
         counts_of_key = int(counts_of_key)
--- a/cirq-superstaq/cirq_superstaq/service.py
+++ b/cirq-superstaq/cirq_superstaq/service.py
@@ -81,8 +81,14 @@ def counts_to_results(
     samples: List[List[int]] = []
     if not all(counts == int(counts) for counts in counter.values()):
         warnings.warn(
-            "The raw counts are fractional due to measurement error mitigation; please use "
-            "service.get_counts to see raw results.",
+            "The raw counts contain fractional values due to measurement error mitigation; please "
+            "use service.get_counts to see raw results.",
+            stacklevel=2,
+        )
+    if not all(counts >= 0 for counts in counter.values()):
+        warnings.warn(
+            "The raw counts contain negative values due to measurement error mitigation; please "
+            "use service.get_counts to see raw results.",
             stacklevel=2,
         )
     for key, counts_of_key in counter.items():
--- a/cirq-superstaq/cirq_superstaq/service.py
+++ b/cirq-superstaq/cirq_superstaq/service.py
@@ -58,7 +58,9 @@ def _to_matrix_gate(matrix: npt.ArrayLike) -> cirq.MatrixGate:
 
 
 def counts_to_results(
-    counter: Dict[str, float], circuit: cirq.AbstractCircuit, param_resolver: cirq.ParamResolver
+    counter: Union[Dict[str, int], Dict[str, float]],
+    circuit: cirq.AbstractCircuit,
+    param_resolver: cirq.ParamResolver,
 ) -> cirq.ResultDict:
     """Converts a `collections.Counter` to a `cirq.ResultDict`.
 
--- a/cirq-superstaq/cirq_superstaq/service.py
+++ b/cirq-superstaq/cirq_superstaq/service.py
@@ -79,25 +79,21 @@ def counts_to_results(
     combine_key_names = "".join(measurement_key_names)
 
     samples: List[List[int]] = []
-    for key in counter.keys():
+    if not all(counts == int(counts) for counts in counter.values()):
+        warnings.warn(
+            "The raw counts are fractional due to measurement error mitigation; please use "
+            "service.get_counts to see raw results.",
+            stacklevel=2,
+        )
+    for key, counts_of_key in counter.items():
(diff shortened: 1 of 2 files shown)


cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -91,6 +90,17 @@ def counts_to_results(
...
         # [1, 1] is appended to 'samples' 52 times
+        if int(counts_of_key) != counts_of_key:
+            print(
+                warnings.warn(
+                    "The raw counts are fractional due to measurement error mitigation; please use "
+                    "service.get_counts to see raw results.",
+                    UserWarning,
+                    stacklevel=2,
+                )
+            )
```
richrines1, line 101: remove print, and no need to specify `UserWarning`
```suggestion
            warnings.warn(
                "The raw counts are fractional due to measurement error mitigation; please use "
                "service.get_counts to see raw results.",
                stacklevel=2,
            )
```

cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -78,11 +80,8 @@ def counts_to_results(
 
     samples: List[List[int]] = []
     for key in counter.keys():
```
richrines1, line 82:
```suggestion
    for key, counts_of_key in counter.items():
```

cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -91,6 +90,17 @@ def counts_to_results(
         # Appends all the keys onto 'samples' list number-of-counts-in-the-key times
         # If collections.Counter({"01": 48, "11": 52}), [0, 1] is appended to 'samples` 48 times and
         # [1, 1] is appended to 'samples' 52 times
+        if int(counts_of_key) != counts_of_key:
+            print(
+                warnings.warn(
```
richrines1, line 95: we probably don't want to warn for every key - maybe we could just precede the loop with a single check like
```python
if not all(counts == int(counts) for counts in counter.values()):
    warnings.warn(...)
```

cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -78,11 +80,8 @@ def counts_to_results(
 
     samples: List[List[int]] = []
     for key in counter.keys():
-        keys_as_list: List[int] = []
-
         # Combines the keys of the counter into a list. If key = "01", keys_as_list = [0, 1]
-        for index in key:
-            keys_as_list.append(int(index))
+        keys_as_list: List[int] = list(map(int, key))
```
richrines1, line 84: might make sense to use `int(round(key))` to make this a little bit less wrong for ibm devices

cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -58,7 +58,9 @@ def _to_matrix_gate(matrix: npt.ArrayLike) -> cirq.MatrixGate:
 
 
 def counts_to_results(
-    counter: Dict[str, int], circuit: cirq.AbstractCircuit, param_resolver: cirq.ParamResolver
+    counter: Union[Dict[str, int], Dict[str, float]],
+    circuit: cirq.AbstractCircuit,
+    param_resolver: cirq.ParamResolver,
```
richrines1, line 63: ints should count as a floats
```suggestion
    counter: Dict[str, float], circuit: cirq.AbstractCircuit, param_resolver: cirq.ParamResolver
```

cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -91,6 +90,17 @@ def counts_to_results(
...
+            print(
+                warnings.warn(
+                    "The raw counts are fractional due to measurement error mitigation; please use "
+                    "service.get_counts to see raw results.",
+                    UserWarning,
+                    stacklevel=2,
+                )
+            )
+
+        counts_of_key = int(counts_of_key)
```
cjguttormsson, line 103: I think that this maybe should be `round` instead of int, if we're expecting to something get something like `{"0": 7.1, "1": 2.9}`. In either case, there's the potential of getting values that don't add up to the original total, but I think `round` would generally get us closer.

cirq-superstaq/cirq_superstaq/service.py
cjguttormsson: I'm not sure what you're suggesting here, `round` doesn't work on strings. It also already returns an int.

richrines1: oops, referenced the wrong thing in my suggestion - i just meant [this](https://github.com/Infleqtion/client-superstaq/pull/783#discussion_r1331974544) :)

cirq-superstaq/cirq_superstaq/service.py
dowusu-antwi: Done.

cirq-superstaq/cirq_superstaq/service.py
dowusu-antwi: Done.

cirq-superstaq/cirq_superstaq/service.py
dowusu-antwi: Good point -- done.

cirq-superstaq/cirq_superstaq/service.py
dowusu-antwi: > oops, referenced the wrong thing in my suggestion - i just meant https://github.com/Infleqtion/client-superstaq/pull/783#discussion_r1331974544 :)

Done. (Yup, `round` returns an `int`.)

cirq-superstaq/cirq_superstaq/service.py
dowusu-antwi: > ints should count as a floats

On second thought (GitHub checks) `mypy` complains?

cjguttormsson: It looks like that's coming from the interaction of this code with some other code here: https://github.com/Infleqtion/client-superstaq/blob/458a26a8670b1f2571c04114d7f745c80e7c1c3e/cirq-superstaq/cirq_superstaq/job.py#L179

Although mypy appears to be able to figure out that int literals can be treated like floats, it doesn't extend that logic to types contained within composite types. So in other words, treating an `int` as a `float` works, but not treating a `Dict[str, int]` like a `Dict[str, float]`.

Theoretically, I suppose that the solution here would be to propagate out the change to any code that uses this code, but that could potentially cause more problems than it solves. I think that going with the union of the two dict types makes sense here.

cjguttormsson: Interesting, so I just explored this even more and figured out that the suggestion mentioned in the warning -- just changing `Dict` to `Mapping` actually fixes this completely as well. So you could just use `Mapping[str, float]` and be fine here.

dowusu-antwi: Oh, neat! Wonder why `Mapping` is treated differently than `Dict` -- seems like it shouldn't depend on the enclosing object type.

Done.

cjguttormsson (approved):

cirq-superstaq/cirq_superstaq/service.py
richrines1: `Mapping` is covariant and dict is invariant in value type (for this reason we should in general prefer `Mapping` over `Dict` for input annotations, and `dict` for output annotations)

richrines1:
one more weird case we might need to consider (thanks to convo w/ @paaige) - i think qiskit's quasi-probabilities can sometimes be negative?

paaige:
> one more weird case we might need to consider (thanks to convo w/ @paaige) - i think qiskit's quasi-probabilities can sometimes be negative?

Yeah, I've been doing experiments using the runtime sampler (which returns quasi probability distributions) and have had to filter out negative results.

dowusu-antwi:
> > one more weird case we might need to consider (thanks to convo w/ @paaige) - i think qiskit's quasi-probabilities can sometimes be negative?
> 
> Yeah, I've been doing experiments using the runtime sampler (which returns quasi probability distributions) and have had to filter out negative results.

Right good points -- I added another `warnings` case for negative counts.

cirq-superstaq/cirq_superstaq/service_test.py
```diff
@@ -63,6 +63,21 @@ def test_counts_to_results() -> None:
...
+    with pytest.warns(UserWarning, match="raw counts contain fractional"):
+        result = css.service.counts_to_results(
+            {"00": 50.1, "11": 49.9}, circuit, cirq.ParamResolver({})
+        )
+
+    with pytest.warns(UserWarning, match="raw counts contain negative"):
+        result = css.service.counts_to_results(
+            {"00": -50.1, "11": 99.9}, circuit, cirq.ParamResolver({})
+        )
+
```
cjguttormsson, line 80: Could you add asserts for the resulting counts to these two? Just to make it a bit more explicit about what the code does with these unexpected numbers.

dowusu-antwi: Thanks for this -- it uncovered another error. We can't handle negative numbers in the `cirq.ResultDict` for the same reason we can't handle fractional numbers, because you can't add negatively-many elements to an array:
```python
# In counts_to_results:
for key, counts_of_key in counter.items():
        keys_as_list: List[int] = list(map(int, key))

        counts_of_key = round(counts_of_key)
        for _ in range(counts_of_key):
            samples.append(keys_as_list)

# counter = {"00": 2, "11": 3} -> samples = [[0, 0], [0, 0], [1, 1], [1, 1], [1, 1]]

# counter = {"00": -2, "11": 3} -> samples = [[1, 1], [1, 1], [1, 1]]
```

`cirq.ResultDict` needs `samples` for its `measurements` parameter, so I don't see how we can get around this without just ignoring negative counts... (which is maybe fine since we have the warning)

dowusu-antwi: ^ https://github.com/Infleqtion/client-superstaq/issues/785

cjguttormsson (approved):

PR #789, opened 2023-09-23
Factor out `checks-superstaq`

perlinm:
image


perlinm:
There's definitely a lot going on in `.github/workflows/ci.yaml` that I don't understand.  Help appreciated :sweat_smile:

.github/workflows/ci.yml
```diff
@@ -50,7 +50,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install -e ./general-superstaq[dev] -e ./qiskit-superstaq -e ./cirq-superstaq -e ./supermarq-benchmarks
+          pip install -e ./checks-superstaq -e ./general-superstaq -e ./qiskit-superstaq -e ./cirq-superstaq -e ./supermarq-benchmarks
```
perlinm, line 53: Is there a particular reason for the `pylint` to not install `[dev]` versions?

perlinm:
The requirements check is failing because `checks-superstaq` is not on PyPI.  So it's a chicken-and-egg problem.  Maybe we comment out the requirements check in this PR, version bump in a second PR (which should deploy to PyPI), and then un-comment this check in a third PR?

perlinm:
Not sure how to make sense of the failing requirements check in light of [this](https://github.com/Infleqtion/client-superstaq/blob/6683786a66ee2310e1ef6b4bc4419bd8e63e52df/.github/workflows/ci.yml#L235).
image

richrines1:
> Third option: if a package is installed locally and not found on PyPI, just check against the local version. I think I'll do this actually.

isn't this what happens if you just change `gss` to `checks_superstaq`? if so can we move the other changes to requirements.py to another pr?

perlinm:
@richrines1 part of the whole point here is to make the checks independent of other Superstaq packages, so that they are usable by independent projects.  So there should be no version check against against `gss.__version__` (or `checks_ss.__version__`, because the "upstream package" may be something different entirely.

richrines1:
> @richrines1 this PR was prompted by a desire to make the checks independent of other Superstaq packages, so that they would be usable by independent projects. So there should be no version check against against `gss.__version__` (or `checks_ss.__version__`), because the "upstream package" may be something different entirely.

can we still do this in two steps though? factoring out the checks/making a new package is a major change, it would be nice to be able to test/review that separately from changes to functionality, and to isolate the latter to a single pr in case we need to revise/debug/revert/etc (unless there's an obvious reason i'm missing that would make this not work)

.github/workflows/ci.yml
richrines1: nope, mostly there was just no reason to

.github/workflows/ci.yml
```diff
@@ -171,7 +171,7 @@ jobs:
           qiskit_version=$(sed qiskit-superstaq/requirements.txt -ne 's/^qiskit.=/qiskit==/p')
           echo "Using minimum qiskit version: $qiskit_version"
           python -m pip install --upgrade pip
-          pip install "$qiskit_version" -e ./general-superstaq[dev] -e ./qiskit-superstaq[dev]
+          pip install "$qiskit_version" -e ./checks-superstaq -e ./qiskit-superstaq[dev]
```
richrines1, line 174: gss is still necessary here
```suggestion
          pip install "$qiskit_version" -e ./checks-superstaq -e ./general-superstaq -e ./qiskit-superstaq[dev]
```

LICENSE
```diff
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright {yyyy} {name of copyright owner}
+   Copyright 2022 Infleqtion
```
richrines1, line 189:
```suggestion
   Copyright {yyyy} {name of copyright owner}
```
(this is a template that's part of license text; it's not supposed to be filled in)

.github/workflows/ci.yml
perlinm: I guess no need to add it then?  Should I remove `[dev]` from some other places while we're at it?  Maybe it doesn't matter?

.github/workflows/ci.yml
```diff
@@ -152,7 +152,7 @@ jobs:
           cirq_version=$(sed cirq-superstaq/requirements.txt -ne 's/^cirq.=/cirq==/p')
           echo "Using minimum cirq version: $cirq_version"
           python -m pip install --upgrade pip
-          pip install "$cirq_version" -e ./general-superstaq[dev] -e ./qiskit-superstaq -e ./cirq-superstaq[dev]
+          pip install "$cirq_version" -e ./checks-superstaq -e ./qiskit-superstaq -e ./cirq-superstaq[dev]
```
richrines1, line 155:
```suggestion
          pip install "$cirq_version" -e ./checks-superstaq -e ./general-superstaq -e ./qiskit-superstaq -e ./cirq-superstaq[dev]
```

perlinm:
> can we still do this in two steps though?

Makes sense.  Will do 🙂

LICENSE
perlinm: Oh I've been totally doing this wrong elsewhere 😂

.github/workflows/ci.yml
richrines1: it shouldn't really matter much i think - in practice the dev versions are often installed anyway because we usually use a shared cache. leaving it off can make that check a little faster after a cache miss, but the difference is probably pretty small

LICENSE
richrines1: haha to be fair even qiskit [gets it wrong](https://github.com/Qiskit/qiskit/blob/6a38279c291f49f65c2976039cf209dfe58e9cfe/LICENSE.txt#L191) (as did we up until somewhat recently)

(cirq [gets it right](https://github.com/quantumlib/Cirq/blob/8e4e7d147d6d4ea5dd8111ba21a715aa3acd955c/LICENSE#L189) though)

checks-superstaq/checks_superstaq/requirements.py
```diff
@@ -200,10 +200,14 @@ def _check_package_versions(
 def _get_latest_version(package: str) -> str:
     base_package = package.split("[")[0]  # remove options: package_name[options] --> package_name
     pypi_url = f"https://pypi.org/pypi/{base_package}/json"
-    pypi_version = json.loads(urllib.request.urlopen(pypi_url).read().decode())["info"]["version"]
+    try:
+        package_data = urllib.request.urlopen(pypi_url).read().decode()
+        pypi_version = json.loads(package_data)["info"]["version"]
+    except urllib.error.URLError:
+        return checks_superstaq.__version__
```
perlinm, line 207: This is really a temporary fix for the fact that the `requirements.py` check will fail until `checks-superstaq` is on PyPI.  Not unreasonable since we take the `max` of the PyPI and local package versions below anyways.

perlinm:
I presume we should version bump to `0.5.0` in this PR (or soon after)?

richrines1 (commented):
`general-superstaq[dev]` should be replaced with `checks-superstaq` in */dev-requirements.txt. Otherwise, just a few minor questions!

general-superstaq/pyproject.toml
```diff
@@ -5,7 +5,7 @@ authors = [
   { name = "Superstaq development team", email = "superstaq@infleqtion.com" },
 ]
 readme = { file = "README.md", content-type = "text/markdown" }
-license = { text = "Apache 2" }
+license = { text = "Apache-2.0" }
```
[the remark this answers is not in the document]
perlinm: I see `Apache-2.0` on [github](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/licensing-a-repository#searching-github-by-license-type) and [poetry](https://python-poetry.org/docs/pyproject#license).

This is definitely not a big deal either way though.

check/all_.py
```diff
@@ -1,8 +1,7 @@
 #!/usr/bin/env python3
-
 import sys
 
-import general_superstaq.check
+import checks_superstaq as check
```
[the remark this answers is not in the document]
perlinm: Nope.  Maybe "this repo checks superstaq" is a nice play on words?  I am happy to make this `check` though, and even to consider different names entirely -- the naming collision between `./checks-superstaq/` and `./check/all_.py` (for example) is a :-1: for tab completion.

checks-superstaq/pyproject.toml
```diff
@@ -0,0 +1,35 @@
+[project]
+name = "checks-superstaq"
+description = "Check script tools for Superstaq."
+authors = [
+  { name = "Superstaq development team", email = "superstaq@infleqtion.com" },
+]
+readme = { file = "README.md", content-type = "text/markdown" }
+license = { text = "Apache-2.0" }
+requires-python = ">=3.8.0"
+dynamic = ["version", "dependencies", "optional-dependencies"]
```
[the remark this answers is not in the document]
perlinm: Good catch

general-superstaq/pyproject.toml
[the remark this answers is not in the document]
perlinm: I guess for the sake of neater git blame type stuff I'll leave it as it was before

check/all_.py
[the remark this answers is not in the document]
richrines1: haha i like "checks-superstaq" - i was mostly wondering if it would be more consistent to import it with `as checks` instead of `as check` (doesn't help with the tab-completion point though)

general-superstaq/pyproject.toml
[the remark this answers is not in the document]
richrines1: ah good find, `Apache-2.0` seem right then

check/all_.py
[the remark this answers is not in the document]
perlinm: Ah, I see.  Yeah `import checks` makes sense to me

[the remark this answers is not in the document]
perlinm: Oh I see I read your original comment backwards lol (swapping `check` and `checks`)

richrines1 (approved):
nice - this looks solid afaict

i think we should manually deploy checks-superstaq 0.4.25 to pypi, so from now it will be updated normally along with version bumps

perlinm:
Why not just version bump in this PR?  Wouldn't that deploy?  I feel like it doesn't make sense to deploy `checks-superstaq` without simultaneously version bumping `gss`.

PR #800, opened 2023-10-01
Set timeout on Hilbert submission daily integration test to 5 minutes

vtomole:
Hotfix for now. We probably want to put this as a separate test longterm so a different issue will be made by the bot when the test fails.

### DIFF
--- a/.github/workflows/daily-integration-check.yml
+++ b/.github/workflows/daily-integration-check.yml
@@ -12,6 +12,7 @@ jobs:
     env:
       SUPERSTAQ_API_KEY : ${{ secrets.SUPERSTAQ_API_KEY }}
     runs-on: ubuntu-latest
+    timeout-minutes: 15
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/daily-integration-check.yml
```diff
@@ -12,6 +12,7 @@ jobs:
     env:
       SUPERSTAQ_API_KEY : ${{ secrets.SUPERSTAQ_API_KEY }}
     runs-on: ubuntu-latest
+    timeout-minutes: 15
```
perlinm, line 15: 5 min instead of 15?

perlinm (approved):
Question about 5 vs 15 min

.github/workflows/daily-integration-check.yml
vtomole: 5 minutes for the individual Hilbert test. 15 for the whole suite.

PR #802, opened 2023-10-02
Cleanup in `requirements.py`

perlinm:
Main changes:
- Remove explicit check against `checks_superstaq.__version__`, making `checks-superstaq` independent of the other `*-superstaq` repos (and thereby more usable by independent projects such as `QFI-Opt` and `qLDPC`).
- When retrieving the the latest version of a package, take the `max` of the local version and that on PyPI.  If either the local or PyPI version of a package is not found, then that version is set to `0.0.0` before taking the `max`.
  - The local package version is what replaces the call to `checks_superstaq.__version__` mentioned above.
- Remove an unnecessary check for whether `requirements.py` can connect to PyPI.  Either a package is retrieved from PyPI or it isn't.

### DIFF
--- a/checks-superstaq/checks_superstaq/requirements.py
+++ b/checks-superstaq/checks_superstaq/requirements.py
@@ -194,11 +194,12 @@ def _get_latest_version(package: str, silent: bool) -> str:
 def _get_local_version(package: str) -> Optional[str]:
     """Retrieve the local version of a package (if installed)."""
     base_package = package.split("[")[0]  # remove options: package_name[options] --> package_name
+    package_name = base_package.replace("-", "_").lower()
     try:
-        module = importlib.import_module(base_package.replace("-", "_"))
+        module = importlib.import_module(package_name)
         if hasattr(module, "__version__"):
             return module.__version__
-        return importlib.metadata.version(base_package)
+        return importlib.metadata.version(package_name)
     except ModuleNotFoundError:
         return None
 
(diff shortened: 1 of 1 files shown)


checks-superstaq/checks_superstaq/requirements.py
```diff
@@ -197,38 +182,49 @@ def _check_package_versions(
...
+    if not local_version and not pypi_version:
+        raise ModuleNotFoundError(f"Package not installed or found on PyPI: {package}")
+    return max(pypi_version or "0.0.0", local_version or "0.0.0", key=packaging.version.parse)
+
+
+def _get_local_version(package: str) -> Optional[str]:
+    """Retrieve the local version of a package (if installed)."""
+    base_package = package.split("[")[0]  # remove options: package_name[options] --> package_name
+    try:
+        return importlib.metadata.version(base_package)
```
richrines1, line 198: is there a way to do this such that it reflects changes to `_version.py` in editable installs before we re-run `pip install`? (this would be the case if we just checked `<module>.__version__`, but the package metadata is only updated during installation)

perlinm: Ahh good point.  Let me try.

perlinm: Okay: I return `module.__version__` if found.

richrines1: nice thanks!

richrines1 (approved):
lgtm!

richrines1 (approved):

PR #805, opened 2023-10-04
fix incorrect line in checks-superstaq/pyproject.toml

perlinm:
I got (at least) one line wrong when copying+modifying the `pyproject.toml` file from `general-superstaq`.  Fixing it here.

### DIFF
--- a/checks-superstaq/pyproject.toml
+++ b/checks-superstaq/pyproject.toml
@@ -20,7 +20,7 @@ build-backend = "setuptools.build_meta"
 include = ["checks_superstaq*"]
 
 [tool.setuptools.package-data]
-general_superstaq = ["checks_superstaq/checks-pyproject.toml"]
+checks_superstaq = ["checks-pyproject.toml"]
 
 [tool.setuptools.dynamic.version]
 attr = "checks_superstaq._version.__version__"


checks-superstaq/pyproject.toml
```diff
@@ -20,7 +20,7 @@ build-backend = "setuptools.build_meta"
 include = ["checks_superstaq*"]
 
 [tool.setuptools.package-data]
-general_superstaq = ["checks_superstaq/checks-pyproject.toml"]
+checks_superstaq = ["checks-pyproject.toml"]
```
perlinm, line 23: [This line](https://github.com/Infleqtion/client-superstaq/blob/234d831b0701f6dc3e791a233e1cd1d50b66a27e/general-superstaq/pyproject.toml#L23) makes me think it's safe to do
```checks_superstaq = ["checks-pyproject.toml"]```
instead of
```checks_superstaq = ["checks_superstaq/checks-pyproject.toml"]```

cjguttormsson: I believe you're correct, this example from the docs shows it: https://setuptools.pypa.io/en/latest/userguide/datafiles.html#:~:text=%5Btool.setuptools.package%2Ddata%5D%0A%22*%22%20%3D%20%5B%22*.txt%22%5D%0Amypkg1%20%3D%20%5B%22data1.rst%22%5D

cjguttormsson (approved):

PR #808, opened 2023-10-09
bump version to 0.4.26

richrines1:
for now, imports all `checks-superstaq` submodules in `gss.check` to maintain backwards-compatibility. `gss.check` should be removed for good in version 0.5.0


bharat-thotakura (approved):
LGTM

PR #815, opened 2023-10-12
Update `get_balance()` return to credits

bharat-thotakura:


cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -388,22 +388,6 @@ def get_job(self, job_id: str) -> css.job.Job:
         """
         return css.job.Job(client=self._client, job_id=job_id)
 
-    def get_balance(self, pretty_output: bool = True) -> Union[str, float]:
```
richrines1, line 391: good catch

richrines1 (approved):
lgtm!

PR #816, opened 2023-10-12
Circuit metadata bug fix in `compiled_circuits()` given `index`

bharat-thotakura:
Before this PR, calling `job.compiled_circuits(index=0)` would yield an error if the job contained only a single circuit. This was because the `index` parameter returned just a `qiskit.QuantumCircuit`, leading to errors in the `compiled_circuits()` for loop in `qss` specifically during metadata assignment.


richrines1:
changes look good! how hard would it be to add a simple regression test?

bharat-thotakura:
> changes look good! how hard would it be to add a simple regression test?

Just added a new test to ensure both ways work

richrines1 (approved):
lgtm!

PR #819, opened 2023-10-16
Don't check for `$` in daily integration test

vtomole:


bharat-thotakura (approved):
My bad for missing this, LGTM!

PR #824, opened 2023-10-19
Fix passing of dd_strategy kwargs in compile

paaige:


richrines1:
needs superstaq_provider too i think

richrines1:
can we also add something like this: https://github.com/Infleqtion/client-superstaq/blob/900f50948c87090e34f08029f2751a56f8db17f6/qiskit-superstaq/qiskit_superstaq/superstaq_provider_test.py#L250-L254
(but with `dd_strategy`/etc) to `test_ibmq_compile` as a regression test?

paaige:
> can we also add something like this:
> 
> https://github.com/Infleqtion/client-superstaq/blob/900f50948c87090e34f08029f2751a56f8db17f6/qiskit-superstaq/qiskit_superstaq/superstaq_provider_test.py#L250-L254
> 
> 
> (but with `dd_strategy`/etc) to `test_ibmq_compile` as a regression test?

Added tests to all (css:service, qss:provider/backend)

richrines1:
> Added tests to all (css:service, qss:provider/backend)

thanks! does the service_test update still need to be pushed by any chance? looks like the file is unchanged

paaige:
> > Added tests to all (css:service, qss:provider/backend)
> 
> thanks! does the service_test update still need to be pushed by any chance? looks like the file is unchanged

... yeah, good catch! thanks

richrines1 (approved):
lgtm % formatting! thanks for catching this

Issue #785, opened 2023-09-21
Handle negative counts from `quasi_probability_dist`


dowusu-antwi:
### What is happening?

Because we do Measurement Error Mitigation, sometimes there are negative counts. We should convert probability distributions to the nearest possible normalized version. See [here](https://qiskit.org/ecosystem/mthree/probs.html) and [here](https://qiskit.org/documentation/stubs/qiskit.result.QuasiDistribution.html#qiskit.result.QuasiDistribution.nearest_probability_distribution).

### What are the current alternatives?

Round negative counts to zero, i.e., for `cirq.ResultDict` outputs in `cirq_superstaq`.

perlinm:
+1 for applying maximum-likelihood corrections rather than setting negative entries to zero.

teaguetomesh:
@richrines1 @vtomole - this got brought up during SiW today. Rather than opening a new issue I'll paste my reproducing example here. 

Essentially, I wanted to run the Supermarq bit code benchmarks on IBM Guadalupe, but the `Bitcode.score()` function throws an error because it expects an actual probability distribution to be passed in. Code to reproduce the error:

```python
import cirq_superstaq as css
import supermarq

service = css.Service()
bitcode_job = service.get_job('c27dfa88-8164-4130-af8f-cac83893b27a')
counts1 = bitcode_job.counts()

bitcode_bmark = supermarq.bit_code.BitCode(3, 1, [1,0,1])
bitcode_bmark.score(counts1)
```

teaguetomesh:
Two things:
1. Digging into this I realized the benchmark `score` function was failing because I had incorrectly initialized the object with a `str` rather than a `List[int]`. Opened #825 to make sure the benchmarks catch this.
2. The code to properly convert the quasidist back into a regular counts dictionary is given below. (not quite perfect because the final counts are not integers, but we could instead just return a distribution by dividing by total shots)

```python
quasdist = qiskit.result.QuasiDistribution(counts1, shots=bitcode_job.repetitions())
neardist = quasdist.nearest_probability_distribution()
correct_dist = {f'{key:0{quasdist._num_bits}b}': value for key, value in neardist.items()}
```

richrines1:
imo it was a mistake for us to just return quasi probabilities as if they were standard probabilities. i think this should be fixed in a server-side update to return both the actual and quasi probabilities in a separate fields

PR #825, opened 2023-10-19
Check for proper typing in inputs in `bit_code` and `phase_code`

teaguetomesh:
Found this bug while looking into #785, but does not close that issue.


teaguetomesh:
Actually, could someone help me get the checks to pass here? I wrote the test functions to check for invalid inputs, but that causes the GitHub actions to fail?

supermarq-benchmarks/supermarq/benchmarks/phase_code.py
```diff
@@ -21,7 +21,12 @@ class PhaseCode(Benchmark):
 
     def __init__(self, num_data_qubits: int, num_rounds: int, phase_state: List[int]) -> None:
```
bharat-thotakura, line 22: The type check is failing because `mypy` is catching the type error before the specific `ValueError`'s are raised due to the type hint `List[int]`. Doing something like:  
```suggestion
    def __init__(self, num_data_qubits: int, num_rounds: int, phase_state: typing.Sized) -> None:
```
will make the type check pass by allowing `phase_state` to accept a broader range of input types (mainly objects that support `len()`) and then raising `ValueError`'s

supermarq-benchmarks/supermarq/benchmarks/bit_code.py
```diff
@@ -20,7 +20,12 @@ class BitCode(Benchmark):
 
     def __init__(self, num_data_qubits: int, num_rounds: int, bit_state: List[int]) -> None:
```
bharat-thotakura, line 21: Likewise from earlier comment: 
```suggestion
    def __init__(self, num_data_qubits: int, num_rounds: int, bit_state: typing.Sized) -> None:
```

supermarq-benchmarks/supermarq/benchmarks/bit_code.py
```diff
@@ -20,7 +20,12 @@ class BitCode(Benchmark):
...
         if len(bit_state) != num_data_qubits:
-            raise ValueError("The length of `bit_state` must match the number of data qubits")
+            raise ValueError("The length of `bit_state` must match the number of data qubits.")
+        if not isinstance(bit_state, List):
+            raise ValueError("`bit_state` must be a List[int].")
+        else:
+            if not isinstance(bit_state[0], int):
```
bharat-thotakura, line 27: Is this supposed to only validate the first entry of `bit_state` as an `int`?

supermarq-benchmarks/supermarq/benchmarks/phase_code.py
teaguetomesh: That makes sense, but I think it would be preferable if there was a way to keep the type in the function definition as `List[int]` -- this is clearest way to tell a user what the type should be, versus using a broader type requirement and then checking for what we want in the `__init__`.

I'm not sure if that is possible though

supermarq-benchmarks/supermarq/benchmarks/bit_code.py
teaguetomesh: Haha I think there's probably a more elegant way of doing this check. Every element of the list should be an `int` but it seemed clunky to me to iterate through every single element.

supermarq-benchmarks/supermarq/benchmarks/phase_code.py
bharat-thotakura: Yeah, I thought that having `List[int]` would be more appropriate for the `__init__` and `typing.Sized` would be more appropriate for a general `validate_list_input()` function.

Looking around, it seems for our purposes, using something like `typing.cast` might get the job done for the tests while still keeping `List[int]`

supermarq-benchmarks/supermarq/benchmarks/bit_code.py
bharat-thotakura: Using `set` is efficient enough, so I think this could solve that issue: 
```suggestion
            if not set(phase_state).issubset({0, 1})::
```

supermarq-benchmarks/supermarq/benchmarks/phase_code_test.py
```diff
@@ -1,4 +1,6 @@
 # pylint: disable=missing-function-docstring,missing-class-docstring
+from typing import cast
```
bharat-thotakura, line 2: The type checker should pass with this change 
```suggestion
from __future__ import annotations

from typing import cast
```

bharat-thotakura (approved):
LGTM now

PR #827, opened 2023-10-26
Separate daily integration test so that failure can assign issue to Alex

vtomole:


.github/hilbert-integration-check-failure-issue.md
```diff
@@ -0,0 +1,4 @@
+---
+title: Hilbert integration test failed
+assignees: aradnaevinf
```
singular-value, line 3:
```suggestion
assignees: aradnaevinf, singular-value, SalahedeenIssa
```

(Since @SalahedeenIssa and I are doing more Hilbert stuff)

checks/pytest_.py
```diff
@@ -7,4 +7,8 @@
     args = sys.argv[1:]
     args += ["--exclude", "docs/source/apps/*"]
     args += ["--exclude", "docs/source/optimizations/ibm/ibmq_dd.ipynb"]
+    args += [
+        "--exclude",
+        "client-superstaq/cirq-superstaq/cirq_superstaq/hilbert_daily_integration_test.py",
+    ]
```
singular-value, line 13: should qss hilbert_daily_integration_test be excluded as well?

singular-value (changes requested):

checks/pytest_.py
vtomole: Yep. Added.

.github/hilbert-integration-check-failure-issue.md
vtomole: Done.

singular-value (approved):
lgtm if you confirm this is the human description of the desired behavior:
- every day at 14:00 UTC, we use both css and qss to compile & enqueue the described circuit to Hilbert. (_Is the Hilbert team/machine aware and ready for this?_)
- the github action will block on `job.counts` until the circuit has executed. This isn't a problem because _____. _____ = open source actions are free; also there is exponential backoff in the gss client so that the counts() calls won't flood Superstaq server?

vtomole:
> Is the Hilbert team/machine aware and ready for this?

Yes. @aradnaevinf can confirm. 

> open source actions are fre

Apparently [this is not true](https://github.com/Infleqtion/client-superstaq/pull/801) to a certain limit. We got charged a lot for sitting on Hilbert's queue for hours. https://github.com/Infleqtion/server-superstaq/pull/2938 needs to go in so jobs don't keep idling.

singular-value:
> Apparently https://github.com/Infleqtion/client-superstaq/pull/801 to a certain limit. We got charged a lot for sitting on Hilbert's queue for hours. https://github.com/Infleqtion/server-superstaq/pull/2938 needs to go in so jobs don't keep idling.

Is that true? My memory is hazy, but I thought we looked into it and we weren't actually being billed for an action from the server, not the client.

vtomole:
> I thought we looked into it and we weren't actually being billed for an action from the server

You mean that we were being billed for the server? That could be it but i remember that we were overbilled around the time when our public Gituhub actions ran for like 9 hours cause Hilbert wasn't accepting jobs.

vtomole:
Going to try out https://github.com/Infleqtion/client-superstaq/pull/845 first.

PR #835, opened 2023-11-03
Add further target validation for `submit_qubo`

bharat-thotakura:
Fixes https://github.com/Infleqtion/server-superstaq/issues/2683. 

(`gss.testing.py` was added because it will be further used by https://github.com/Infleqtion/client-superstaq/pull/788)


pnoell (commented):
LGTM but I'll leave approval to someone on the compiler team

knsmith (approved):
LG2M!

PR #841, opened 2023-11-10
don't retry on timeouts

richrines1:
fixes: https://github.com/Infleqtion/server-superstaq/issues/2650

the reason we were retrying indefinitely when jobs timed out was actually a `response.json()` call in `_handle_status_codes`. The gateway timeout response isn't a valid json string, so this would trigger a `requests.JSONDecodeError` before we'd normally throw a superstaq exception. Then because we catch these exceptions in `_make_requests`, we'd fall through and retry even though gateway_timeout isn't included in RETRIABLE_STATUS_CODES

this prevents this by putting another try..except around the `response.json()` call. i also added special handling for gateway timeouts in particular, telling users to try submitting a smaller batch


vtomole (approved):
LGTM

PR #842, opened 2023-11-10
Failing type check & `submit_qubo` daily integration test fix 

bharat-thotakura:
Updates qpu -> simulator for target name in `submit_qubo` and addresses failing type check with https://github.com/Infleqtion/client-superstaq/pull/842#issuecomment-1806334638


richrines1:
the test is failing due to a new mypy release. simple fix is to change line 488 in `qubit_gates` to:
```python
            return sum(cirq.num_qubits(g) for g in self.component_gates[:first_instance])
```

richrines1 (approved):
lgtm!

PR #845, opened 2023-11-17
Check for Hilbert availability before submitting

vtomole:
This has been pulled from https://github.com/Infleqtion/client-superstaq/pull/827 except it assigns me and doesn't separate the test. Going to try this out before merging the other PR.

### DIFF
--- a/cirq-superstaq/cirq_superstaq/daily_integration_test.py
+++ b/cirq-superstaq/cirq_superstaq/daily_integration_test.py
@@ -311,7 +311,6 @@ def test_submit_to_hilbert_qubit_sorting(service: css.Service) -> None:
     """Regression test for https://github.com/Infleqtion/client-superstaq/issues/776"""
     target = "cq_hilbert_qpu"
     num_qubits = service.target_info(target)["num_qubits"]
-
     qubits = cirq.LineQubit.range(num_qubits)
     circuit = cirq.Circuit(
         css.ParallelRGate(np.pi / 2, 0.0, 24).on(*qubits),
--- a/cirq-superstaq/cirq_superstaq/daily_integration_test.py
+++ b/cirq-superstaq/cirq_superstaq/daily_integration_test.py
@@ -303,11 +303,15 @@ def test_submit_to_provider_simulators(target: str, service: css.Service) -> Non
     assert job.counts(0) == {"11": 1}
 
 
-@pytest.mark.skip(reason="Can't be executed when Hilbert is set to not accept jobs")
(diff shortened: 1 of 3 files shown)


cirq-superstaq/cirq_superstaq/daily_integration_test.py
```diff
@@ -303,7 +303,10 @@ def test_submit_to_provider_simulators(target: str, service: css.Service) -> Non
     assert job.counts(0) == {"11": 1}
 
 
-@pytest.mark.skip(reason="Can't be executed when Hilbert is set to not accept jobs")
+@pytest.mark.skipif(
+    "cq_hilbert_qpu" not in css.Service().get_targets()["compile-and-run"],
+    reason="Can't be executed when Hilbert is set to not accept jobs",
+)
```
bharat-thotakura, line 309: Post `v0.5.0`, (see https://github.com/Infleqtion/client-superstaq/pull/788 and https://github.com/Infleqtion/server-superstaq/pull/2822) `get_targets()` no longer returns a dictionary and hence doesn't have a `["compile-and-run"]` anymore. You would have to do something like: 

```ruby
cq_target = gss.typing.Target(
        target="cq_hilbert_qpu",
        supports_submit=False,
        supports_submit_qubo=False,
        supports_compile=True,
        available=True,
        retired=False,
    )
cq_target not in css.Service().get_targets()
```
or maybe: 

```ruby
non_submit_targets = [Target.target for Target in css.Service().get_targets(supports_submit=False)]
"cq_hilbert_qpu" in non_submit_targets
```

qiskit-superstaq/qiskit_superstaq/daily_integration_test.py
```diff
@@ -247,7 +247,11 @@ def test_submit_to_provider_simulators(target: str, provider: qss.SuperstaqProvi
     assert job.result().get_counts() == {"11": 1}
 
 
-@pytest.mark.skip(reason="Can't be executed when Hilbert is set to not accept jobs")
+@pytest.mark.skipif(
+    "cq_hilbert_qpu"
+    not in qss.SuperstaqProvider()._client.get_targets()["superstaq_targets"]["compile-and-run"],
```
bharat-thotakura, line 252: See previous comment. Also, I think you should be able to call `get_targets()` directly from `qss.SuperstaqProvider()` since it inherits from `gss.service.Service()` which contains a `get_targets()` wrapper.

vtomole:
Hilbert will be offline.

PR #851, opened 2023-12-07
Return `initial_logical_to_physical` in compiler output

bharat-thotakura:
Fixes https://github.com/Infleqtion/server-superstaq/issues/2395


richrines1 (approved):
lgtm! have you tested against the server?

bharat-thotakura:
> > Let me know if you have any recommended compiled circuits to also test/check on
> 
> that looks good to me! (as long as it works in qss too). Not sure it's necessary, but fwiw another example might be a 5+ qubit ghz circuit sent to ibmq_compile with target="ibmq_brisbane_qpu" - in this case the initial and final maps should be different, and neither should be the identity

Can confirm it works in `qss` as well. Tested & compiled a 9-qubit BV circuit in `qss` & `css` with `target="ibmq_brisbane_qpu"` and the final maps were different.

Will merge!

PR #855, opened 2023-12-12
wrap pulse schedule deserialization in try..except block

richrines1:
this often fails if the user has an older version of qiskit. pulse schedules are deprecated anyway, so this shouldn't cause ibmq_compile to fail (the user should still have the corresponding pulse gate sequence)

### DIFF
--- a/qiskit-superstaq/qiskit_superstaq/compiler_output.py
+++ b/qiskit-superstaq/qiskit_superstaq/compiler_output.py
@@ -218,7 +218,10 @@ def read_json(json_dict: Dict[str, str], circuits_is_list: bool) -> CompilerOutp
         pulse_gate_circuits = qss.deserialize_circuits(json_dict["pulse_gate_circuits"])
 
     if "pulses" in json_dict:
-        pulse_sequences = gss.serialization.deserialize(json_dict["pulses"])
+        try:
+            pulse_sequences = gss.serialization.deserialize(json_dict["pulses"])
+        except Exception:
+            pulse_sequences = None
 
     if circuits_is_list:
         return CompilerOutput(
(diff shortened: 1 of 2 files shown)


qiskit-superstaq/qiskit_superstaq/compiler_output.py
```diff
@@ -218,7 +218,10 @@ def read_json(json_dict: Dict[str, str], circuits_is_list: bool) -> CompilerOutp
         pulse_gate_circuits = qss.deserialize_circuits(json_dict["pulse_gate_circuits"])
 
     if "pulses" in json_dict:
-        pulse_sequences = gss.serialization.deserialize(json_dict["pulses"])
+        try:
+            pulse_sequences = gss.serialization.deserialize(json_dict["pulses"])
+        except Exception:
+            pulse_sequences = None
```
bharat-thotakura, line 224: Just wondering -- would it be useful to have a warning raised (like in `css`) for the `Exception`?

bharat-thotakura (approved):
LGTM (w/ just a minor question)

qiskit-superstaq/qiskit_superstaq/compiler_output.py
richrines1: good question, i was debating this. My feeling was that it might be more confusing than valuable, in that we expect the user to use the pulse gate circuits (which should still be available when this fails) instead of the pulse schedules anyway - the schedules are mostly still attached to facilitate the diagrams in some of our demo notebooks, which will still work as long as they're run with the most recent qiskit version (and which we should be migrating to the timeline drawer after https://github.com/Qiskit/qiskit/issues/10582 is fixed)

bharat-thotakura: I see, that makes sense to me. PR looks good then!

PR #857, opened 2023-12-14
Modern annotations for `qss`

bharat-thotakura:
2nd part to fix https://github.com/Infleqtion/server-superstaq/issues/2969


qiskit-superstaq/qiskit_superstaq/serialization.py
```diff
@@ -15,11 +16,11 @@
 import qiskit_superstaq as qss
 
 T = TypeVar("T")
-RealArray = Union[int, float, List["RealArray"]]
+RealArray = TypeVar("RealArray")
```
richrines1, line 19: it's probably better to just delete this and change the annotation for `json_encoder` to `-> dict[str, object]` (we won't be able to use the correct alias again until we drop support for python 3.8)

bharat-thotakura: Sounds good, done

richrines1 (approved):
lgtm!

PR #856, opened 2023-12-14
Modern annotations in `css`

bharat-thotakura:
Adds modern annotations to `cirq-superstaq` (1st part of https://github.com/Infleqtion/server-superstaq/issues/2969)

PR #858, opened 2023-12-14
Remaining modern annotations update

bharat-thotakura:
3rd part to close https://github.com/Infleqtion/server-superstaq/issues/2969

(Checks require https://github.com/Infleqtion/client-superstaq/pull/856 and https://github.com/Infleqtion/client-superstaq/pull/857 to fully pass)


paaige (approved):
lgtm!

richrines1 (approved):

checks-superstaq/requirements.txt
```diff
@@ -1,5 +1,5 @@
 black[jupyter,colorama]>=23.1.0
-flake8>=3.8.4
+flake8-modern-annotations~=1.5.0
```
richrines1, line 2:
```suggestion
flake8-modern-annotations>=1.5.0
```

richrines1:
@bharat-thotakura just noticed `checks-superstaq` itself still uses the old-style annotations

it's passing the github CI flake8 check in because none of the files include `from __future__ import annotations` and the CI checks are run with python 3.8, the combination of which disables the modern annotation checker. But when e.g. i run `checks/flake8_.py` locally w/ python 3.11 it fails

to prevent this in the future we can (and imo should) enforce modern annotations everywhere by adding the following to our pyproject.toml:
```
modern-annotations-postponed = "always"
modern-annotations-deprecated = "always"
modern-annotations-union = "always"
modern-annotations-optional = "always"
```

bharat-thotakura:
Ah, I didn't realize I missed that, thanks! I just made https://github.com/Infleqtion/client-superstaq/pull/866 to fix this, but let me know if there's anything else still missing

PR #861, opened 2023-12-26
remove duplicate `qiskit-ibm-provider` req in qss/example-requirements

richrines1:
`qiskit-ibm-provider` is now part of qss's standard requirements, so doesn't need to be example-requirements too


bharat-thotakura (approved):
LGTM

PR #864, opened 2023-12-28
update some checks-superstaq reqs

richrines1:
the previous lower bounds of a few requirements were too low for our tests to pass


paaige (approved):
lgtm

PR #871, opened 2024-01-04
Add pulse gate circuits to qss.superstaqjob

paaige:


qiskit-superstaq/qiskit_superstaq/superstaq_job.py
```diff
@@ -300,6 +300,49 @@ def input_circuits(
...
+        """Gets the pulse gate circuit returned by this job.
+
+        Args:
+            index: An optional index of the pulse gate circuit to retrieve.
+
+        Returns:
+            The `qiskit.QuantumCircuit` pulse gate circuit.
+
+        Raises:
+            ValueError: If the job was not run on an IBM pulse device.
```
bharat-thotakura, line 315: Is this still relevant here since there's no `raise ValueError("Some circuits could not be deserialized.")` like in `css`?

qiskit-superstaq/qiskit_superstaq/superstaq_job.py
```diff
@@ -300,6 +300,49 @@ def input_circuits(
...
+    def pulse_gate_circuits(
+        self, index: int | None = None
+    ) -> qiskit.QuantumCircuit | list[qiskit.QuantumCircuit]:
+        """Gets the pulse gate circuit returned by this job.
+
+        Args:
+            index: An optional index of the pulse gate circuit to retrieve.
+
+        Returns:
+            The `qiskit.QuantumCircuit` pulse gate circuit.
```
bharat-thotakura, line 312: minor nit: 
```suggestion
            A single or list of pulse gate circuits.
```

qiskit-superstaq/qiskit_superstaq/superstaq_job.py
```diff
@@ -300,6 +300,50 @@ def input_circuits(
         """
         return self._get_circuits("input_circuit", index)
 
+    def pulse_gate_circuits(
+        self, index: int | None = None
+    ) -> qiskit.QuantumCircuit | list[qiskit.QuantumCircuit]:
```
bharat-thotakura, line 305: nit: It may be worth defining overloads since we do it for the other circuit methods:
 ```suggestion
    @overload
    def pulse_gate_circuits(self, index: int) -> qiskit.QuantumCircuit:
        ...
        
    @overload
    def pulse_gate_circuits(self, index: None = None) -> list[qiskit.QuantumCircuit]:
        ...
        
    def pulse_gate_circuits(
        self, index: int | None = None
    ) -> qiskit.QuantumCircuit | list[qiskit.QuantumCircuit]:
```

Comment (paaige) [reply qiskit-superstaq/qiskit_superstaq/superstaq_job.py:305]: I didn't for parity with cirq-superstaq's pulse_gate_circuit methods

Comment (richrines1) [reply qiskit-superstaq/qiskit_superstaq/superstaq_job.py:305]: i think for css it didn't really matter because the return type is just `Any` (because we don't want to assume qiskit is installed), but imo it would be nice to include the overloads here

File: qiskit-superstaq/qiskit_superstaq/superstaq_job.py [hunk 4]
```diff
@@ -300,6 +300,57 @@ def input_circuits(
...
 
+    @overload
+    def pulse_gate_circuits(self, index: int) -> qiskit.QuantumCircuit:
+        ...
+
+    @overload
+    def pulse_gate_circuits(self, index: None = None) -> list[qiskit.QuantumCircuit]:
+        ...
+
+    def pulse_gate_circuits(
```
richrines1, line 311: out of curiosity how complicated would it be to have this use `_get_circuits`? it seems like it'd have to be updated to throw an error (instead of refreshing) if  `circuit_type` is in `_job_info[job_id]` but its value is None, but then it would avoid some duplication

paaige: How do you  feel about moving that check into get_circuits? I guess a weird thing is it only applies to pulse_gate_circuits but in theory it could apply to other circuits that are only supported by specific devices in the future

richrines1: yeah i think that's fine/the way you did it seems reasonable to me!

qiskit-superstaq/qiskit_superstaq/superstaq_job.py
```diff
@@ -240,6 +240,9 @@ def _get_circuits(
         ):
             self._refresh_job()
 
+        if not all(self._job_info[job_id].get(circuit_type) is not None for job_id in job_ids):
```
richrines1, line 243: nit
```suggestion
        if any(self._job_info[job_id].get(circuit_type) is None for job_id in job_ids):
```

qiskit-superstaq/qiskit_superstaq/superstaq_job.py
```diff
@@ -300,6 +304,27 @@ def input_circuits(
...
+        ...
+
+    @overload
+    def pulse_gate_circuits(self, index: None = None) -> list[qiskit.QuantumCircuit]:
+        ...
+
+    def pulse_gate_circuits(
+        self, index: int | None = None
+    ) -> qiskit.QuantumCircuit | list[qiskit.QuantumCircuit]:
+        """Gets the pulse gate circuit returned by this job.
```
richrines1, line 318: nit
```suggestion
        """Gets the pulse gate circuit(s) returned by this job.
```

richrines1 (approved):
lgtm % docstring nit!

PR #872, opened 2024-01-04
Remove `qubovert` req in `css`

bharat-thotakura:
Currently, `qubovert>=1.2.3` is in the `requirements.txt` of `general-superstaq`, and given that `cirq-superstaq` requires `general-superstaq`, I think the explicit requirement is not needed anymore


richrines1 (approved):
good catch, thanks!

PR #874, opened 2024-01-09
pin symengine 0.11

richrines1:
symengine is required to deserialize pulse gate circuits from superstaq. qiskit only requires `>=0.9.0`, but deserialization [will fail](https://github.com/symengine/symengine/blob/bbaed4724003b882c26ef09395d87f94c1d5a451/symengine/basic.cpp#L68) if server and client aren't using the same version (up to patch)


paaige (approved):
lgtm!

Issue #873, opened 2024-01-08
pip install <qiskit/cirq>-superstaq on Windows


vietphamngoc:
### What is happening?

The installation of <qiskit/cirq>-superstaq on Windows throws the following error:

File "C:\Users\viet.phamngoc\AppData\Local\Temp\pip-build-env-k7mfbk84\overlay\Lib\site-packages\setuptools\config\expand.py", line 215, in _load_spec
          spec.loader.exec_module(module)  # type: ignore
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "<frozen importlib._bootstrap_external>", line 936, in exec_module
        File "<frozen importlib._bootstrap_external>", line 1074, in get_code
        File "<frozen importlib._bootstrap_external>", line 1004, in source_to_code
        File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
        File "C:\Users\viet.phamngoc\client-superstaq\qiskit-superstaq\qiskit_superstaq\_version.py", line 1
          ../../general-superstaq/general_superstaq/_version.py
          ^
      SyntaxError: invalid syntax

### How can we reproduce the issue?

```
conda create -n venv-qiskitsuperstaq python=3.11
conda activate venv-qiskitsuperstaq
cd client-superstaq
cd qiskit-superstaq
pip install -e ".[dev]"


### What should happen?

pip install should pass without error. Seems to be working on Linux

### Environment

- **Superstaq version**: 
- **general-superstaq version**: 
- **cirq-superstaq version**: 
- **qiskit-superstaq version**: 
- **Operating system**:  Windows 10 Pro
- **Python version**: 
- **Any other relevant dependencies**: 


### Any additional context?

_No response_

dowusu-antwi:
(@vietphamngoc discovered while we onboarded)

dowusu-antwi:
We might be forgetting to use [`os.path.normpath`](https://stackoverflow.com/questions/31629398/python-setuptools-is-stripping-slashes-from-path-arguments-on-windows) somewhere? Like in https://github.com/Infleqtion/client-superstaq/blob/768943b9a8c59a97101ec5d44055a832c1fceed0/checks-superstaq/checks_superstaq/check_utils.py#L403

richrines1:
[...]

PR #875, opened 2024-01-09
Run Window's check on `cmd`

vtomole:
Trying to reproduce https://github.com/Infleqtion/client-superstaq/issues/873


stephanielee9:
It's been reproduced

PR #879, opened 2024-01-12
add typing_extensions to type_checking_exempt_modules

perlinm:
This way something like `from typing_extensions import Self` does not need to be moved into a type checking block to keep `flake8` happy (similarly to how we make `collections.abc` exempt for the same reason)


perlinm:
@richrines1 for context: I'm actually adding this because I needed `from typing_extensions import Self` in `qLDPC`.  I thought it might be okay to add it this exemption to here instead of adding a `# REPO-SPECIFIC CONFIG`.  Having said that, I don't think any of the superstaq packages use `typing_extensions`, though, so I'd be okay closing this PR without merging if we don't want to add "clutter".

richrines1:
so i think this might be a case where the TYPE_CHECKING block makes sense - afaik typing_extensions should only need to be loaded for type checks and not during runtime, so it's probably good only load it when type checking. in principle non-dev users shouldn't even need to have typing-extensions installed to use the package (i think this isn't true atm due to its use in the `galois` package though)

(the same isn't true for `typing` or `collections.abc` because the former needs to be imported to load the `TYPE_CHECKING` variable itself, and then `collections.abc` is loaded within `typing` anyway)

perlinm:
Gocha, thanks

PR #884, opened 2024-01-25
unpin mpl

richrines1:


qiskit-superstaq/requirements.txt
```diff
@@ -1,5 +1,5 @@
 general-superstaq~=0.5.6
-matplotlib~=3.0
```
saeubank, line 2: Why do we even have matplotlib in qiskit-superstaq? Do/should we be making graphs?

supermarq-benchmarks/requirements.txt
```diff
@@ -1,4 +1,3 @@
 cirq-superstaq~=0.5.6
-matplotlib<=3.5.2
```
saeubank, line 2: I think we need this in supermarq-benchmarks.

saeubank: Used in https://github.com/Infleqtion/client-superstaq/blob/6f2ffe50658096b97062f5d9c3a5deff30938c77/supermarq-benchmarks/supermarq/plotting.py#L255

richrines1: it's already a requirement of cirq-core - all this was doing was limiting the max version (for reasons i can't really remember but which no longer seem relevant?)

qiskit-superstaq/requirements.txt
richrines1: good question. i'll try removing it

saeubank (approved):

PR #888, opened 2024-01-29
update black req

richrines1:
black versions 23.* and 24.* are not compatible, so this updates our minimum requirement to the latter (and reformats accordingly)


paaige (approved):
lgtm

PR #890, opened 2024-01-30
Remove `pydantic` upper bound

vtomole:
Not needed due to https://github.com/Infleqtion/client-superstaq/pull/883

See https://github.com/Infleqtion/client-superstaq/pull/788/files#r1373726825 for more context.


bharat-thotakura (approved):
LGTM

Issue #889, opened 2024-01-29
notebook check is very slow


richrines1:
the CI notebook check generally takes 10-15 minutes to complete (compared to 1-3 minutes for each of the other checks)

ideally we should speed up whatever steps are making it so slow, but in the interim maybe it would make sense to split it up into separate tests for `*_qss.ipynb`, `*_css.ipynb`, and "other" so at least the run in parallel?

stephanielee9:
~7 minutes now

Issue #772, opened 2023-09-15
`provider.backends()` is too slow


richrines1:
### What is happening?

running `provider.backends()` is taking 45-60s to complete. About 15s of this is the time it takes to get the list of targets (and so this is about how long `css.Service().get_targets()` takes), The remaining 30s is just due to the calls to `target_info` for each backend

imo we should be returning more info from the get_targets endpoint to obviate the need for all the subsequent target_info calls, and anything not returned as part of `get_targets` should be lazily retrieved by backends if necessary

### How can we reproduce the issue?

```python
provider = qss.SuperstaqProvider()
service = css.Service()

t0 = time.time()
_ = provider.backends()
t1 = time.time()
_ = service.get_targets()
t2 = time.time()

print("provider.get_backends():", t1 - t0)
print("service.get_targets():", t2 - t1)
```
which just gave me
```
provider.get_backends(): 44.30445456504822
service.get_targets(): 14.650813341140747
```

### What should happen?

provider.backends() shouldn't take much longer than service.get_targets(), or about 15s (ideally both should be faster)

### Environment

- **Superstaq version**: 0.4.37
- **client-superstaq version**: 0.4.22

paaige:
provider.backends has the following extra code (vs service.get_targets):
`targets = self._client.get_targets()["superstaq_targets"]
        backends = []
        for target in targets["compile-and-run"]:
            backends.append(self.get_backend(target))`
i assume the targets in `targets["compile-and-run"]`should already be included in `superstaq_targets`.

richrines1:
^ so i think this actually brings up an unrelated issue: now that `compile()` is part of `SuperstaqBackend`, `provider.get_backends()` shouldn't be limited to just compile-and-run targets

i think the ideal way to fix the performance issue would be to:
1. complete https://github.com/Infleqtion/server-superstaq/issues/2813 so we get more information from `get_targets()`
2. allow this information to be used when constructing `SuperstaqBackend` objects (e.g. with a `SuperstaqBackend.from_dict` classmethod)
3. if constructed in this way, defer the `target_info()` call to whenever the info is actually needed (instead of putting it in `backend.__init__`)

Issue #752, opened 2023-09-12
Re-evaluate exclusion of dd tutorial from pytest


singular-value:
We added this as a quickfix https://github.com/Infleqtion/client-superstaq/blob/52af3b6094038e02678e715ecbdd1c6a0afad6fe/check/pytest_.py#L10

perlinm:
For context: the issue is that the notebook instantiates an `IBMProvider`, which uses an IBM account and spends credits on that account.

vtomole:
Will be fixed implementing BYO IBM token in Superstaq.

perlinm:
Is it still possible to run (test) the notebook without providing a token though?

stephanielee9:
@knsmith we'll change submission to `dry-run` and pull up old results

vtomole:
The dd tutorial needs to be re-written to be able to remove the exclusion.

PR #892, opened 2024-01-30
Split up notebook check

bharat-thotakura:
Interim test for https://github.com/Infleqtion/client-superstaq/issues/889

(Also re-runs some notebooks for updated target outputs and adds the `available=True` flag in `_HPCA_Tutorial.ipynb`)


bharat-thotakura:
Reminder: https://github.com/Infleqtion/client-superstaq/issues/752

.github/workflows/ci.yml
```diff
@@ -238,11 +238,23 @@ jobs:
...
-      SUPERSTAQ_API_KEY : ${{ secrets.SUPERSTAQ_API_KEY }}
-      IBMQ_TOKEN : ${{ secrets.IBMQ_TOKEN }}
+      SUPERSTAQ_API_KEY: ${{ secrets.SUPERSTAQ_API_KEY }}
+      IBMQ_TOKEN: ${{ secrets.IBMQ_TOKEN }}
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - type: "qiskit-superstaq"
+            pattern: "*_qss.ipynb"
```
richrines1, line 249: notebooks from qiskit-superstaq/ should probably be included here as well
```suggestion
            pattern: "*_qss.ipynb qiskit-superstaq"
```

.github/workflows/ci.yml
```diff
@@ -238,11 +238,23 @@ jobs:
...
+            dependencies: "-e ./checks-superstaq -e ./general-superstaq[dev] -e ./qiskit-superstaq[dev]"
+          - type: "cirq-superstaq"
+            pattern: "*_css.ipynb"
```
richrines1, line 252: ditto
```suggestion
            pattern: "*_css.ipynb cirq-superstaq"
```

.github/workflows/ci.yml
```diff
@@ -238,11 +238,23 @@ jobs:
...
+            dependencies: "-e ./checks-superstaq -e ./general-superstaq[dev] -e ./cirq-superstaq[dev]"
+          - type: "other docs"
+            pattern: "--exclude=**/*_css.ipynb --exclude=**/*_qss.ipynb"
```
richrines1, line 255: assuming prior (+ cleanup)
```suggestion
            pattern: "--exclude *_css.ipynb *_qss.ipynb 'cirq-superstaq/*' 'qiskit-superstaq/*'"
```

i think all this leaves is `ibmq_dd_strategies` and the HPCA tutorial though. The former could probably just be renamed `ibmq_dd_strategies_qss` (it doesn't use css at all), and the latter should maybe be split up into qss and css versions? (definitely outside the scope of this pr though)

.github/workflows/ci.yml
```diff
@@ -238,11 +238,23 @@ jobs:
...
-      IBMQ_TOKEN : ${{ secrets.IBMQ_TOKEN }}
+      SUPERSTAQ_API_KEY: ${{ secrets.SUPERSTAQ_API_KEY }}
+      IBMQ_TOKEN: ${{ secrets.IBMQ_TOKEN }}
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - type: "qiskit-superstaq"
+            pattern: "*_qss.ipynb"
+            dependencies: "-e ./checks-superstaq -e ./general-superstaq[dev] -e ./qiskit-superstaq[dev]"
```
richrines1, line 250: imo we should also remove the actions/cache... block from these tests and change this to
```suggestion
            dependencies: "-e ./checks-superstaq -e ./general-superstaq -e ./qiskit-superstaq[examples]"
```
so this doubles as a test of our example requirements (and hopefully catches things like #885)

.github/workflows/ci.yml
```diff
@@ -238,11 +238,23 @@ jobs:
...
+          - type: "cirq-superstaq"
+            pattern: "*_css.ipynb"
+            dependencies: "-e ./checks-superstaq -e ./general-superstaq[dev] -e ./cirq-superstaq[dev]"
```
richrines1, line 253:
```suggestion
            dependencies: "-e ./checks-superstaq -e ./general-superstaq -e ./cirq-superstaq[examples]"
```

.github/workflows/ci.yml
```diff
@@ -238,28 +238,36 @@ jobs:
...
+        include:
+          - type: "qiskit-superstaq"
+            pattern: "*_qss.ipynb qiskit-superstaq"
+            dependencies: "-e ./checks-superstaq -e ./general-superstaq -e ./qiskit-superstaq[examples]"
+          - type: "cirq-superstaq"
+            pattern: "*_css.ipynb cirq-superstaq"
+            dependencies: "-e ./checks-superstaq -e ./general-superstaq -e ./cirq-superstaq[examples]"
+          - type: "other docs"
+            pattern: "--exclude *_css.ipynb *_qss.ipynb 'cirq-superstaq/*' 'qiskit-superstaq/*'"
+            dependencies: "-e ./checks-superstaq -e ./general-superstaq[dev] -e ./qiskit-superstaq[dev] -e ./cirq-superstaq[dev] -e ./supermarq-benchmarks[dev]"
```
richrines1, line 256: (sorry one more nit)
```suggestion
            dependencies: "-e ./checks-superstaq -e ./general-superstaq -e ./qiskit-superstaq[examples] -e ./cirq-superstaq[examples] -e ./supermarq-benchmarks"
```

richrines1 (approved):
this lgtm! when you're ready to merge we can switch the github rules to the new set of checks

thanks for updating this :)

bharat-thotakura:
> this lgtm! when you're ready to merge we can switch the github rules to the new set of checks
> 
> thanks for updating this :)

Thanks! I'm good to merge this but also opened https://github.com/Infleqtion/client-superstaq/pull/893/files as a brief solution to drop the 'Notebook check for other docs' now rather than later -- if https://github.com/Infleqtion/client-superstaq/pull/893 looks good. Otherwise, we can keep the current update as is.

bharat-thotakura:
Update: With https://github.com/Infleqtion/client-superstaq/pull/893 merged, this PR has been updated to only have the 'Notebook check for qiskit/cirq-superstaq' 

cc: @vtomole @richrines1

richrines1:
updated the requirements, so should be good to merge!

given that the checks still takes a while there might be an argument for splitting "supermarq" off into its own category down the line...

bharat-thotakura:
^ I'm hoping the resolution to https://github.com/Infleqtion/client-superstaq/issues/772 would void the need for a further update since 'Notebook check for cirq-superstaq' finishes much faster and the cause seems to be the call to `.backends()`

richrines1:
> ^ I'm hoping the resolution to #772 would void the need for a further update since 'Notebook check for cirq-superstaq' finishes much faster and the cause seems to be the call to `.backends()`

ahh yep that makes sense

bharat-thotakura:
Made some interim changes with [1f953b0](https://github.com/Infleqtion/client-superstaq/pull/892/commits/1f953b0ca01fc19b907e4a889d846beaf9333e0e) that should hopefully cut down the `qss` notebook check time.

supermarq-benchmarks/examples/Supermarq_HPCA_Tutorial_css.ipynb
richrines1, line 1: do you know why this file has such a large diff?

bharat-thotakura: In trying to resolve the merge conflict, I manually resolved the changes and pushed ... and that seems to have caused. Looking at the notebook differences though, nothing new should have been added (except for a markdown cell update: 'Service / Provider' -> 'Service')

PR #893, opened 2024-01-31
Clearer `qss` and `css` split for some notebooks 

bharat-thotakura:
Updates some of the notebooks (and possibly dependent files) to have clearer `qss` and `css` variants.

Motivated by https://github.com/Infleqtion/client-superstaq/pull/892#discussion_r1472119080, this PR should ideally remove the necessity of the 'Notebook check for other docs' in #892 for now


supermarq-benchmarks/supermarq/benchmarks/bit_code.py
```diff
@@ -73,6 +73,14 @@ def circuit(self) -> cirq.Circuit:
 
         return circuit
 
+    def qiskit_circuit(self) -> qiskit.QuantumCircuit:
```
richrines1, line 76: for parity, wdyt about also changing `circuit()` to `cirq_circuit()` in each benchmark subclass? then we could add the following to `supermarq.benchmark.Benchmark` so we wouldn't need to change any tests or break anything:
```python
    @abc.abstractmethod
    def cirq_circuit(self) -> cirq.Circuit | list[cirq.Circuit]:
        """Returns the cirq circuit(s) corresponding to the current benchmark parameters."""

    def circuit(self) ->  cirq.Circuit | list[cirq.Circuit]:
        """Returns the quantum circuit(s) corresponding to the current benchmark parameters."""
        return self.cirq_circuit()
```

supermarq-benchmarks/supermarq/benchmarks/bit_code.py
```diff
@@ -73,6 +73,14 @@ def circuit(self) -> cirq.Circuit:
...
+        """Generates bit code circuit.
+
+        Returns:
+            A `qiskit.QuantumCircuit`.
+        """
+        raise NotImplementedError("Use circuit() method instead.")
```
richrines1, line 82: wdyt about adding this to `supermarq.Benchmark` so we always have a fallback implementation?
```python
    def qiskit_circuit(self) ->  qiskit.QuantumCircuit | list[qiskit.QuantumCircuit]:
        """Returns the qiskit circuit(s) corresponding to the current benchmark parameters."""
        cirq_circuit = self.cirq_circuit()
        if isinstance(cirq_circuit, cirq.Circuit):
            return supermarq.converters.cirq_to_qiskit(cirq_circuit)
        return [supermarq.converters.cirq_to_qiskit(c) for c in cirq_circuit]
```

richrines1 (commented):
nice, thank you for this :) two thoughts:

supermarq-benchmarks/supermarq/benchmarks/bit_code.py
richrines1: (alternatively we could just add this to `Benchmark` and leave the subclasses alone:
```python
    def cirq_circuit(self) ->  cirq.Circuit | list[cirq.Circuit]:
        """Returns the cirq circuit(s) corresponding to the current benchmark parameters."""
        return self.circuit()
```
which imo is slightly more awkward in terms of parity but also would be a smaller change and probably fine)

bharat-thotakura: Went with option 2 as I started with that initially and seemed to require a smaller overall change

supermarq-benchmarks/supermarq/benchmarks/bit_code.py
bharat-thotakura: Done.

supermarq-benchmarks/supermarq/benchmark.py
```diff
@@ -19,14 +21,25 @@ class Benchmark:
...
     """
 
+    def cirq_circuit(self) -> cirq.Circuit | list[cirq.Circuit]:
+        """Returns the cirq circuit(s) corresponding to the current benchmark parameters."""
+        return self.circuit()
+
     @abc.abstractmethod
-    def circuit(self) -> cirq.Circuit | Sequence[cirq.Circuit]:
-        """Returns the quantum circuit corresponding to the current benchmark parameters."""
+    def circuit(self) -> cirq.Circuit | list[cirq.Circuit]:
```
richrines1, line 29: nit: can we move this either up or down so `cirq_circuit` and `qiskit_circuit` are adjacent?

supermarq-benchmarks/supermarq/converters.py
```diff
@@ -4,7 +4,7 @@
 import qiskit
 
 
-def cirq_to_qiskit(circuit: cirq.Circuit) -> qiskit.circuit.QuantumCircuit:
+def cirq_to_qiskit(circuit: cirq.Circuit) -> qiskit.QuantumCircuit:
```
richrines1, line 7: why these changes? `qiskit.QuantumCircuit` should be ok

richrines1 (approved):
lgtm % nit!

supermarq-benchmarks/supermarq/converters.py
bharat-thotakura: It was previously `qiskit.circuit.QuantumCircuit` (and locally that was picking up as `Any` sometimes -- like highlighting docstrings) so I updated it to just `qiskit.QuantumCircuit` to match the typing used in `qiskit-superstaq/`

supermarq-benchmarks/supermarq/benchmark.py
bharat-thotakura: Yeah, that would be nicer -- done

PR #902, opened 2024-02-14
♻️ compute parallelism feature as described in `arXiv:2202.11045`

### LINKED ISSUE #895: Discussion: Parallelism SupermarQ Metric
Unfortunately, this repository does not have discussions enabled because this is neither a feature request nor a bug report.
The point of discussion is the implementation of the parallelism SupermarQ metric:

https://github.com/Infleqtion/client-superstaq/blob/91d947f8cc1d99f90dca58df5248d9016e4a5345/supermarq-benchmarks/supermarq/converters.py#L71-L84

The definition being used in the code significantly differs from the one used in https://arxiv.org/abs/2202.11045, which roughly corresponds to the following pseudo code
```python
((n_g / d) - 1) / (n - 1)
```
with `n_g` being the number of gates, `d` the circuit depth, and `n` the number of qubits.

Some observations:
- While the metric from the paper depends on the number of qubits, the metric in the code does not. This leads to a situation where the paper metric would yield the same value for, e.g., a Hadamard cascade on `n` qubits independent of `n`, but the new metric yields a value that decreases with increasing `n`
- The metric from the paper is not well defined for a single qubit, but presumably the parallelism would/should be `0` in that case, right?
- While it is possible to achieve `parallelism==1.` for a non-empty circuit using the metric from the paper (`d==1`, `n_g==n`), this is no longer possible for the metric in the code since this can only ever happen if either the depth is zero (contradicting the circuit to be non-empty).

It would be much appreciated if you could briefly comment on the above.

burgholzer:
As discussed in #895, this PR refactors the implementation of the parallelism feature to align with the description from the original paper https://arxiv.org/abs/2202.11045. The implementation takes extra care to avoid corner cases where divisions by zero could happen.

Fixes #895

### DIFF
--- a/supermarq-benchmarks/supermarq/converters.py
+++ b/supermarq-benchmarks/supermarq/converters.py
@@ -71,7 +71,7 @@ def compute_liveness_with_qiskit(circuit: qiskit.QuantumCircuit) -> float:
 def compute_parallelism_with_qiskit(circuit: qiskit.QuantumCircuit) -> float:
     """Compute the parallelism of the given quantum circuit.
 
-    Parallelism feature = max(1 - depth / # of gates, 0).
+    Parallelism feature = max((((# of gates / depth) - 1) /(# of qubits - 1)), 0).
 
     Args:
         circuit: A quantum circuit.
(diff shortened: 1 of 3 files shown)


vtomole (approved):
LGTM!

stephanielee9:
Hi @burgholzer, do you have any updates on this PR?

burgholzer:
> Hi @burgholzer, do you have any updates on this PR?

Sorry for the long wait here. Somehow the message above got lost in a flood of emails.
Just added the necessary tests and rebased on main.

vtomole:
@burgholzer Looks like 1 more line needs to be covered by the test.

burgholzer:
> @burgholzer Looks like 1 more line needs to be covered by the test.

Yeah. That's on me for writing a stupid test 😓 (the `cirq.Circuit()` one also triggers the `num_qubits<=1` condition)
I am having a bit of trouble creating a 0-depth circuit with more than 1 qubit using Cirq. Do you happen to know if there is any way to do that?
In Qiskit, I would just do `QuantumCircuit(2)` and that would be it and it would cover the remaining missing line.

vtomole:
Create a test in `converters_test.py` that passes a Qiskit circuit into `compute_parallelism_with_qiskit` to get that line covered.

burgholzer:
> Create a test in `converters_test.py` that passes a Qiskit circuit into `compute_parallelism_with_qiskit` to get that line covered.

Alright. That should do it. Didn't wanna add tests in that file at first as there were basically no related tests in that file.

PR #906, opened 2024-02-21
Support for `qiskit` 1.0 

bharat-thotakura:
Adds the bare minimum changes needed to support running `qiskit` 1.0.0

Changes in this PR:

- Updates qiskit requirement to `qiskit >= 0.45.0` per [comment](https://github.com/Infleqtion/client-superstaq/pull/906#pullrequestreview-1894408423)
- Gate method updates per https://docs.quantum.ibm.com/api/migration-guides/qiskit-1.0-features#quantumcircuit-gates
- Removed `QuantumCircuit.qasm` per https://docs.quantum.ibm.com/api/migration-guides/qiskit-1.0-features#quantumcircuitqasm
- Removed `qiskit.extensions` per https://docs.quantum.ibm.com/api/migration-guides/qiskit-1.0-features#qiskitextensions


cirq-superstaq/cirq_superstaq/compiler_output.py
```diff
@@ -200,10 +200,12 @@ def read_json(json_dict: dict[str, Any], circuits_is_list: bool) -> CompilerOutp
...
             except Exception as e:
                 s = "s" if circuits_is_list else ""
-                if qiskit.__version__ < "0.24":
+                qiskit_version = qiskit.__version__
+                package_name = "qiskit" if qiskit_version >= "1.0.0" else "qiskit-terra"
+                if qiskit_version < "0.24":
                     warnings.warn(
                         f"Your compiled pulse sequence{s} could not be deserialized, likely "
-                        f"because your Qiskit Terra installation (version {qiskit.__version__}) is "
+                        f"because your {package_name} installation (version {qiskit_version}) is "
```
richrines1, line 208: should always just be "Qiskit" - the "terra" name was deprecated well before 1.0.0
```suggestion
                        f"because your Qiskit installation (version {qiskit_version}) is "
```

cirq-superstaq/cirq_superstaq/compiler_output.py
```diff
@@ -215,15 +217,15 @@ def read_json(json_dict: dict[str, Any], circuits_is_list: bool) -> CompilerOutp
                         "https://github.com/Infleqtion/client-superstaq/issues containing "
                         "the following information (as well as any other relevant context):\n\n"
                         f"cirq-superstaq version: {css.__version__}\n"
-                        f"qiskit-terra version: {qiskit.__version__}\n"
+                        f"{package_name} version: {qiskit.__version__}\n"
```
richrines1, line 220:
```suggestion
                        f"Qiskit version: {qiskit.__version__}\n"
```

richrines1 (commented):
is the 1.0.0 pin necessary? it seems like all of the changes in this pr would be compatible with `qiskit>=0.45.0`, and the linked upgrading issues shouldn't matter if e.g. someone has 0.45.1 and wants to stay on that version

(we definitely do have to drop `0.44.*` though because the qiskit.extensions gates weren't moved until `0.45.0`)

bharat-thotakura:
> is the 1.0.0 pin necessary? it seems like all of the changes in this pr would be compatible with `qiskit>=0.45.0`, and the linked upgrading issues shouldn't matter if e.g. someone has 0.45.1 and wants to stay on that version
> 
> (we definitely do have to drop `0.44.*` though because the qiskit.extensions gates weren't moved until `0.45.0`)

I was thinking about dropping the `qiskit-ibm-provider>=0.7.2` req actually, and it seems like we may need to because of the [failing](https://github.com/Infleqtion/client-superstaq/actions/runs/7995448575/job/21835808819?pr=906) qiskit-superstaq coverage check.  If we do that, then I am imagining we go back to using `qiskit.qpy.dump(circuits, buf)` but in light of https://github.com/Infleqtion/client-superstaq/blob/3ab6c9c4afa1487dff50c6a8179ccf6f09d78c10/qiskit-superstaq/qiskit_superstaq/serialization.py#L114-L117 we may want to use something like `qiskit.qpy.dump(circuits, buf, version=qiskit.qpy.common.QPY_COMPATIBILITY_VERSION)`. But `version` was only added in 1.0.0 so it may cause issues for users with qiskit 0.45.1. If not, I think you're right that we should support `qiskit>=0.45.0`.

cirq-superstaq/cirq_superstaq/compiler_output.py
bharat-thotakura: Done in [b8286ee](https://github.com/Infleqtion/client-superstaq/pull/906/commits/b8286ee831933cbacd71e16fb2774a9323ea5eb0)

cirq-superstaq/cirq_superstaq/compiler_output.py
```diff
@@ -200,10 +200,11 @@ def read_json(json_dict: dict[str, Any], circuits_is_list: bool) -> CompilerOutp
                 pulses = gss.serialization.deserialize(json_dict["pulses"])
             except Exception as e:
                 s = "s" if circuits_is_list else ""
-                if qiskit.__version__ < "0.24":
+                qiskit_version = qiskit.__version__
+                if qiskit_version < "0.24":
```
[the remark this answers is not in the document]
bharat-thotakura: Sure, seems like dependents of `qiskit-terra` usually require `>=0.23` and the reasoning previously was that "ibmq_compile requires Qiskit Terra version 0.23.* to deserialize compiled ..." (and that was just bumped up to 0.24)

richrines1 (approved):
lgtm!

PR #910, opened 2024-02-23
Remove qubovert requirement

bharat-thotakura:
Removes the qubovert requirement in `gss` as it was primarily used for typing purposes. It should be an optional install for users like in the MIS notebook.


vtomole (approved):
LGTM

PR #914, opened 2024-03-06
drop references to `toshiba_bifurcation_machine`

richrines1:
* updates tests and notebooks to reflect the unavailability of toshiba-bifurcation-machine
* sets default submit_qubo target to `ss_unconstrained_simulator`

PR #918, opened 2024-03-07
clean up qubo interface and fix tests

richrines1:
* moves result deserialization into `service.submit_qubo()`, so the user doesn't have to manually call `read_json_qubo_result()` on the result (this is what we do for every other endpoint)
* removes `gss.qubo` (currently this only contains `read_json_qubo_result()`, which can be replaced with a single line in `gss.Service`)
* removes hard-coded target validation in `submit_qubo` (which is currently tied to testing code)
* fixes various minor testing issues
* adds `gss.validation.validate_qubo` to validate qubo inputs


general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -335,9 +336,9 @@ def compile(self, json_dict: dict[str, str]) -> dict[str, str]:
 
     def submit_qubo(
         self,
-        qubo: dict[tuple[()] | tuple[str | int] | tuple[str | int, str | int], int | float],
+        qubo: Mapping[tuple[TQuboKey, ...], float],
```
bharat-thotakura, line 339: One question I have is whether we should have some client-side validation of `qubo` to check that the qubo keys have a max length of 2, i.e., only quadratic terms

richrines1: good idea, just added a validator

general-superstaq/general_superstaq/service.py
```diff
@@ -159,12 +161,12 @@ def get_targets(
 
     def submit_qubo(
         self,
-        qubo: dict[tuple[()] | tuple[str | int] | tuple[str | int, str | int], int | float],
+        qubo: Mapping[tuple[TQuboKey, ...], float],
         target: str,
         repetitions: int = 1000,
         method: str | None = None,
         max_solutions: int = 1000,
```
bharat-thotakura, line 168: The utility of `max_solutions` may need to be re-evaluated in light of [updates](https://github.com/Infleqtion/server-superstaq/blob/73e2fca00ddb8870f224d5014d1a2bfcddc982f9/superstaq/request_converter.py#L434-L437) (related to https://github.com/Infleqtion/client-superstaq/pull/914), but I think its fine as far as the scope of this PR is concerned

bharat-thotakura (approved):
Lgtm!

PR #919, opened 2024-03-07
[TEST] don't run github actions for draft prs

richrines1:


.github/workflows/ci.yml
```diff
@@ -9,6 +9,7 @@ on:
     paths-ignore:
       - "general-superstaq/general_superstaq/_version.py"
   pull_request:
+    types: [review_requested, ready_for_review]
```
bharat-thotakura, line 12: Do we have to specifiy
```suggestion
    types: [opened, synchronize, reopened, review_requested, ready_for_review]
```
also? Or are those assumed by default?

bharat-thotakura (commented):
I like having the option to run actions while in drafts, though I assume this doesn't prevent manual trigger so the PR changes should be fine. 

Though a point of clarification for me: I've noticed that the actions [just skip](https://github.com/Infleqtion/server-superstaq/actions/runs/8118049315?pr=2997) sometimes even with a manual trigger -- though presumably that happens if you also add something like ` if: github.event.pull_request.draft == false`? Or would that be case here too?

PR #922, opened 2024-03-08
Add `supports_submit_qubo` filter to `backends()`

bharat-thotakura:
Fixes an issue with `ss_unconstrained_simulator` being excluded from a `qiskit_superstaq.SuperstaqProvider().backends()` call after recent updates such as https://github.com/Infleqtion/client-superstaq/pull/914.

### DIFF
--- a/docs/source/get_started/access_info/access_info_qss.ipynb
+++ b/docs/source/get_started/access_info/access_info_qss.ipynb
@@ -110,7 +110,7 @@
     "## Backend Information\n",
     "In addition to account information, the ``SuperstaqProvider`` object also gives you a list of all the devices and simulators to which you have access, as well as additional information about those backends.\n",
     "\n",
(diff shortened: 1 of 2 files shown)


bharat-thotakura:
(The diff in the notebook is due to a ruff format automatically applied locally, the main change in it is the mention of the `supports_submit_qubo` filter -- see https://github.com/Infleqtion/client-superstaq/pull/922#issuecomment-1986011841 for visualization)

richrines1:
oh wow good catch, this is my bad :)

docs/source/get_started/access_info/access_info_qss.ipynb
richrines1, line 1: it looks like the formatter is indenting the notebook's json more than jupyter does? in which case it's probably good to undo before we merge (otherwise it will probably just be reverted in the next commit which touches this file and we'll end up with a bunch of unnecessary changes in our git history)

bharat-thotakura: Sure, will revert

PR #923, opened 2024-03-14
`measured_clbit_indices` to `classical_bit_mapping`

richrines1:
replaces `measured_clbit_indices` (which returns a list of clbit indices) with `classical_bit_mapping` (which returns a dictionary mapping those indices to the indices of the measured qubit)

### DIFF
--- a/qiskit-superstaq/qiskit_superstaq/compiler_output.py
+++ b/qiskit-superstaq/qiskit_superstaq/compiler_output.py
@@ -62,7 +62,7 @@ def measured_qubit_indices(circuit: qiskit.QuantumCircuit) -> list[int]:
 def classical_bit_mapping(circuit: qiskit.QuantumCircuit) -> dict[int, int]:
     """Returns the index of the (final) measured qubit associated with each classical bit.
 
-    If more than one measurements are assigned to the same classical bit, only the final measurement
+    If more than one measurement is assigned to the same classical bit, only the final measurement
     is considered.
 
     Args:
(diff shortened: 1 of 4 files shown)


qiskit-superstaq/qiskit_superstaq/compiler_output.py
```diff
@@ -59,28 +59,33 @@ def measured_qubit_indices(circuit: qiskit.QuantumCircuit) -> list[int]:
     return sorted(circuit.find_bit(qubit).index for qubit in measured_qubits)
 
 
-def measured_clbit_indices(circuit: qiskit.QuantumCircuit) -> list[int]:
-    """Returns the indices of the classical bits in the input quantum circuit.
+def classical_bit_mapping(circuit: qiskit.QuantumCircuit) -> dict[int, int]:
+    """Returns the index of the (final) measured qubit associated with each classical bit.
+
+    If more than one measurements are assigned to the same classical bit, only the final measurement
+    is considered.
```
bharat-thotakura, line 66: micro-nit: may sound more natural, but could just be wording preference
```suggestion
    If more than one measurement is assigned to the same classical bit, only the final measurement
    is considered.
```

bharat-thotakura (approved):
This is a useful extension, lgtm!

qiskit-superstaq/qiskit_superstaq/compiler_output.py
richrines1: agreed, thanks

PR #929, opened 2024-03-21
drop .pulse_lists from aqt_compile

richrines1:
these haven't been used (or maintained) in ages


vtomole (approved):
LGTM. Can't find `pulse_lists` in repo after grepping.

PR #931, opened 2024-03-26
Add install script for all clients

perlinm:
Lets me (or anyone else) just run `./install.sh` rather than manually running `pip install -e '.[dev]'` in each of the client projects.

PR #932, opened 2024-03-28
Install script bugfix

perlinm:
`echo cd ...` --> `cd ...`


perlinm:
Oops nvm this was already taken care of in #931.  I just had an old version of the file locally.

PR #933, opened 2024-04-04
Hilbert -> Sqorpius

vtomole:
First step in renaming Hilbert to Sqorpius. Just a find-and-replace for now. Will need to make sure the code actually runs by making the appropriate server changes.


bharat-thotakura (approved):
LGTM

PR #934, opened 2024-04-05
update minimum versions of some dependencies

richrines1:
current minimum versions of these packages are incompatible with other requirements

also updates standalone checks to confirm that all tests pass with the current minima (tests won't pass until gh settings are changed accordingly)


bharat-thotakura (approved):
LGTM

PR #942, opened 2024-05-02
add gate timing to pulse gate circuits

richrines1:
among other things enables qiskit's timeline viewer


bharat-thotakura (approved):
LGTM

PR #943, opened 2024-05-03
speed up mermin bell benchmark

richrines1:
significantly speeds up the mermin bell constructor by removing the expensive sympy logic in `_mermin_operator` (instead we can just iterate over every pauli string with an odd number of Ys, with eigenvalues `(-1)**((num_ys-1)/2)`) (something similar is described in https://arxiv.org/pdf/2005.11271)

constructor times before:
```
MerminBell(8):  0.553s
MerminBell(10): 2.998s
MerminBell(12): 18.75s
MerminBell(14): 82.12s
```
after:
```
MerminBell(8):  0.001s
MerminBell(10): 0.003s
MerminBell(12): 0.010s
MerminBell(14): 0.043s
...
MerminBell(18): 0.815s
MerminBell(20): 4.515s
```

### DIFF
--- a/supermarq-benchmarks/supermarq/benchmarks/hamiltonian_simulation.py
+++ b/supermarq-benchmarks/supermarq/benchmarks/hamiltonian_simulation.py
@@ -133,7 +133,7 @@ class HamiltonianSimulation(Benchmark):
 
         return circuit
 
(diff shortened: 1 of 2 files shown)


vtomole (approved):
LGTM

PR #944, opened 2024-05-03
adding back in module info for clients

stephanielee9:
Adding API module info back into docs (noting that this brings up the old warnings when you build docs)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -79,9 +79,6 @@ autoapi
 _build_rtd
 docs/source/modules.rst
 docs/source/setup.rst
-docs/source/general_superstaq.rst
-docs/source/cirq_superstaq.rst
-docs/source/qiskit_superstaq.rst
 docs/_build/
 
 # PyBuilder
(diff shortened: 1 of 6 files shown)


cdbf1:
I think this has actually be separately completed by https://github.com/Infleqtion/client-superstaq/pull/1030

vtomole:
image

stephanielee9:
Not a huge issue but I had left this open before because I hadn't yet figured out why these semicolons were in the docs - do we care about them?
image

cdbf1:
> Not a huge issue but I had left this open before because I hadn't yet figured out why these semicolons (between Parameters and Returns here) were in the docs - do we care about them? image

Ah yeah, these colons have been bugging me for ages and I can't work it out! I'm wondering if there is something in our css that is causing a bug somewhere

cdbf1:
Update: I have located the source of the random colons in the html. Sphinx is currently generating html like:
```
<dt class="field-even">Returns<span class="colon">:</span></dt>
```

However, removing the `<span>` tag: 
```
<dt class="field-even">Returns</dt>
```
Removes the spurious comma, while *keeping* the single comma after `Return`.

Next step is to work out how we can stop sphinx adding the `<span>` tags.

PR #946, opened 2024-05-07
add `method=` argument to GHZ benchmark

richrines1:
allows selection between a "ladder" circuit (the current implementation), "star" (the typical GHZ structure, with all CNOTs sharing the same control), and "fanout" (using a log-depth fanout circuit)


bharat-thotakura (approved):
LGTM!

PR #947, opened 2024-05-07
add benchmark classes to `benchmarks.__init__`

richrines1:
so we can use e.g. `supermarq.benchmarks.HamiltonianSimulation` instead of `supermarq.benchmarks.hamiltonian_simulation.HamiltonianSimulation`


vtomole (approved):
LGTM

PR #955, opened 2024-05-28
remove qiskit.providers.Provider inheritence

richrines1:
`qiskit.providers.Provider` [has been deprecated](https://github.com/Qiskit/qiskit/blob/473c3c2e58f855332532c2c47a11db5695c0180f/qiskit/providers/provider.py#L47-L52). we don't inherit anything from it anyway.


richrines1:
(test won't pass without #954)

Issue #958, opened 2024-06-04
IQM device provider hasn't been added yet, blocks `provider.backends()`


dowusu-antwi:
### What is happening?

IQM device is available, but access (and access to list of backends) is blocked because the provider `iqm` hasn't been added to the target prefix list.

### How can we reproduce the issue?

```
import qiskit_superstaq as qss
provider = qss.SuperstaqProvider()
provider.backends()
```

Should produce the following error:
> `ValueError: 'iqm_garnet_qpu' does not have a valid target prefix. Valid prefixes are: ['aqt', 'aws', 'cq', 'qtm', 'ibmq', 'ionq', 'oxford', 'quera', 'rigetti', 'qscout', 'ss', 'toshiba'].`

### What should happen?

Add `iqm` to target prefix list.

### Environment

- **qiskit-superstaq version**: 0.5.17
- **Operating system**: Ubuntu 22.04
- **Python version**: 3.11
- **Any other relevant dependencies**: None

PR #960, opened 2024-06-06
Remove prefix hardcoding in `validate_target()`

bharat-thotakura:
Addresses part of https://github.com/Infleqtion/client-superstaq/issues/958 by unblocking the `backends()` endpoint of `qiskit-superstaq`. 

Will also allow calls to the `compile` endpoint with `iqm_*` backends (though further optimization may be needed).

### DIFF
--- a/general-superstaq/general_superstaq/validation.py
+++ b/general-superstaq/general_superstaq/validation.py
@@ -48,6 +48,7 @@ def validate_target(target: str) -> None:
         "rigetti",
         "qscout",
         "ss",
+        "iqm",
         "toshiba",
     ]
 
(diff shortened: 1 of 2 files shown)


richrines1 (approved):
lgtm! though we should think about removing the hard-coded vendors entirely..

bharat-thotakura:
^ I can go ahead & remove the following: https://github.com/Infleqtion/client-superstaq/blob/464e9008525fe662c80b47eafadc988276e1d6e7/general-superstaq/general_superstaq/validation.py#L66-L71
I just wasn't sure what the original motivation was to have that check or what other condition we should replace it with, if any

richrines1:
i think it's fine to remove. i believe the motivation was just to catch easy errors before they hit the server, but now that our list of supported devices is somewhat dynamic this doesn't make much sense

richrines1 (approved):

PR #961, opened 2024-06-10
Disable socket

### LINKED ISSUE #962: add --disable-socket to setup.cfg after next release of pytest-socket
Currently we pass --disable-socket to pytest in pytest_.py and coverage_.py. The next version of pytest-socket will fix this by adding https://github.com/miketheman/pytest-socket/pull/186 to override the config file, so once that's released we should use it

spencerdearman:
## Adding --disable-socket with updated release of pytest-socket
- Updating the .toml file to include overall disable-socket.
- Updating the pytest_.py and coverage_.py file to use the "--force-enable-socket" flag under notebook and integration statements.

Closes #962

### DIFF
--- a/checks-superstaq/checks_superstaq/pytest_.py
+++ b/checks-superstaq/checks_superstaq/pytest_.py
@@ -63,6 +63,7 @@ def run(
     if parsed_args.notebook:
         include = include or "*.ipynb"
     elif parsed_args.integration:
+        args_to_pass += ["--force-enable-socket"]
         include = include or "*_integration_test.py"
     else:
         include = include or "*.py"
--- a/checks-superstaq/checks_superstaq/pytest_.py
+++ b/checks-superstaq/checks_superstaq/pytest_.py
@@ -61,7 +61,6 @@ def run(
     if parsed_args.notebook:
         include = include or "*.ipynb"
     elif parsed_args.integration:
-        args_to_pass += ["--force-enable-socket"]
         include = include or "*_integration_test.py"
     else:
         include = include or "*.py"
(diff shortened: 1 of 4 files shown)


checks-superstaq/checks_superstaq/pytest_.py
```diff
@@ -71,13 +70,10 @@ def run(
     files = check_utils.extract_files(parsed_args, include, exclude, silent)
 
     if parsed_args.notebook:
-        args_to_pass += ["--nbmake"]
+        args_to_pass += ["--nbmake", "--force-enable-socket"]
     elif not parsed_args.integration:
         files = check_utils.get_test_files(files, exclude=exclude, silent=silent)
```
richrines1, line 75: nit: can we keep the two `force-enable-socket` additions together this block, e.g.
```suggestion
    elif parsed_args.integration:
        args_to_pass += ["--force-enable-socket"]
    else:
        files = check_utils.get_test_files(files, exclude=exclude, silent=silent)
```
instead of adding the integration version in the block above?

spencerdearman: Updated to include change

richrines1 (commented):
looking good, just one more request: update checks-superstaq/requirements.txt to require `pytest-socket>=0.7.0` (the version that adds `--force-enable-socket`)

richrines1 (approved):
lgtm! please update the pr description to reference the original issue (e.g. "closes: #xxxx"), so the issue will automatically be closed once this is merged

once merged we'll have to deploy to make the changes are available upstream

PR #966, opened 2024-06-10
Add modular coverage check

perlinm:
In [`qLDPC/checks/coverage_.py`](https://github.com/Infleqtion/qLDPC/blob/b81bb2621dbd1289f06eb4ae3821de28b0532205/checks/coverage_.py), I enforce that every file is covered by its own test file.  I figured this capability might be nice to add to `checks_superstaq.coverage_` directly, with a `--modular` flag.

With this PR, downstream libraries can enforce modular coverage more simply with something like [this](https://github.com/Infleqtion/qLDPC/blob/5d977d5a643259fbcb9467724c535c176beee00c/checks/coverage_.py), or
```python
#!/usr/bin/env python3
import sys

import checks_superstaq

if __name__ == "__main__":
    exit(checks_superstaq.coverage_.run(*sys.argv[1:], "--modular"))
```


checks-superstaq/checks_superstaq/coverage_.py
```diff
@@ -43,14 +48,42 @@ def run(
...
 
+    if not parsed_args.modular:
+        return _run_on_files(files, test_files, pytest_args)
+
+    else:
+        # run checks on individual files, skipping repeats
+        exit_codes = {}
+        checked_test_files: list[str] = []
+        for file in files:
+            test_files = check_utils.get_test_files([file], exclude=exclude, silent=silent)
```
perlinm, line 66:
```suggestion
            test_files = check_utils.get_test_files([file], exclude=exclude, silent=True)
```

vtomole (approved):
LGTM

PR #967, opened 2024-06-12
Including weight parameter to submit_aces methods

natibek:
Gives client-side users the ability to specify the weights to be used for the ACES jobs.


cirq-superstaq/cirq_superstaq/service_test.py
```diff
@@ -766,6 +766,43 @@ def test_aces(
...
+    with pytest.raises(TypeError, match="Expected weights to be of type Iterable"):
+        service.submit_aces(
+            target="ss_unconstrained_simulator",
+            qubits=[0, 1],
+            shots=100,
+            num_circuits=10,
+            mirror_depth=5,
+            extra_depth=5,
+            noise=cirq.NoiseModel.from_noise_model_like(cirq.depolarize(0.1)),
+            weights=1,  # type: ignore
```
vtomole, line 783: Why are we type ignoring here?

natibek: Those tests check if the methods are handling unexpected types correctly by inputting incorrectly typed weight values. Ingoring the types here so mypy doesn't throw errors while running checks because the incorrect type is intended.

general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -502,7 +503,7 @@ def submit_aces(
             tag: Tag for all jobs submitted for this protocol.
             lifespan: How long to store the jobs submitted for in days (only works with right
                 permissions).
-
+            weights: The weights of the Puali strings.
```
vtomole, line 506:
```suggestion
            weights: The weights of the Pauli strings.
```

cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -944,11 +945,13 @@ def submit_aces(
             tag: Tag for all jobs submitted for this protocol.
             lifespan: How long to store the jobs submitted for in days (only works with right
                 permissions).
+            weights: The weights of the Pauli strings.
 
         Returns:
             A string with the job id for the ACES job created.
 
         Raises:
+            TypeError: If the weights are not an Iterable type.
```
vtomole, line 954: Remove this. We are not raising a `TypeError` anymore.

general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -521,8 +522,11 @@ def submit_aces(
             "extra_depth": extra_depth,
         }
 
+        if weights:
+            json_dict["weights"] = weights
         if method:
             json_dict["method"] = method
+
```
vtomole, line 529: Remove newline.

cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -961,6 +964,12 @@ def submit_aces(
         elif isinstance(noise, cirq.NoiseModel):
             noise_dict["cirq_noise_model"] = cirq.to_json(noise)
 
+        if weights is not None:
```
vtomole, line 967: @richrines1  Where did you suggest we should move this validation logic to?

richrines1: if the weights definitely have to be positive integers we could use something like:
```python
if weights is not None:
    weights = list(weights)
    for weight in weights:
        gss.validation.validate_integer_param(weight, min_val=1)
```
(imo it's probably not necessary to explicitly check that weights is Iterable - it should be obvious enough from the type annotation, and if not `list(weights)` will throw a pretty clear error if given a non-iterable anyway)

vtomole: @natibek  As @richrines1 said IRL, have the validator handle this.

vtomole (approved):
LGTM

PR #968, opened 2024-06-16
clean up modular coverage check

richrines1:
a couple of minor updates to the `--modular` coverage check (which i'm already finding useful):
1. instead of reporting each file's coverage as we check it, use `coverage run --append` on each file to combine the results into a single report at the end of the check (while still only counting lines covered by the correct test file)
2. require full coverage of the test files themselves, when they're included in the file list (currently coverage of test files is ignored in modular mode)

also adds a `--branch` flag to the coverage check (used in either mode), which requires all possible branches to be taken for full coverage (useful full for finding possible corner cases)


vtomole (approved):
LGTM

perlinm:
Nice!  Sounds like I might want to use `--append` in my projects by default.

I haven't heard of `--branch` before, but looking into it, sounds pretty strict :sweat_smile:

PR #969, opened 2024-06-18
fix typing issues

richrines1:
* fixes various typing issues in `supermarq.plotting`
* adds pydantic mypy plugin
* adds numpy and pydantic to checks-superstaq reqs (necessary for plugins)


vtomole (approved):
LGTM

PR #974, opened 2024-06-21
Integration test inconsistencies fix

spencerdearman:
Addresses inconsistencies found when running the deploy_integration test without the --integration flag


checks/pytest_.py
vtomole, line 1: You want to move this to [checks-superstaq](https://github.com/Infleqtion/client-superstaq/blob/main/checks-superstaq/checks_superstaq/pytest_.py). Otherwise downstream repos won't be able to inherit it.

vtomole:
Why not change this to 
```python
 elif (parsed_args.integration) or ("--integration" not in args and any(
        re.match(r".*_integration_test\.py$", arg) for arg in args
    ))::
        args_to_pass += ["--force-enable-socket"]
```

vtomole (commented):
`:~/client-superstaq$ ./checks/pytest_.py cirq-superstaq/cirq_superstaq/daily_integration_test.py` is not doing anything when i run it on this branch.

vtomole (approved):
Didn't work for me cause I didn't install it locally. LGTM

PR #976, opened 2024-06-24
Add mulitcore support for pytest

saeubank:
Use pytest-xdist for faster pytest tests

### DIFF
--- a/checks-superstaq/checks_superstaq/pytest_.py
+++ b/checks-superstaq/checks_superstaq/pytest_.py
@@ -58,13 +58,7 @@ def run(
     if "pytest" in parsed_args.skip:
         return 0
 
-    exclude = (
-        [exclude]
-        if isinstance(exclude, str)
-        else []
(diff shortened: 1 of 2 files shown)


checks-superstaq/checks_superstaq/pytest_.py
```diff
@@ -86,7 +86,8 @@ def run(
         integration_setup()
 
     return subprocess.call(
-        ["python", "-m", "pytest", *files, *args_to_pass], cwd=check_utils.root_dir
+        ["python", "-m", "pytest", *files, *args_to_pass, "-n=auto"],
```
richrines1, line 89: can we add `"-n=auto"` to the addopts line in pyproject.toml instead of passing it here?

richrines1 (approved):
lgtm!

it'd be nice to figure out how to do this for the coverage check too (afaik it's possible, but would definitely require a bit of configuration)

PR #978, opened 2024-07-01
Pylint plugin to check that return type annotation for `__init__` functions is `-> None`

natibek:
The return type annotation for `__init__` is inconsistently enforced by `mypy`. The plugin will warn when there is no explicit return type annotation and when the return type in the annotation is not `None`.

### DIFF
--- a/checks-superstaq/checks_superstaq/init_return.py
+++ b/checks-superstaq/checks_superstaq/init_return.py
@@ -1,0 +1,53 @@
+"""Check type type annotation of __init__ functions"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from astroid import Const, nodes
+from pylint.checkers import BaseChecker
+
+if TYPE_CHECKING:
+    from pylint.lint import PyLinter
+
+
+class InitChecker(BaseChecker):
+    """Checker for return type annotations for class constructors (__init__).
+
+    * Check that all __init__ functions have the return type annotation '-> None'.
+    """
+
+    name = "init-return-check"
+    msgs = {
+        "W9022": (
+            "Missing -> None return type annotation for __init__ method",
+            "missing-return-type-annotation-init",
+            "Missing type annotation for __init__ causes mypy check inconsistentcies.",
+        ),
+        "W9023": (
+            "Incorrect return type annotation for __init__ method. Expected None but found '%s'",
+            "incorrect-return-type-annotation-init",
+            "__init__ functions should only have return type None.",
+        ),
+    }
+
+    def visit_functiondef(self, node: nodes.FunctionDef) -> None:
+        """Called for function and method definitions (def).
+        Checks if the return type annotation for __init__ is explicity None or incorrect.
+
+        Args:
+            node: Node for a function or method definition in the AST
+        """
+        if node.name == "__init__":
+            if not node.returns:
+                self.add_message("missing-return-type-annotation-init", node=node)
+            elif not isinstance(node.returns, Const):
+                self.add_message(
+                    "incorrect-return-type-annotation-init",
+                    args=node.returns.as_string(),
+                    node=node,
+                )
+
+
+def register(linter: PyLinter) -> None:
+    linter.register_checker(InitChecker(linter))
--- a/checks-superstaq/checks_superstaq/init_return.py
+++ b/checks-superstaq/checks_superstaq/init_return.py
@@ -50,4 +50,9 @@ class InitChecker(BaseChecker):
 
 
 def register(linter: PyLinter) -> None:
+    """Registers plugin to be accessed by pylint
+
+    Args:
+        linter: The base pylinter which the custom checker will inherit from
+    """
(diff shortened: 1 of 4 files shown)


checks-superstaq/checks_superstaq/init_return.py
richrines1, line 1: can we rename this to make it obvious it's a pylint plugin? e.g. something like `pylint_init_return.py`

checks-superstaq/checks_superstaq/init_return.py
```diff
@@ -0,0 +1,58 @@
+"""Check type type annotation of __init__ functions"""
```
richrines1, line 1:
```suggestion
"""Check type annotation of __init__ functions"""
```

checks-superstaq/checks_superstaq/init_return.py
```diff
@@ -0,0 +1,58 @@
...
+        "W9023": (
+            "Incorrect return type annotation for __init__ method. Expected None but found '%s'",
+            "incorrect-return-type-annotation-init",
+            "__init__ functions should only have return type None.",
+        ),
+    }
+
+    def visit_functiondef(self, node: nodes.FunctionDef) -> None:
+        """Called for function and method definitions (def).
+        Checks if the return type annotation for __init__ is explicity None or incorrect.
```
richrines1, line 36:
```suggestion
        """Called for function and method definitions (def).

        Checks if the return type annotation for __init__ is explicitly None or incorrect.
```

checks-superstaq/checks_superstaq/init_return.py
```diff
@@ -0,0 +1,58 @@
...
+
+class InitChecker(BaseChecker):
+    """Checker for return type annotations for class constructors (__init__).
+
+    * Check that all __init__ functions have the return type annotation '-> None'.
+    """
+
+    name = "init-return-check"
+    msgs = {
+        "W9022": (
```
richrines1, line 22: out of curiosity - where did the codes come from?

checks-superstaq/checks_superstaq/init_return.py
```diff
@@ -0,0 +1,58 @@
...
+            "__init__ functions should only have return type None.",
+        ),
+    }
+
+    def visit_functiondef(self, node: nodes.FunctionDef) -> None:
+        """Called for function and method definitions (def).
+        Checks if the return type annotation for __init__ is explicity None or incorrect.
+
+        Args:
+            node: Node for a function or method definition in the AST
```
richrines1, line 39:
```suggestion
            node: Node for a function or method definition in the AST.
```

checks-superstaq/checks_superstaq/init_return.py
```diff
@@ -0,0 +1,58 @@
...
+            elif not isinstance(node.returns, Const):
+                self.add_message(
+                    "incorrect-return-type-annotation-init",
+                    args=node.returns.as_string(),
+                    node=node,
+                )
+
+
+def register(linter: PyLinter) -> None:
+    """Registers plugin to be accessed by pylint
```
richrines1, line 53:
```suggestion
    """Registers plugin to be accessed by pylint.
```

checks-superstaq/checks_superstaq/init_return.py
```diff
@@ -0,0 +1,58 @@
...
+
+    Args:
+        linter: The base pylinter which the custom checker will inherit from
```
richrines1, line 56:
```suggestion
        linter: The base pylinter which the custom checker will inherit from.
```

general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -58,7 +58,7 @@ def __init__(
         ibmq_instance: str | None = None,
         ibmq_channel: str | None = None,
         **kwargs: Any,
-    ):
+    ) -> None:
```
richrines1, line 61: nice :)

checks-superstaq/checks_superstaq/init_return.py
natibek: The largest code I found was W9021 which was being used for one of the docstring pylint checks. The next available one was W9022.

general-superstaq/general_superstaq/superstaq_client.py
natibek: That's one way to find out it works!

checks-superstaq/checks_superstaq/init_return.py
richrines1: ah ok thanks. is there any risk that a future pylint update might introduce a new warning conflicting with this?

natibek: If there is a future pylint change that leads to a new warning with the same code, pylint will throw an error whenever it is called and it will state the conflicting codes. I can check if there are conventions for what the codes should be for custom plugins so that they don't cause conflicts or if there are ways of checking for free warning codes.

richrines1 (approved):
lgtm! this is great, thanks :)

PR #983, opened 2024-07-08
License header checker and adder

natibek:
There are inconsistencies in when license headers are added to source code (and whether the added ones are the same). This PR solves this with an opt-in checker that checks if the source code has a license header and if it does, whether it is the correct one. It accounts for `shebang` lines, comments at the beginning of files unrelated to license headers, and `pylint` and `mypy` disable lines.

* Like the other checks, it can be called with `./checks/license_.py` from the root directory. 
* `-i` can be used to perform an incremental check. This is not enabled by default.
* A specific file can be passed as an argument and just that will be checked. By default, it checks all the files matching `*.py`.
* The `--apply` flag is used to fix the header problems. By default, it only checks if the headers are correct and does not make any fixes.
    * It will remove bad headers and replace them with the correct ones.
    * If no license header is found, it will add one.


richrines1 (commented):
thank you for this!! added some inline suggestions. a few comments more generally:
* there should be no references to "server" here. Any dynamic/repo-specific behavior should be handled via arguments or the config file - we would like checks in checks-superstaq to generalize outside of our repos as much as possible (within reason)
* once we correct outdated headers once, it seems like we shouldn't need to keep checking them? in which case maybe the "outdated" functionality doesn't need to live in in this script - we can save the code you use to make these initial corrections somewhere internally, and then use this script to check headers from here on out
* if we still need to explicitly check for the string "Infleqtion" in places, maybe we could add a "licensee" value to the config file in addition to the header, instead of hard coding it?
* similarly, i can maybe see why it's unavoidable but i feel like the hard-coded "apache" checks somewhat defeat the purposes of saving the header in the config. do you think there's an easy way to check if the headers are ~the same, up to licensee/year? maybe we could allow the header in pyproject.toml to include `{YEAR}` and `{LICENSEE}` tags, which we could convert to wildcards when comparing against existing licenses

also fwiw it's also ok if this script doesn't handle every possible case perfectly - if it gets confused it can always just throw an error saying to fix the headers manually :)

natibek:
> * once we correct outdated headers once, it seems like we shouldn't need to keep checking them? in which case maybe the "outdated" functionality doesn't need to live in in this script - we can save the code you use to make these initial corrections somewhere internally, and then use this script to check headers from here on out

That makes sense. We can also keep it but change the logic a bit. After the initial fix, instead of checking for ColdQuanta in the license header, we can check if it belongs to the licensee but is a different license. This can catch cases of changing the license provider.

> * similarly, i can maybe see why it's unavoidable but i feel like the hard-coded "apache" checks somewhat defeat the purposes of saving the header in the config. do you think there's an easy way to check if the headers are ~the same, up to licensee/year? maybe we could allow the header in pyproject.toml to include `{YEAR}` and `{LICENSEE}` tags, which we could convert to wildcards when comparing against existing licenses

I added a few more fields to replace the hard-coded variables. The cirq license header check pylint plugin does something similar. However, apache 2.0 licenses seem to have 2 different formattings from what I have seen in the license headers and that would mess with the matching if we use the wild card approach.

checks-superstaq/checks_superstaq/license_header_format_.py
```diff
@@ -0,0 +1,518 @@
...
+    char_count = 0
+    with open(file, "r+") as f:
+        for line_num, line in enumerate(f):
+            char_count += len(line)
+            if (
+                "Copyright" in line
+                and license_header.start_line_num <= line_num + 1 < license_header.end_line_num
+            ):
+                if line[-2] == ",":
+                    prepend += line[:-1] + f" 2024 {licensee}.\n"
```
richrines1, line 225: silly legal q: will we need to update the year in every file on 1/1/2025? or do they stay the same until if/when we update the file?

if the former we might want to replace `2024` with e.g. `datetime.datetime.now().year`. if the latter maybe we want to make the last two digits wildcards?

vtomole: > will we need to update the year in every file on 1/1/2025?

No. Given that, we can pick whichever path is easier.

natibek:
@richrines1 can you please take a look? The biggest change is that I am using `difflib.SequenceMatcher` to check if the body of the license header matches the header specified in the `pyproject.toml` file. Instead of checking if the license name (eg Apache) is in the header, we check if the header body matches the provided header body. If it matches above a threshold and the licensee is not included in the copyright line and the header is editable, the licensee is appended to the header.

checks-superstaq/checks_superstaq/license_header_format_.py
```diff
@@ -0,0 +1,546 @@
...
+        .replace('"', r"\"")
+        .strip()
+    )
+
+    pattern = re.compile(target)
+    valid = False
+
+    for license_header in license_header_lst:
+        similar_body = (
+            difflib.SequenceMatcher(None, body, license_header.license_header).ratio() > 0.94
```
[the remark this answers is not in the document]
richrines1: we also might want to generalize this a bit to determine similarity, e.g. by comparing
```python
    "".join(line.lstrip("#").strip().lower() for line in header.splitlines()),
    "".join(line.lstrip("#").strip().lower() for line in expected_header.splitlines()),
```
so that licenses will always get marked as similar if they only differ by cases/whitespace/comment style/etc

checks-superstaq/checks_superstaq/license_header_format_.py
```diff
@@ -0,0 +1,546 @@
...
+            break
+        else:
+            copyright_line += line
+    copyright_line = copyright_line.replace("{YEAR}", r"20\d{2}").replace("{LICENSEE}", licensee)
+    appended_pattern = re.compile(rf"Copyright .* 20\d{{2}} {licensee}")
+
+    target = (
+        expected_license_header.replace("{YEAR}", r"20\d{2}")
+        .replace("{LICENSEE}", licensee)
+        .replace("\n", "")
```
richrines1, line 191: why are we removing the newlines?

checks-superstaq/checks_superstaq/license_header_format_.py
```diff
@@ -0,0 +1,546 @@
...
+        .replace("(", r"\(")
+        .replace(")", r"\)")
+        .replace(".", r"\.")
+        .replace("'", r"\'")
+        .replace('"', r"\"")
```
richrines1, line 196: can we use `re.escape()` to do this?

checks-superstaq/checks_superstaq/license_header_format_.py
```diff
@@ -0,0 +1,546 @@
...
+    for license_header in license_header_lst:
+        similar_body = (
+            difflib.SequenceMatcher(None, body, license_header.license_header).ratio() > 0.94
+            or not body
+        )
+        if re.match(pattern, license_header.license_header.replace("\n", "")) and similar_body:
+            license_header.header_type = HeaderType.VALID
+            valid = True
+        elif similar_body and re.search(appended_pattern, license_header.license_header):
+            license_header.header_type = HeaderType.VALID
```
richrines1, line 212: we might want to treat this differently - if the licenses are similar but not exactly the same we should probably still rewrite them to match the expected formatting

checks-superstaq/checks_superstaq/license_header_format_.py
```diff
@@ -0,0 +1,546 @@
...
+        license_header_lst: List of the license_header objects for each header in a file.
+        licensee: The owner of the expected license.
+        editable: Whether similar licenses can be edited to include the license owner instead
+            adding the entire license to the file.
+
+    Returns: Whether there is a valid license header in a file or not.
+    """
+    copyright_line = ""
+    body = ""
+    copyright_pattern = re.compile(r"Copyright .*")
```
richrines1, line 175: should we make this one case insensitive? e.g.
```suggestion
    copyright_pattern = re.compile(r"Copyright .*", flags=re.IGNORECASE)
```

richrines1 (commented):
(partial review)

checks-superstaq/checks_superstaq/license_header_format_.py
```diff
@@ -0,0 +1,547 @@
...
+    whether similar licenses are editable.
+    """
+
+    data: dict[str, Any] = tomlkit.parse(Path("pyproject.toml").read_text())
+    try:
+        expected_license_header = str(data["tool"]["license_header_format"]["license_header"])
+    except KeyError:
+        print(
+            "Under [tool.license_header_format] add a `license_header` field filled with the"
+            "license header that should be added to source code files in the repository."
```
richrines1:
```suggestion
            "Under [tool.license_header_format] add a `license_header` field filled with the "
            "license header that should be added to source code files in the repository."
```

natibek:
@richrines1 can you take one last look at this please? I have responded to the comments.

vtomole:
From dev meeting: @natibek  to check if there is an existing tool

checks-superstaq/checks_superstaq/all_.py
```diff
@@ -98,6 +113,11 @@ def run(*args: str) -> int:
     checks_failed |= format_.run(*args_to_pass, exit_on_failure=exit_on_failure, silent=True)
     checks_failed |= lint_.run(*args_to_pass, exit_on_failure=exit_on_failure, silent=True)
 
+    checks_failed |= license_header_.run(
+        *args_to_pass,
+        exit_on_failure=exit_on_failure,
+        silent=True,
+    )
```
bharat-thotakura: Would this require non-Infleqtion users of `checks-superstaq` (if they run `./checks/all_.py`) to have a license header config?

dowusu-antwi:
@natibek do you want someone to take this over?

checks-superstaq/checks_superstaq/all_.py
bharat-thotakura: Made it an active flag to be enabled in https://github.com/Infleqtion/client-superstaq/pull/983/commits/d3e108af5c6be9b28c773a065c6f7d2467a0d001

.github/workflows/ci.yml
```diff
@@ -38,6 +38,27 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v5
+      - name: Set up Python 3.13
+        uses: actions/setup-python@v6
+        with:
+          python-version: 3.13
+      - uses: actions/cache@v4
+        with:
+          path: ${{ env.pythonLocation }}
+          key: ${{ env.pythonLocation }}-${{ hashFiles('**/pyproject.toml', '**/*requirements.txt', '.github/workflows/ci.yml') }}
```
richrines1: we're just installing checks-superstaq so probably no benefit in using the cache
```suggestion
```

checks-superstaq/checks_superstaq/license_header_.py
```diff
@@ -0,0 +1,82 @@
...
+            cwd=check_utils.root_dir,
+        )
+        if return_code > 0 and not parsed_args.fix:
+            command = "./checks/license_header_.py --fix"
+            text = (
+                f"Run '{command}' (from the repo root directory) to format "
+                "files with correct license header."
+            )
+            print(check_utils.warning(text))  # noqa: T201
+            return 1
```
richrines1: can we just fall through to return `return_code`?
```suggestion
```

checks-superstaq/checks_superstaq/licenses.py
richrines1, line 1: this shouldn't conflict with any standard packages/builtins so don't think we need the trailing underscore in the filename

also wdyt about calling this something shorter, like `licenses.py`?

checks-superstaq/checks_superstaq/license_header_.py
bharat-thotakura: Done in https://github.com/Infleqtion/client-superstaq/pull/983/commits/0dafb06e891e50f54bdb844a0c20884221002c25

checks-superstaq/checks_superstaq/licenses.py
bharat-thotakura: Oh yeah, much prefer something shorter! I didn't want to change things up too much, but updated in https://github.com/Infleqtion/client-superstaq/pull/983/commits/30656bb1db60e681d8698dc3508e1518050efcfa

bharat-thotakura:
Should this be a required check in the CI @richrines1 @vtomole?

natibek:
> Not sure if this is downstream from licensepy (cc: @natibek) or if we're okay with just running ./checks/format_.py post  ./checks/licenses.py --fix

@bharat-thotakura, Thank you for finishing the PR! I double-checked the implementation of the formatting, and it does add a new line after the updated license header in an attempt to consistently create separation between the new header and other content in the file. This was especially important to handle the cases where the wrong header is extracted from part of a larger comment. I'll test the outputs with Ruff and make changes so that it doesn't cause formatting issues.

PR #986, opened 2024-07-10
drop qiskit-ibm-provider from qss

richrines1:
qiskit-ibm-provider is deprecated, so this removes it from user-facing code/requirements. We still use it in a few places in the docs; these should also be updated so we can drop it from example-requirements too

requires us to bump our qiskit req to 1.0.0, hopefully this has been around long enough that it isn't an issue for anyone


natibek (approved):
LGTM.

PR #995, opened 2024-07-16
Import annotations checker plugin

natibek:
PyLint plugin to check if `from __future__ import annotations` is present in source code. It does not check any files with names starting with `_`.

### DIFF
--- a/checks-superstaq/checks_superstaq/checks-pyproject.toml
+++ b/checks-superstaq/checks_superstaq/checks-pyproject.toml
@@ -67,6 +67,7 @@ load-plugins = [
   "pylint.extensions.docparams", 
   "pylint.extensions.docstyle", 
   "checks_superstaq.pylint_init_return",
+  "checks_superstaq.pylint_import_annotations",
 ]
 output-format = "colorized"
 score = false
@@ -85,6 +86,7 @@ enable = [
   "duplicate-argument-name",
   "expression-not-assigned",
   "function-redefined",
+  "import-future-annotations",
   "inconsistent-mro",
   "init-return-check",
   "init-is-generator",
--- a/checks-superstaq/checks_superstaq/checks-pyproject.toml
+++ b/checks-superstaq/checks_superstaq/checks-pyproject.toml
@@ -64,8 +64,8 @@ include = ["./*"]
 max-line-length = 100
 disable = "all"
 load-plugins = [
-  "pylint.extensions.docparams", 
-  "pylint.extensions.docstyle", 
+  "pylint.extensions.docparams",
+  "pylint.extensions.docstyle",
   "checks_superstaq.pylint_init_return",
   "checks_superstaq.pylint_import_annotations",
 ]
--- a/checks-superstaq/checks_superstaq/pylint_import_annotations.py
+++ b/checks-superstaq/checks_superstaq/pylint_import_annotations.py
@@ -1,0 +1,66 @@
+"""Check for whether python file has from __future__ import annotations as the first import."""
+
+from __future__ import annotations
+
+import os
+from typing import TYPE_CHECKING
+
+from pylint.checkers import BaseChecker
+from pylint.checkers.utils import only_required_for_messages
+
+if TYPE_CHECKING:
+    from astroid import nodes
+    from pylint.lint import PyLinter
+
+
+class ImportAnnotationsChecker(BaseChecker):
+    """Checker for whether future annotations is imported in python file."""
+
+    name = "import-future-annotations"
+    msgs = {
+        "W6071": (
+            "Missing from __future__ import annotations",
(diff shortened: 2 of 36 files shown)


checks-superstaq/checks_superstaq/pylint_import_annotations.py
```diff
@@ -0,0 +1,69 @@
...
+        """Function sets the flag for import annoutations found to false for each module.
+
+        Returns nothing.
+        """
+        self.found_import_annotations = False
+
+    @only_required_for_messages("import-future-annotations")
+    def visit_importfrom(self, node: nodes.ImportFrom) -> None:
+        """Checks if 'from __future__ import annotations' is one of the imports in a module.
+        Args:
```
richrines1, line 38:
```suggestion

        Args:
```

checks-superstaq/checks_superstaq/pylint_import_annotations.py
```diff
@@ -0,0 +1,69 @@
...
+        )
+
+    def leave_module(self, node: nodes.Module) -> None:
+        """Method reports if 'from __future__ import annotations' was found in a file that does not
+        start with '_'.
+
+        Args:
+            node: The currect module
+
+        Returns nothing.
```
richrines1, line 56: not necessary
```suggestion
```

checks-superstaq/checks_superstaq/checks-pyproject.toml
```diff
@@ -67,6 +67,7 @@ load-plugins = [
   "pylint.extensions.docparams", 
   "pylint.extensions.docstyle",
```
richrines1, line 68: predates this pr, but while we're here let's get rid of these extra spaces
```suggestion
  "pylint.extensions.docparams",
  "pylint.extensions.docstyle",
```

checks-superstaq/checks_superstaq/pylint_import_annotations.py
```diff
@@ -0,0 +1,69 @@
...
+            or node.modname == "__future__"
+            and any("annotations" in name for name in node.names)
+        )
+
+    def leave_module(self, node: nodes.Module) -> None:
+        """Method reports if 'from __future__ import annotations' was found in a file that does not
+        start with '_'.
+
+        Args:
+            node: The currect module
```
richrines1, line 54:
```suggestion
            node: The current module
```

checks-superstaq/checks_superstaq/pylint_import_annotations.py
```diff
@@ -0,0 +1,69 @@
...
+    from astroid import nodes
+    from pylint.lint import PyLinter
+
+
+class ImportAnnotationsChecker(BaseChecker):
+    """Checker for whether future annotations is imported in python file."""
+
+    name = "import-future-annotations"
+    msgs = {
+        "W6071": (
```
richrines1, line 21: is there a reason not to use W6063?

checks-superstaq/checks_superstaq/pylint_import_annotations.py
```diff
@@ -0,0 +1,69 @@
...
+            "Missing from __future__ import annotations",
+            "missing-annotations-import",
+            "Missing annotations import causes compatibility issues with older python versions.",
+        ),
+    }
+
+    def visit_module(self, _: nodes.Module) -> None:
+        """Function sets the flag for import annoutations found to false for each module.
```
richrines1, line 29:
```suggestion
        """Sets the flag for import annotations found to False for each module.
```

checks-superstaq/checks_superstaq/pylint_import_annotations.py
```diff
@@ -0,0 +1,69 @@
...
+        """
+        self.found_import_annotations = False
+
+    @only_required_for_messages("import-future-annotations")
+    def visit_importfrom(self, node: nodes.ImportFrom) -> None:
+        """Checks if 'from __future__ import annotations' is one of the imports in a module.
+        Args:
+            node: All the ImportFrom nodes in a module.
+
+        Returns nothing.
```
richrines1, line 41:
```suggestion
```

richrines1 (commented):
nice! just some nits

checks-superstaq/checks_superstaq/pylint_import_annotations.py
natibek: If that plugin is ever extended, then it would be taken. Also, I was trying to replicate the grouping structure that Pylint has with their plugins where each plugin starts at `xxx1` and expands from there. However, since this is an internal plugin, it is reasonable to use `W6063` to group together all our plugins. Which one should I go with?

richrines1 (approved):
lgtm!

Issue #990, opened 2024-07-12
Make multicore optional for pytest


pgoiporia:
I've noticed that running individual checks takes much longer with multicore pytest. This is because it takes time to instantiate and collect the workers. It would be useful to have pytest default to multicore but have the option to pass a flag that indicates I want to run on a single core.

pgoiporia:
Added suggestion. Maybe we default to multicore when using `./checks/pytest_.py` but default to running on a single core when we specify any files/specific tests.

vtomole:
Duplicate of https://github.com/Infleqtion/client-superstaq/issues/9

PR #996, opened 2024-07-18
allow overriding number of pytest threads

richrines1:
changes the argument order in the subprocess call to pytest to allow manually overriding of the number of threads (or disabling multithreading altogether with `pytest_.py .... -n0`)

partially addresses #990, but as per that issue we should also consider changing the default (and maybe even choosing it dynamically based on the number of files we're testing?)


bharat-thotakura (approved):
LGTM (works for me locally)

Issue #822, opened 2023-10-18
Job cancellation not implemented


ColinCampbellCQ:
### What is happening?

Trying to cancel a job raises a NotImplementedError
![image](https://github.com/Infleqtion/client-superstaq/assets/108485140/7a8eaaea-6810-4f40-8cf7-b70e0f7495e5)

stephanielee9:
@vtomole (or perhaps @ColinCampbellCQ), by Oct 31, can you either:
* complete work to close this issue (if a week is enough time), or
* comment in this issue 1) the plan of action to close this issue, 2) a rough timeline of when you expect that work to be done, and 3) assign this issue

vtomole:
@ColinCampbellCQ  Will work on this.

ColinCampbellCQ:
I have no idea how to do this.
The best I could think of was for a CQJob on the server side to add
```
def cancel(self):
     self._job.cancel()
```
but I don't even know where to start for other types of backends or what needs to be done on the client side

stephanielee9:
@richrines1 to type up resolution / plan for closing this issue

stephanielee9:
@richrines1 Wanted to check if you'd still be able to type up a plan for closing this issue this week?

richrines1:
possible steps to fixing:
1. define a generic `SuperstaqJob.cancel(options)` method in superstaq/job.py (which should do nothing)
2. define `**Job.cancel(options)` methods in any vendor-specific `SuperstaqJob` subclasses which support job cancellation (e.g. `AzureJob`, `CQJob`)
    * `job.cancel()` should probably always check `job.get_status()` before doing anything, and only cancel the job if it's not already in a terminal state
    * for cq and ibm jobs, the `options` dictionary will need to be handled in the same way it is in `job.to_dict` (it might contain credentials necessary to access the job). i don't think it will need to be used in any of the other subclasses
3. add a `/cancel_jobs` endpoint to web/server.py, which will look very similar to `/fetch_jobs` but will call `job.cancel()` on the expanded jobs instead of `job.to_dict()`
4. define `gss._SuperstaqClient.cancel_jobs()` to reach the new endpoint (will look very similar to `_SuperstaqJob.fetch_jobs()`)
5. add `cancel()` methods to both `qss.SuperstaqJob` and `css.Job`, which will call `_client.cancel_jobs()` with a list of job ids

stephanielee9:
medium priority

bharat-thotakura:
[...]

PR #998, opened 2024-07-23
Job Cancelation

natibek:
Addressed #822.


general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -797,15 +824,17 @@ def _make_request(self, request: Callable[[], requests.Response]) -> requests.Re
...
         # Initial backoff of 100ms.
         delay_seconds = 0.1
         while True:
             try:
                 response = request()
+
+                response_content = self._handle_response(response)
                 if response.ok:
-                    return response
+                    return response_content
```
richrines1, line 837: thanks for noticing/fixing this! one thought: we probably only want to print warnings in the event of non-retriable status codes. what if we just replaced these lines in `_handle_status_codes`: https://github.com/Infleqtion/client-superstaq/blob/9c99804ba2947ca9139f79ef1347949d39660243/general-superstaq/general_superstaq/superstaq_client.py#L756-L759
to
```python
        try:
            json_content = self._handle_response(response)
        except requests.JSONDecodeError:
            json_content = None
```
(and then reverted the change to `_handle_response`, which shouldn't be necessary anymore)

general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -192,6 +191,32 @@ def get_job(self, job_id: str) -> dict[str, str]:
...
+    ) -> dict[str, list[str] | dict[str, str]] | str:
+        """Cancel jobs associated with given job ids.
+
+        Args:
+            job_ids: The UUID of the job (returned when the job was created).
+            kwargs:  Extra options needed to fetch jobs.
+                - cq_token: CQ Cloud credentials.
+
+        Returns:
+            A dictionary mapping the job id to the cancellation attempt result or an error message.
```
richrines1, line 207: might need an update: by the time we get here i think all that's left in the dict is `{"succeeded": [<ids>]}`

(also might be reasonable to just return the list of ids, instead of the full dictionary)

cirq-superstaq/cirq_superstaq/job_test.py
```diff
@@ -53,6 +53,15 @@ def mocked_get_job_requests(*job_dicts: dict[str, Any]) -> mock._patch[mock.Mock
     )
 
 
+def test_cancel(job: css.Job) -> None:
+    with mock.patch(
+        "general_superstaq.superstaq_client._SuperstaqClient.cancel_jobs", return_value=None
+    ) as mock_cancel:
```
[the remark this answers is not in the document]
natibek: The original test passes so the mock seems to be called twice. The suggestion passes too with a minor fix.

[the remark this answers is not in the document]
richrines1: great, can we make the same change in qss then?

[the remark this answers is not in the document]
natibek: Can you please explain why the new mock is better? Is the previous one stopping the program calls too early by returning `None` at `_client.cancel_jobs()`?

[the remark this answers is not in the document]
richrines1: > Is the previous one stopping the program calls too early by returning `None` at `_client.cancel_jobs()`?

exactly - in general we want to be mocking as little as possible. basically all `job.cancel()` does is call `client.cancel_jobs` - if we mock that all we're really testing is the mock itself, not the actual interface between the two methods

in this case the logic is simple enough that it probably doesn't matter. But say we updated `_client.cancel_jobs` to take a different kind of input, but forgot to update `job.cancel()` - the previous test wouldn't catch this because we were replacing its call to `_client.cancel_jobs()` with a mocked function that accepts any input (note in this example the error would hopefully be caught by mypy, but that's not always the case)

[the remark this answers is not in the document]
natibek: Got it. Thank you!

cirq-superstaq/cirq_superstaq/job.py
```diff
@@ -151,6 +151,19 @@ def status(self) -> str:
         self._update_status_queue_info()
         return self._overall_status
 
+    def cancel(self, **kwargs: object) -> None:
+        """Cancel the current job if it is not in a terminal state.
+
+        Args:
+            kwargs:  Extra options needed to fetch jobs.
```
richrines1, line 158:
```suggestion
            kwargs: Extra options needed to fetch jobs.
```

general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -192,17 +191,41 @@ def get_job(self, job_id: str) -> dict[str, str]:
...
+    def cancel_jobs(
+        self,
+        job_ids: Sequence[str],
+        **kwargs: object,
+    ) -> list[str]:
+        """Cancel jobs associated with given job ids.
+
+        Args:
+            job_ids: The UUIDs of the jobs (returned when the jobs were created).
+            kwargs:  Extra options needed to fetch jobs.
```
richrines1, line 203:
```suggestion
            kwargs: Extra options needed to fetch jobs.
```

general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -192,17 +191,41 @@ def get_job(self, job_id: str) -> dict[str, str]:
...
+
+        Returns:
+            A list of the job ids of the jobs that successfully cancelled.
+
+        Raises:
+            SuperstaqServerException: For other API call failures.
+        """
+        json_dict: dict[str, Any] = {
```
richrines1, line 211:
```suggestion
        json_dict: dict[str, list[str]] = {
```

qiskit-superstaq/qiskit_superstaq/superstaq_job.py
```diff
@@ -177,6 +177,19 @@ def _check_if_stopped(self) -> None:
                 self._job_id, self._overall_status
             )
 
+    def cancel(self, **kwargs: object) -> None:
+        """Cancel the current job if it is not in a terminal state.
+
+        Args:
+            kwargs:  Extra options needed to fetch jobs.
```
richrines1, line 184:
```suggestion
            kwargs: Extra options needed to fetch jobs.
```

richrines1 (commented):
lg % a few nits

general-superstaq/general_superstaq/superstaq_client.py
richrines1: ah my bad. what about this?
```suggestion
        json_dict: dict[str, str | Sequence[str]] = {
```

richrines1 (approved):
lgtm!

PR #1002, opened 2024-07-29
Add get_user_info functionality

### LINKED ISSUE #1001: Add functionality to get user info from the server
### What is happening?

Simple functionality for a user to retrieve their user info from the server

_No response_

### Any additional context?

_No response_

cdbf1:
Add functionality to query the new `/get_user_info` server endpoint. 

Closes #1001


natibek:
This looks good but there are a few docstring nits. pylint was supposed to catch them but we found out we did not enable the optional checks in the `docparams` plugin to check if docstrings have information about what functions return and raise.

general-superstaq/general_superstaq/service.py
```diff
@@ -159,6 +159,43 @@ def get_targets(
...
+        """Gets a dictionary of the user's info.
+
+        .. note::
+
+            SUPERTECH users can submit optional :code:`name` and/or :code:`email` keyword only
+            arguments which can be used to search for the info of arbitrary users on the server.
+
+        Args:
+            name: Defaults to None.
+            email: str
```
natibek, line 186: missing return documentation.

general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -227,6 +227,38 @@ def get_balance(self) -> dict[str, float]:
...
+                Is only available to SUPERTECH users. Defaults to None.
+        """
+        if query is None:
+            user_info = self.get_request("/get_user_info", query={})
+            try:
+                return list(user_info.values())[0]
+            except IndexError:
+                raise gss.SuperstaqServerException(
+                    "Something has gone wrong, the server has returned an empty response."
+                )
```
richrines1, line 257: can this actually occur?

if so, let's just check if the dict is empty directly instead of using a try..except, e.g.:
```python
if not user_info:
    raise...
return ...
```

general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -227,6 +227,38 @@ def get_balance(self) -> dict[str, float]:
...
+        .. note::
+
+            SUPERTECH users can submit an optional :code:`query` argument which can be used
+            to search for the info of arbitrary users on the server.
+
+        Args:
+            query: An optional dictionary of query parameters to use to search for user records.
+                Is only available to SUPERTECH users. Defaults to None.
+        """
+        if query is None:
```
richrines1, line 250: imo this return logic should be moved to `Service.user_info` so this function always returns the same type (and we don't have to overload both signatures)

in that case it might also be simpler to move the construction of `query` here, so this would have the signature
```python
get_user_info(self, name: str | None, email: str | None) -> dict[str, dict[str, object]]:
```

general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -669,11 +702,19 @@ def request() -> requests.Response:
             Returns:
                 The Flask GET request object.
             """
-            return requests.get(
-                f"{self.url}{endpoint}",
-                headers=self.headers,
-                verify=self.verify_https,
-            )
+            if query is None:
```
richrines1, line 705: is this necessary? it seems like passing `json=None` should be the same as not passing it

general-superstaq/general_superstaq/service.py
```diff
@@ -159,6 +159,43 @@ def get_targets(
         )
         return self._client.get_targets(**filters)
 
+    @overload
+    def get_user_info(self) -> dict[str, object]: ...
+
+    @overload
+    def get_user_info(self, *, name: str) -> dict[str, dict[str, object]]: ...
```
richrines1, line 166: would it make sense to convert this to `list[dict[str, object]]`, given that the keys of the outer dict are already included within the corresponding values?

(also can `object` be replaced with something like `str | float`?)

general-superstaq/general_superstaq/superstaq_client.py
cdbf1: It shouldn't occur, but I would rather catch the error if it did

richrines1 (approved):
lgtm!

PR #1004, opened 2024-07-29
Enable docparams checks

natibek:
Although we import the `docparams` plugin from pylint's extensions, the checks for ensuring docstrings contain return, yield, parameter, and raise documentation are optional and not enabled. We have been enforcing the inclusion of these through pr reviews.

### DIFF
--- a/cirq-superstaq/cirq_superstaq/daily_integration_test.py
+++ b/cirq-superstaq/cirq_superstaq/daily_integration_test.py
@@ -16,6 +16,11 @@ import cirq_superstaq as css
 
 @pytest.fixture
 def service() -> css.Service:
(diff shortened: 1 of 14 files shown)


cirq-superstaq/cirq_superstaq/daily_integration_test.py
```diff
@@ -16,6 +16,11 @@
 
 @pytest.fixture
 def service() -> css.Service:
+    """Fixture for cirq_suerstaq service.
```
vtomole, line 19:
```suggestion
    """Fixture for cirq_superstaq service.
```

vtomole (approved):

PR #1007, opened 2024-07-31
Add option to disable multicore and auto disable when only 1 file

### LINKED ISSUE #990: Make multicore optional for pytest

saeubank:
Closes #990

### DIFF
--- a/checks-superstaq/checks_superstaq/pytest_.py
+++ b/checks-superstaq/checks_superstaq/pytest_.py
@@ -53,6 +53,11 @@ def run(
         action="store_true",
         help="Run pytest on *_integration_test.py files.",
     )
+    parser.add_argument(
+        "--single_core",
+        action="store_true",
+        help="Run pytest without using xdist.",
+        )
 
     parsed_args, args_to_pass = parser.parse_known_intermixed_args(args)
     if "pytest" in parsed_args.skip:
@@ -82,11 +87,14 @@ def run(
     if not files:
         return 0
 
+    if not parsed_args.single_core and len(files) != 1:
+        args_to_pass += ["-n=auto"]
+
     if parsed_args.integration and integration_setup:
         integration_setup()
 
     return subprocess.call(
-        ["python", "-m", "pytest", "-n=auto", *files, *args_to_pass],
+        ["python", "-m", "pytest", *files, *args_to_pass],
         cwd=check_utils.root_dir,
     )
 
--- a/checks-superstaq/checks_superstaq/pytest_.py
+++ b/checks-superstaq/checks_superstaq/pytest_.py
@@ -87,8 +87,10 @@ def run(
     if not files:
         return 0
 
-    if not (parsed_args.single_core or parsed_args.incremental):
-        args_to_pass += ["-n=auto"]
+    # Dont mulitcore if checking fewer than 4 files
+    if not (parsed_args.single_core or len(files) < 4):
+        # setting before other args so -n can be overwritten
+        args_to_pass = ["-n=auto", *args_to_pass]
 
     if parsed_args.integration and integration_setup:
         integration_setup()
--- a/checks-superstaq/checks_superstaq/pytest_.py
+++ b/checks-superstaq/checks_superstaq/pytest_.py
@@ -87,7 +87,7 @@ def run(
     if not files:
         return 0
 
-    if not parsed_args.single_core and len(files) != 1:
+    if not (parsed_args.single_core or parsed_args.incremental):
         args_to_pass += ["-n=auto"]
 
     if parsed_args.integration and integration_setup:
--- a/checks-superstaq/checks_superstaq/pytest_.py
+++ b/checks-superstaq/checks_superstaq/pytest_.py
@@ -57,7 +57,7 @@ def run(
         "--single_core",
         action="store_true",
         help="Run pytest without using xdist.",
-        )
+    )
 
     parsed_args, args_to_pass = parser.parse_known_intermixed_args(args)
     if "pytest" in parsed_args.skip:
--- a/checks-superstaq/checks_superstaq/pytest_.py
+++ b/checks-superstaq/checks_superstaq/pytest_.py
@@ -87,8 +87,13 @@ def run(
     if not files:
         return 0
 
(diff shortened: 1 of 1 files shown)


checks-superstaq/checks_superstaq/pytest_.py
```diff
@@ -82,11 +87,14 @@ def run(
     if not files:
         return 0
 
+    if not parsed_args.single_core and len(files) != 1:
+        args_to_pass += ["-n=auto"]
```
richrines1, line 91: this should be put before the other `args_to_pass` so command line arguments will take precedence (for example currently i can pass `./checks/pytest_.py -n0` to override `-n=auto` and disable threading)

saeubank: you can disable threading with the new `--single_core` command.

Also are you referring to `parser.parse_known_intermixed_args` for the other `args_to_pass`?

checks-superstaq/checks_superstaq/pytest_.py
```diff
@@ -82,11 +87,14 @@ def run(
     if not files:
         return 0
 
+    if not parsed_args.single_core and len(files) != 1:
```
richrines1, line 90: do we want to disable threading whenever any files are manually passed in (per [@pgoiporia's suggestion](https://github.com/Infleqtion/client-superstaq/issues/990#issuecomment-2231300487))? e.g.
```suggestion
    if not (parsed_args.single_core or parsed_args.files or parsed_args.incremental):
```

saeubank: Oh, I misread that suggestion. For some reason I thought it said when there is only 1 file.

But that suggested change wont work. `parsed_args.files` always has values in it or the test wont run.
I'll change it to exclude incremental though.

checks-superstaq/checks_superstaq/pytest_.py
richrines1: sure but currently i can also pass in `-n2`/`-n5`/etc :) and also it's nice to have the same arguments work whether i'm using `./checks/pytest_.py` or `pytest` directly. This would work if we just change this to e.g.
```suggestion
        args_to_pass = ["-n=auto", *args_to_pass]
```

(im just referring to the other values in `args_to_pass` - which contains any arguments i pass in to `./checks/pytest_.py` that aren't recognized by the parser)

checks-superstaq/checks_superstaq/pytest_.py
richrines1: > `parsed_args.files` always has values in it or the test wont run

is this true? when i run `./checks/pytest_.py` the tests will run even though `parsed_args.files` should be empty i think

though maybe another option is to disable threading below some threshold `len(files)` - for example if im only testing 2-4 files it usually seems faster to run them serially

checks-superstaq/checks_superstaq/pytest_.py
saeubank: Ah, I see. I'll change this

checks-superstaq/checks_superstaq/pytest_.py
saeubank: > is this true? when i run ./checks/pytest_.py the tests will run even though parsed_args.files should be empty i think

The script auto populates `files` with all the valid test files and quits if it doesnt find any.


So it looks like checking for `incremental` doesnt work either for some reason, (not sure why since I see it in `check_utils` which we import from)
I get this error
```
if not (parsed_args.single_core or parsed_args.incremental):
                                       ^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'Namespace' object has no attribute 'incremental'
```

> though maybe another option is to disable threading below some threshold len(files) - for example if im only testing 2-4 files it usually seems faster to run them serially

We can do this for now in this PR, but I think some of the test files can have a lot of tests so mileage may vary based on which test files are being run. (also ofc is hardware dependent)

richrines1: > The script auto populates `files` with all the valid test files and quits if it doesnt find any.

yep `files` will be populated but `parsed_args.files` should still just contain the files actually passed in i think. But that said e.g. `./checks/pytest_py cirq-superstaq` would count as "passing in a file" even though it checks all of cirq-superstaq, so maybe this isn't ideal anyway?

> So it looks like checking for `incremental` doesnt work either for some reason, (not sure why since I see it in `check_utils` which we import from) I get this error

oops my bad, it needs to be `parsed_args.revisions is not None` - it's redirected by the `dest=` line here: https://github.com/Infleqtion/client-superstaq/blob/77922748c02172f711e5ff3585b0e168a7e8c482/checks-superstaq/checks_superstaq/check_utils.py#L348
(but maybe we don't need to do this at all if we're adding a threshold anyway?)

> We can do this for now in this PR, but I think some of the test files can have a lot of tests so mileage may vary based on which test files are being run. (also ofc is hardware dependent)

yeah fair enough.. it would be nice to have a better heuristic for this. I'm also fine with making the threshold lower, as long as it's disabled when `-s` is passed in/can be disabled manually otherwise

saeubank: Actually just checking for the `-s` should be good right? It makes the logic a lot simpler too.

richrines1: i think that's definitely the most important case, but it would be really nice if it was disabled by default for small numbers of files too

coming back to this:

> that said e.g. ./checks/pytest_py cirq-superstaq would count as "passing in a file" even though it checks all of cirq-superstaq, so maybe this isn't ideal anyway?

i just tried it locally and checking all of cirq-superstaq is still faster with xdist disabled on my system (3s vs. 5.9s). so maybe something like this would make sense here?
```python
if not parsed_args.files and parsed_args.revisions is None and "-s" not in args_to_pass:
    # enable threading...
```
(so basically it will only be activated when testing the full repo without `-s`, in which case we definitely do want it)

richrines1 (approved):
lgtm! thanks :)

PR #1010, opened 2024-08-02
lazy-load qss backend configuration

### LINKED ISSUE #772: `provider.backends()` is too slow

richrines1:
fixes #772

### DIFF
--- a/qiskit-superstaq/qiskit_superstaq/conftest.py
+++ b/qiskit-superstaq/qiskit_superstaq/conftest.py
@@ -2,45 +2,8 @@ from __future__ import annotations
 
 import general_superstaq as gss
 import pytest
-import qiskit
 
 import qiskit_superstaq as qss
-
-
-class MockSuperstaqBackend(qss.SuperstaqBackend):
-    """Stand-in for `SuperstaqBackend` that the tests can call."""
-
-    def __init__(self, provider: qss.SuperstaqProvider, target: str) -> None:
-        """Initializes a `SuperstaqBackend`.
-
-        Args:
-            provider: Provider for a Superstaq backend.
-            target: A string containing the name of a target backend.
-        """
-        self._provider = provider
-        self.configuration_dict = {
-            "backend_name": target,
-            "backend_version": "n/a",
-            "n_qubits": -1,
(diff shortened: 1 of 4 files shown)


bharat-thotakura (approved):
LGTM

(Testing locally, `get_backends()` and `get_targets()` are now comparable in terms of time taken!)

bharat-thotakura:
Should we revert https://github.com/Infleqtion/client-superstaq/pull/892/commits/1f953b0ca01fc19b907e4a889d846beaf9333e0e after merging this PR?

richrines1:
> Should we revert [1f953b0](https://github.com/Infleqtion/client-superstaq/commit/1f953b0ca01fc19b907e4a889d846beaf9333e0e) after merging this PR?

ah good question. i think the first change in that commit should be reverted (and is small enough that i'll just add it to this pr). i kinda feel like the others are maybe nicer as they are though? imo the `get_targets` output is more interesting/informative than showing the output of `backends()` now.. wdyt?

bharat-thotakura:
Yeah sure, I'd agree with that and I guess it keeps both the `css` and `qss` notebook outputs w.r.t. targets the same

PR #1012, opened 2024-08-02
barriers should never commute

richrines1:
adds `_commutes_` and `_commutes_on_qids_` magic methods to `css.Barrier` to indicate that barriers should never commute

(also adds one line in `job_test.py` to prevent a superfluous warning from appearing in the pytest output)


bharat-thotakura (approved):
LGTM

Issue #897, opened 2024-02-07
External contributor PRs can't pass notebook checks


vtomole:
### What is happening?

External PRs don't have access to the Superstaq API keys so the notebook checks will fail.

### How can we reproduce the issue?

Fork `client-superstaq`  and make a PR with a non-consequential change.

### What should happen?

All checks should pass.

### Environment

`client-superstaq` repo.

bharat-thotakura:
Wait for verification from https://github.com/Infleqtion/client-superstaq/pull/1232 before closing

bharat-thotakura:
Completed in https://github.com/Infleqtion/client-superstaq/pull/1232

PR #1016, opened 2024-08-08
Add option to use `ruff` for formatting and linting

perlinm:
[Ruff](https://docs.astral.sh/ruff/) ([GitHub](https://github.com/astral-sh/ruff)) is a Python linter and formatter that plays the role of `black`, `isort`, `flake8`, and `pylint`, and does it all much faster (for whatever that is worth).  This PR adds the **option** to use `ruff` in place of these tools to `checks-superstaq`.

Specifically, this PR functionally adds:
- `ruff_format_.py` (which aspires to replace one day replace `format_.py`)
- `ruff_lint_.py` (which aspires to one day replace `flake8_.py` and `pylint_.py`, at which point it can be renamed to (say) `lint_.py`).
- A `--ruff` flag to `checks_superstaq.all_` to use `ruff` instead of the previous formatters+linters.

Note that `ruff_format_.py` has a `--fix` option instead of `--apply` for parity with the `--fix` option recognized by `ruff check` (which is called by `ruff_lint_.py`).  That's because `ruff format` [refuses to change a program's AST](https://github.com/astral-sh/ruff/issues/8926#issuecomment-1834048218), so it's not allowed to change import order.  Import order therefore has to be handled by `ruff check` with `./checks/ruff_lint_.py --fix`.  (Incidentally, maybe `ruff_lint_.py` should be renamed to `ruff_check_` because it calls `ruff check`?  But then if+when this repo switches to using ruff it will feel silly to call `checks/check_.py`.  Or maybe if+when `ruff` takes over we can have `ruff_format.py` and `ruff_check.py`?)

`ruff` comes with many reasonable defaults, so I did not need to add much to `pyproject.toml`.  Having said that, `checks-superstaq` has extensive configuration of `flake8` and `pylint`, and I am sure that not all of these linting rules are enforced by ruff.  I leave it up to you all @Infleqtion to make the call on whether it's okay to merge in a "partial" switch to ruff (at the cost of having duplicate formatting/linting tools in the repo) vs. whether you'd like to do one big switch (configs and all).

At the moment, all checks of `ruff_lint_.py` pass, and `ruff_format_.py` asks for a few minor formatting changes.  The requested formatting changes are not included in this PR (though if I do make the changes, `checks/format_.py` is happy with all except those made to one `.pynb` file).


vtomole (approved):
LGTM! Let's create an issue to add a ruff Github action to `client-superstaq`.

perlinm:
Any idea why the qiskit notebook check is failing?  I don't think this PR should have affected that.

bharat-thotakura:
It's probably failing because of https://github.com/Infleqtion/client-superstaq/issues/897. Though we can still merge after @vtomole adjusts the requirements I think

vtomole:
@perlinm I've adjusted the requirements.

Issue #1006, opened 2024-07-30
Support $SU(2)$ benchmarking


dowusu-antwi:
### What is happening?

Add SU(2) benchmarking to Superstaq/Supermarq QCVVB tools. E.g., see Fig. 2 in Lukin paper [here](https://www.nature.com/articles/s41586-023-06481-y):
![image](https://github.com/user-attachments/assets/f51a7ee2-ac8a-454b-b38a-bd13f93836ca)

### What are the current alternatives?

N/A. Should be facilitated by https://github.com/Infleqtion/client-superstaq/pull/992.

### Any additional context?

SU(2) is particularly well-suited to neutral atom devices. Adding this supports "write-once, target-any".

cdbf1:
I took a look at this and I don't think the experiment in Fig 6 of that paper is quite what we want for a general fidelity characterization protocol, so instead I've implemented the protocol they have in Fig 3 which seems to work quite nicely.

We could separately look into what they do in Fig 6 as a method for characterizing phase errors.

PR #1017, opened 2024-08-08
SU2 benchmarking

cdbf1:
Add SU2 benchmarking to the QCVV library

Closes: #1006


dowusu-antwi (commented):
Some nits:

supermarq-benchmarks/supermarq/qcvv/su2.py
```diff
@@ -0,0 +1,269 @@
...
+
+        Args:
+            include_two_qubit_gate: Whether to include the two qubit gate being measured
+
+        Returns:
+            The sub circuit to be repeated when building the full circuit
+        """
+        return cirq.Circuit(
+            self._haar_random_rotation().on(self.qubits[0]),
+            self._haar_random_rotation().on(self.qubits[1]),
```
richrines1, line 232: i believe the same rotation is supposed to be applied to both gates, so that they can be done with a single global pulse on systems that support it (the language in the paper is "random global single-qubit rotations between sequences of CZ entangling gates")

cdbf1: Sorry only just seen this comment.
Yes I agree, thanks. I must have misinterpreted the paper. Now fixed 😃

cdbf1:
Update 10th March - fixed conflicts with updated QCVV framework
Ready for re-review

supermarq-benchmarks/supermarq/qcvv/su2.py
```diff
@@ -0,0 +1,365 @@
...
+        """
+        if self.data is None:
+            raise RuntimeError("No data stored. Cannot perform analysis.")
+
+        fit = linregress(
+            x=self.data["num_two_qubit_gates"],
+            y=np.log(4 / 3 * (self.data["00"] - 1 / 4)),
+            # Scale the y coordinate to account for limit of the decay being 1/4
+        )
+        gate_fid = np.exp(fit.slope)
```
dowusu-antwi, line 159: Should we validate that `gate_fid` is in `[0, 1]`, or is this done elsewhere, or perhaps implied by construction? We would want users to still see such a result for experiment debugging purposes, but even then maybe we should raise an exception.

supermarq-benchmarks/supermarq/qcvv/su2.py
```diff
@@ -0,0 +1,365 @@
...
+        Returns:
+            A named tuple of the final results from the experiment.
+        """
+        if self.data is None:
+            raise RuntimeError("No data stored. Cannot perform analysis.")
+
+        fit = linregress(
+            x=self.data["num_two_qubit_gates"],
+            y=np.log(4 / 3 * (self.data["00"] - 1 / 4)),
+            # Scale the y coordinate to account for limit of the decay being 1/4
```
dowusu-antwi: Should we validate that `4 / 3 * (self.data["00"] - 1 / 4) > 0`?

μ-nit: Magic numbers -- the `1/4` is explained but should we also include commentary on the `4/3`? IIRC it's because of [how depolarizing error is defined](https://quantumai.google/reference/python/cirq/depolarize):
image

supermarq-benchmarks/supermarq/qcvv/su2.py
```diff
@@ -0,0 +1,365 @@
...
+        Returns:
+            The sub circuit to be repeated when building the full circuit
+        """
+        # Chose a global rotation
+        rotation = self._haar_random_rotation()
+        return cirq.Circuit(
+            rotation.on(self.qubits[0]),
+            rotation.on(self.qubits[1]),
+            (
+                self.two_qubit_gate(*self.qubits).with_tags("no_compile")
```
dowusu-antwi: I wonder if users might get the wrong idea and think the `no_compile` tag means we always simulate it. If we document it, we can say something in `base_experiment.py`.

supermarq-benchmarks/supermarq/qcvv/su2.py
```diff
@@ -0,0 +1,365 @@
...
+
+    @staticmethod
+    def _haar_random_rotation() -> cirq.Gate:
+        """Returns:
+        Haar randomly sampled SU(2) rotation.
+        """
+        gate: cirq.Gate | None = None
+        while gate is None:
+            gate = cirq.single_qubit_matrix_to_phxz(cirq.testing.random_special_unitary(dim=2))
+        return gate
```
dowusu-antwi: Is there a way to make sure we recognize a potential infinite loop here? What if the user sets a global random seed? (I assume that would affect `cirq.testing.random_special_unitary`, but I don't know.)

supermarq-benchmarks/supermarq/qcvv/su2_test.py
dowusu-antwi, line 1: image

Is this tested? Tbh I'm not fully sure what this means -- it sounds like showing that the number of single-qubit gates is always the same no matter the depth? But each `_component` contains more Haar random single-qubit gates...so it must be referring to something else.

supermarq-benchmarks/supermarq/qcvv/su2.py
cdbf1: In this case the 4/3 actually just comes from rescaling rather than the depolarizing error fun - I'll add a comment. 
On checking whether the argument is >0, we currently don't check this for other routines and just rely on the log function to catch any value errors. In principal we could have `4 / 3 * (self.data["00"] - 1 / 4)<0` and still be able to get a valid fit using a different fitting method. Longer term I would actually like to add a base method for fitting exponentials which works out the best approach to use automatically.

supermarq-benchmarks/supermarq/qcvv/su2.py
cdbf1: I would like to address this for all the QCVV routine using a standardized exponential-fitting tool (see comment below).

supermarq-benchmarks/supermarq/qcvv/su2_test.py
cdbf1: Yes, this is enforced in the `_build_circuits()` method with the lines:
```
circuit = cirq.Circuit(
                *[self._component(include_two_qubit_gate=True) for _ in range(depth)],
                *[self._component(include_two_qubit_gate=False) for _ in range(max_depth - depth)],
            )
```

supermarq-benchmarks/supermarq/qcvv/su2.py
cdbf1: I should have notices this before but the `single_qubit_matrix_to_phxz` returns None if the input is close to the identity so instead of the loop we just return an identity gate.

PR #1018, opened 2024-08-08
Use Python 3.12 base for non-pytest/min coverage checks

bharat-thotakura:
This PR updates the CI checks to use a Python 3.12 base instead of Python 3.8 (which reaches EOL in Oct 2024). This will also allow the pipeline to run against the latest versions of the `css` and `qss` dependencies as some newer releases have already dropped support for Python 3.8/3.9 for example.


bharat-thotakura:
Unrelated to this PR specifically, but I think we should enforce that the `superstaq / Pytest and Coverage check (3.12) (pull_request)` check is `Required` like the other coverage checks

supermarq-benchmarks/supermarq/plotting.py
```diff
@@ -205,8 +208,11 @@ def plot_benchmark(
     theta = radar_factory(num_spokes)
 
     _, ax = plt.subplots(dpi=150, subplot_kw=dict(projection="radar"))
+    assert isinstance(ax, RadarAxesMeta)
 
     title, labels, case_data = data
```
richrines1, line 213: definitely out of scope of this pr, but these should really be three different inputs..

richrines1 (approved):
lgtm!

richrines1:
> Unrelated to this PR specifically, but I think we should enforce that the `superstaq / Pytest and Coverage check (3.12) (pull_request)` check as `Required` like the other coverage checks

done.

supermarq-benchmarks/supermarq/plotting.py
bharat-thotakura: Went made this change in https://github.com/Infleqtion/client-superstaq/pull/1024 since it seemed like a quick minor fix

PR #1021, opened 2024-08-08
disable warning for compiler outputs missing qtrl pulse sequencer

richrines1:
aqt_compile is now intended to work with or without qtrl


bharat-thotakura (approved):
LGTM

PR #1038, opened 2024-08-26
update dd-strategies notebook

richrines1:
updates the ibmq-dd-strategies notebook to (1) present our newer dd strategies, (2) target ibm_brisbane instead of perth (which has been retired), and (3) use `qiskit-ibm-runtime`'s local service instead of `qiskit-ibm-provider` (which has been deprecated). This final change means we no longer require an ibm token for our notebook test

also updates qiskit-superstaq's example-requirements to require `qiskit-ibm-runtime` instead of `qiskit-ibm-provider`. the latter is still used in the ibmq-dd notebook, however this is currently excluded from the notebook check and also relies on outdated devices and strategies and so needs to be updated anyway


qiskit-superstaq/example-requirements.txt
```diff
@@ -1,4 +1,4 @@
 matplotlib>=3.7.0
 notebook>=6.5.5
 pylatexenc>=2.0
-qiskit-ibm-provider>=0.11.0
+qiskit-ibm-runtime>=0.27.0
```
bharat-thotakura, line 4: I think we may need to update the other [DD notebook](https://github.com/Infleqtion/client-superstaq/blob/main/docs/source/optimizations/ibm/ibmq_dd.ipynb) as well before we drop `qiskit-ibm-provider`. 

(An aside, but I feel like this notebook should also show up on the readthedocs)

richrines1:
> This is mainly a style preferential comment so the current default works as well and this can be ignored, but it might make it everso less cramped if we added a cell above that was:

oh that's nice, done :)

qiskit-superstaq/example-requirements.txt
richrines1: > I think we may need to update the other [DD notebook](https://github.com/Infleqtion/client-superstaq/blob/main/docs/source/optimizations/ibm/ibmq_dd.ipynb) `try..except` install

done. (though that notebook also really needs to be updated - as is it won't run even with the correct installs)

> (An aside, but I feel like this notebook should also show up on the readthedocs)

good idea, done.

richrines1:
@bharat-thotakura thanks for the review! i ended up changing the notebook a little further - it occurred to me that there was no reason to include qiskit's transpilation at all, when the purpose of the notebook is just to compare superstaq's different strategies (the other dd notebook is supposed to be the one comparing to qiskit proper)

qiskit-superstaq/example-requirements.txt
bharat-thotakura: >  as is it won't run even with the correct installs

That's a valid point lol, I guess somewhat for the sake of principle and working towards the eventual update. Will make an issue as a reminder

bharat-thotakura (approved):
LGTM

PR #1031, opened 2024-08-16
Move supermarq examples etc. to the Docs folder and include in documentation.

cdbf1:
Move supermarq examples etc. to the Docs folder and include in documentation.

We may want to modify the descriptive .rst files so open to review on this.

Closes #1028

PR #1039, opened 2024-08-26
clean up dependencies and ci

richrines1:
* splits `[examples]` and `[dev]` dependencies for css and qss (currently `[dev]` is a strict superset of `[examples]`, however most of those dependencies aren't necessary outside of the notebook checks
* fixes a conflict between the minimum allowed `pytest` and `pytest-xdist` versions
* cleans up what gets installed for each check in `ci.yml`
* forces the ci cache to be reset when ci.yml itself is changed


.github/workflows/ci.yml
```diff
@@ -229,10 +229,10 @@ jobs:
         include:
           - type: "qiskit-superstaq"
             pattern: "*_qss.ipynb qiskit-superstaq"
-            dependencies: "-e ./checks-superstaq -e ./general-superstaq -e ./qiskit-superstaq[examples] -e ./cirq-superstaq -e ./supermarq-benchmarks"
+            dependencies: "-e ./checks-superstaq -e ./general-superstaq -e ./qiskit-superstaq[examples] -e ./cirq-superstaq"
           - type: "cirq-superstaq"
             pattern: "*_css.ipynb cirq-superstaq"
-            dependencies: "-e ./checks-superstaq -e ./general-superstaq -e ./qiskit-superstaq -e ./cirq-superstaq -e ./supermarq-benchmarks"
+            dependencies: "-e ./checks-superstaq -e ./general-superstaq -e ./qiskit-superstaq -e ./cirq-superstaq[examples]"
```
bharat-thotakura, line 235: Is the idea to introduce a seperate check with supermarq in light of new (tangentially, see [discussion](https://github.com/Infleqtion/server-superstaq/discussions/3352#discussioncomment-10463631)) QCVV notebooks

.github/workflows/ci.yml
```diff
@@ -131,13 +131,13 @@ jobs:
...
+            dependencies: "-e ./general-superstaq[dev]"
           - type: "qiskit-superstaq"
-            dependencies: "./general-superstaq ./qiskit-superstaq[dev]"
+            dependencies: "-e ./general-superstaq -e ./qiskit-superstaq[dev]"
           - type: "cirq-superstaq"
-            dependencies: "./general-superstaq ./qiskit-superstaq ./cirq-superstaq[dev]"
+            dependencies: "-e ./general-superstaq -e ./qiskit-superstaq -e ./cirq-superstaq[dev]"
           - type: "supermarq-benchmarks"
-            dependencies: "./general-superstaq ./qiskit-superstaq ./cirq-superstaq ./supermarq-benchmarks[dev]"
+            dependencies: "-e ./general-superstaq -e ./qiskit-superstaq -e ./cirq-superstaq -e ./supermarq-benchmarks[dev]"
```
bharat-thotakura, line 140: Just for understanding, but I recall we had to [get rid of](https://github.com/Infleqtion/client-superstaq/pull/940/files) the editable installs. What was the key issue previously?

.github/workflows/ci.yml
bharat-thotakura: (outside the scope of this PR, but while on the topic of notebook checks, is https://github.com/Infleqtion/client-superstaq/issues/752 now stale?)

richrines1: yeah was gonna ask you about this - right now the supermarq notebooks (and dd notebook) are excluded from tests: https://github.com/Infleqtion/client-superstaq/blob/8f68a759b743c159dadfa93c0f0613bcc1b83304/checks/pytest_.py#L10-L11

the dd notebook definitely should be excluded (it submits to hardware, and needs a rewrite anyway), but i wasn't sure if the other exclusions were intentional/necessary?

if we add them back in, i do think i'd prefer to make it a third "supermarq-benchmarks" type here though, instead of including them in the css/qss ones (both to make them finish faster, and so these double as validators of the example-requirements files)

.github/workflows/ci.yml
richrines1: hmm not sure, but tbh not sure why i made this change either so reverted

.github/workflows/ci.yml
bharat-thotakura: I'm not entirely sure what the rationale was for `args += ["--exclude", "docs/source/apps/*"]` (seems like added in https://github.com/Infleqtion/client-superstaq/pull/428), but my question was originally prompted from noticing that, for example, the `cirq-superstaq` notebook check doesn't include `supermarq-benchmarks/examples/Supermarq_HPCA_Tutorial_css.ipynb` -- however, I'm realizing now this has more to do with https://github.com/Infleqtion/client-superstaq/pull/1031 and the aforementioned exclusion haha. 

I'm on the side that we probably should create a separate 'supermarq-benchmarks` notebook check if we plan to add new notebooks and the checks can pinpoint incompatibility with new changes (if they don't take _too_ long). However, happy to leave this to a separate PR once those new notebooks are consolidated and we can check runtime or something.

.github/workflows/ci.yml
bharat-thotakura: My best guess if I recall correctly was that this breaks during client version bump because the new version isn't deployed yet or something

.github/workflows/ci.yml
richrines1: ok, added a supermarq notebook check to the matrix, currently just covering the notebooks that seemed to be excluded accidentally

PR #1040, opened 2024-08-27
Add Colab and Binder badges to Supermarq notebooks

bharat-thotakura:


richrines1:
do we have to add the `pip install ..` lines to these to make them runable?

bharat-thotakura:
> do we have to add the `pip install ..` lines to these to make them runable?

Yeah, good catch. Just added the `try..except` for `supermarq` install (in https://github.com/Infleqtion/client-superstaq/pull/1040/commits/ceb41a869c3b3c547c3aac0bbf13ec6afcd97d82) if that's what you're referring to

PR #1043, opened 2024-08-28
Adding `eeroq_compile` endpoint to `cirq_superstaq.Service`

pgoiporia:


pgoiporia:
Turns out we don't need this

PR #1048, opened 2024-08-29
Support sub-job status checking & cancelling in `qss`

bharat-thotakura:
For some `qss` parity with https://github.com/Infleqtion/client-superstaq/pull/1034

### DIFF
--- a/qiskit-superstaq/qiskit_superstaq/superstaq_job.py
+++ b/qiskit-superstaq/qiskit_superstaq/superstaq_job.py
@@ -168,8 +168,7 @@ class SuperstaqJob(qiskit.providers.JobV1):
         raises an exception if it is.
 
         Raises:
-            SuperstaqUnsuccessfulJobException: If the job has been cancelled or has
-                failed.
+            SuperstaqUnsuccessfulJobException: If the job has been cancelled or has failed.
             SuperstaqServerException: If unable to get the status of the job from the API.
         """
         if self._overall_status in ("Cancelled", "Failed"):
--- a/qiskit-superstaq/qiskit_superstaq/superstaq_job.py
+++ b/qiskit-superstaq/qiskit_superstaq/superstaq_job.py
@@ -169,7 +169,7 @@ class SuperstaqJob(qiskit.providers.JobV1):
 
         Raises:
             SuperstaqUnsuccessfulJobException: If the job has been cancelled or has
-        failed.
+                failed.
             SuperstaqServerException: If unable to get the status of the job from the API.
         """
         if self._overall_status in ("Cancelled", "Failed"):
@@ -177,10 +177,11 @@ class SuperstaqJob(qiskit.providers.JobV1):
                 self._job_id, self._overall_status
             )
 
-    def cancel(self, **kwargs: object) -> None:
+    def cancel(self, index: int | None = None, **kwargs: object) -> None:
         """Cancel the current job if it is not in a terminal state.
 
         Args:
+            index: The index of the specific sub-job.
             kwargs: Extra options needed to fetch jobs.
 
         Raises:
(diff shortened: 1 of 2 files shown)


qiskit-superstaq/qiskit_superstaq/superstaq_job.py
```diff
@@ -169,33 +169,41 @@ def _check_if_stopped(self) -> None:
 
         Raises:
             SuperstaqUnsuccessfulJobException: If the job has been cancelled or has
-        failed.
+                failed.
```
richrines1, line 172: nit: it looks like this would fit on one line

qiskit-superstaq/qiskit_superstaq/superstaq_job.py
```diff
@@ -206,7 +214,8 @@ def _refresh_job(self) -> None:
             result = self._backend._provider._client.fetch_jobs(jobs_to_fetch)
             self._job_info.update(result)
 
-        self._update_status_queue_info()
+        if index is None:
+            self._update_status_queue_info()
```
richrines1, line 218: i think we want to do this for any index: e.g. maybe we're checking the last incomplete job in the batch, so if it's done we should update the overall status too
```suggestion
        self._update_status_queue_info()
```
however we'll have to update the logic in `_update_status_queue_info` to handle jobs we haven't retrieved yet (see next suggestion)

qiskit-superstaq/qiskit_superstaq/superstaq_job.py
```diff
@@ -219,7 +228,6 @@ def _update_status_queue_info(self) -> None:
         """
 
         job_id_list = self._job_id.split(",")  # separate aggregated job ids
-
         status_occurrence = {self._job_info[job_id]["status"] for job_id in job_id_list}
```
richrines1, line 231: (see previous)
```suggestion
        status_occurrence = {
            self._job_info[job_id].get("status", "Submitted") for job_id in job_id_list
        }
```

qiskit-superstaq/qiskit_superstaq/superstaq_job.py
```diff
@@ -354,12 +365,17 @@ def status(self) -> qiskit.providers.jobstatus.JobStatus:
...
 
-        self._refresh_job()
-        status = self._overall_status
+            self._refresh_job()
+            status = self._overall_status
+            return status_match.get(status)
 
+        id_to_check = self._job_id.split(",")[index]
+        self._refresh_job(index)
+        status = self._job_info[id_to_check]["status"]
```
richrines1, line 378: i think this method can be simplified a bit knowing that `_refresh_job` already skips jobs in a terminal state, e.g.:
```python
self._refresh_job(index)
if index is None:
    status = self._overall_status
else:
    id_to_check = self._job_id.split(",")[index]
    status = self._job_info[id_to_check]["status"]
return status_match.get(status)

File: qiskit-superstaq/qiskit_superstaq/superstaq_job.py (hunk 1 above)
Comment (bharat-thotakura) [reply qiskit-superstaq/qiskit_superstaq/superstaq_job.py]: You're right, fixed!

File: qiskit-superstaq/qiskit_superstaq/superstaq_job.py (hunk 4 above)
Comment (bharat-thotakura) [reply qiskit-superstaq/qiskit_superstaq/superstaq_job.py]: Thanks, simplified in [68e5e6e](https://github.com/Infleqtion/client-superstaq/pull/1048/commits/68e5e6ea26dc9316cf042dd59e288cebe45cd9c0). I left: https://github.com/Infleqtion/client-superstaq/blob/68e5e6ea26dc9316cf042dd59e288cebe45cd9c0/qiskit-superstaq/qiskit_superstaq/superstaq_job.py#L368-L370
intact since I thought it was a quicker check than the whole `job_ids` loop in `_refresh_job()` and then a call to `_update_status_queue_info()`

File: qiskit-superstaq/qiskit_superstaq/superstaq_job.py [hunk 5]
```diff
@@ -354,12 +365,16 @@ def status(self) -> qiskit.providers.jobstatus.JobStatus:
...
         }
 
-        if self._overall_status in self.TERMINAL_STATES:
-            return status_match.get(self._overall_status)
-
-        self._refresh_job()
-        status = self._overall_status
+        if index is None:
+            if self._overall_status in self.TERMINAL_STATES:
+                return status_match.get(self._overall_status)
```
richrines1, line 370: nit
```suggestion
        if index is None and self._overall_status in self.TERMINAL_STATES:
            return status_match.get(self._overall_status)
```

richrines1 (approved):
lgtm % (optional) nit!

~we should make the same change in css.job~ edit: oops, [you're way ahead of me](https://github.com/Infleqtion/client-superstaq/pull/1034) :)

Issue #1011, opened 2024-08-02
validate qubit types in cirq-superstaq


richrines1:
### What is happening?

we only support qubit types defined in cirq-core. cirq-rigetti and cirq-pasqal both define custom qubit types, which currently result in internal server errors when submitted to superstaq, e.g.
```python
css.Service().compile(
    cirq.Circuit(cirq.X(cirq_rigetti.AspenQubit(2, 3))),
    target="ss_unconstrained_simulator"
)  # SuperstaqServerException: Internal Server Error (Status code: 500)

css.Service().compile(
    cirq.Circuit(cirq.X(cirq_rigetti.AspenQubit(2, 3))),
    target="ss_unconstrained_simulator"
)  # SuperstaqServerException: Internal Server Error (Status code: 500)
```
we should add a validator to reject these circuits client-side with a helpful message

something like this should be sufficient:
```python
supported_qid_types = (
    cirq.LineQubit, cirq.LineQid, cirq.GridQubit, cirq.GridQid, cirq.NamedQubit, cirq.NamedQid
)
if not all(isinstance(q, supported_qid_types) for q in circuit.all_qubits()):
    raise ...
```

### What are the current alternatives?

* add server-side support for these qubit types
* issue an error server side

PR #1050, opened 2024-08-29
Validate qubit types in `css`

bharat-thotakura:
Fixes https://github.com/Infleqtion/client-superstaq/issues/1011

### DIFF
--- a/cirq-superstaq/cirq_superstaq/validation.py
+++ b/cirq-superstaq/cirq_superstaq/validation.py
@@ -35,7 +35,7 @@ def validate_qubit_types(circuits: cirq.Circuit | Sequence[cirq.Circuit]) -> Non
         )
         raise TypeError(
             f"Input circuit(s) contain unsupported qubit types: {invalid_qubit_types}. "
-            "Valid qubit types are: `cirq.LineQubit`, `cirq.LineQid`, `cirq.Grid>Qubit`, "
+            "Valid qubit types are: `cirq.LineQubit`, `cirq.LineQid`, `cirq.GridQubit`, "
             "`cirq.GridQid`, `cirq.NamedQubit`, and `cirq.NamedQid`."
         )
 
--- a/cirq-superstaq/cirq_superstaq/validation.py
+++ b/cirq-superstaq/cirq_superstaq/validation.py
@@ -39,9 +39,9 @@ def validate_qubit_types(circuits: object) -> None:
 
 
 def validate_cirq_circuits(circuits: object, require_measurements: bool = False) -> None:
-    """Validates that the input is an acceptable `cirq` object for `cirq-superstaq`.
+    """Validates that the input is an acceptable `cirq-core` object for `cirq-superstaq`.
 
-    In particular, this function verfies that `circuits` is either a single `cirq.Circuit`
+    In particular, this function verifies that `circuits` is either a single `cirq.Circuit`
     or a list of `cirq.Circuit` instances. Additionally, also validates that `circuits` only
     contains valid qubit types.
 
--- a/cirq-superstaq/cirq_superstaq/validation.py
+++ b/cirq-superstaq/cirq_superstaq/validation.py
@@ -15,13 +15,13 @@ SUPPORTED_QID_TYPES = (
 
 
 def validate_qubit_types(circuits: cirq.Circuit | Sequence[cirq.Circuit]) -> None:
-    """Verifies that `circuits` consists of valid qubit types only.
+    """Verifies that `circuits` consists of valid (`cirq-core`) qubit types only.
 
     Args:
         circuits: The input circuit(s) to validate.
 
     Raises:
-        TypeError: If an invalid qubit type is found in `circuits`.
+        TypeError: If an unsupported qubit type is found in `circuits`.
     """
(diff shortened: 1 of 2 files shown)


cirq-superstaq/cirq_superstaq/validation.py
```diff
@@ -4,10 +4,46 @@
...
+    cirq.LineQubit,
+    cirq.LineQid,
+    cirq.GridQubit,
+    cirq.GridQid,
+    cirq.NamedQubit,
+    cirq.NamedQid,
+)
+
+
+def validate_qubit_types(circuits: object) -> None:
```
cdbf1, line 17:
```suggestion
def validate_qubit_types(circuits: cirq.Circuit | Sequence[cirq.Circuit]) -> None:
```
I think we should set the type as `circuits: cirq.Circuit | Sequence[cirq.Circuits]` since the function implicitly assumes this (otherwise the `circuit.all_qubits()` call would raise an attribute error)

Given that within the `validate_cirq_circuits()` function below ` validate_qubit_types()` is called after the circuits are validated we should be safe to enforce stricter typing

cirq-superstaq/cirq_superstaq/validation.py
```diff
@@ -4,10 +4,46 @@
...
+    circuits_to_check = [circuits] if isinstance(circuits, cirq.Circuit) else circuits
+    assert isinstance(circuits_to_check, Sequence)
+
+    all_qubits_present: set[cirq.Qid] = set()
+    for circuit in circuits_to_check:
+        all_qubits_present.update(circuit.all_qubits())
+
+    if not all(isinstance(q, SUPPORTED_QID_TYPES) for q in all_qubits_present):
+        raise TypeError(
+            "Input circuit(s) contain unsupported qubit types. Valid qubit types are: "
```
cdbf1, line 35:
```suggestion
            f"Input circuit(s) contain unsupported qubit types: {', '.join(set(type(q) for q in all_qubits_present) - SUPPORTED_QID_TYPES)}. Valid qubit types are: "
```

cirq-superstaq/cirq_superstaq/validation_test.py
```diff
@@ -27,3 +30,23 @@ def test_validate_cirq_circuits() -> None:
 
     with pytest.raises(ValueError, match="Circuit has no measurements to sample"):
         css.validation.validate_cirq_circuits(circuit, require_measurements=True)
```
cdbf1, line 32: Could we add another check somewhere to make sure that the `validate_qubit_type()` function is being called - either by mocking the function or by validating a circuit with a bad qubit type.

cdbf1:
A few minor comments but otherwise looks good :)

cirq-superstaq/cirq_superstaq/validation.py
bharat-thotakura: This crossed my mind but I forgot to come back to it. Adapted your suggestion https://github.com/Infleqtion/client-superstaq/pull/1050/commits/3e04c18ff662916e945a889c844373a6cb349bca!

cirq-superstaq/cirq_superstaq/validation_test.py
bharat-thotakura: I changed the test below to call `validate_qubit_type() explicitly`, but regardless, the error would not have been raised in: https://github.com/Infleqtion/client-superstaq/blob/39506fe68c303abcf914a0aeef6f6a5069edbe66/cirq-superstaq/cirq_superstaq/validation_test.py#L51-L56
originally if `validate_qubit_type()` was not being called on `mock_cirq_circuit`.

Were you imagining something like https://github.com/Infleqtion/client-superstaq/pull/1050/commits/b1cf6220dada2a69f13b5259c1cdecd45ced09a0 for `test_validate_cirq_circuits()`?

cirq-superstaq/cirq_superstaq/validation.py
bharat-thotakura: That is indeed what I started with but for some silly reason stuck with `object` since mypy complained. However, it all checks out now with the current implementation -- so fixed, thanks!

cirq-superstaq/cirq_superstaq/validation.py
```diff
@@ -4,10 +4,46 @@
...
+    """Verifies that `circuits` consists of valid qubit types only.
+
+    Args:
+        circuits: The input circuit(s) to validate.
+
+    Raises:
+        TypeError: If an invalid qubit type is found in `circuits`.
+    """
+    circuits_to_check = [circuits] if isinstance(circuits, cirq.Circuit) else circuits
+    assert isinstance(circuits_to_check, Sequence)
```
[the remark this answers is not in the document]
bharat-thotakura: The general view is to not use `asserts` outside of tests unless you absolutely need to due to the type checker's quirks but this is moot due to https://github.com/Infleqtion/client-superstaq/pull/1050#discussion_r1739012876. Removed!

cdbf1:
Thanks, looks good to me

cdbf1 (approved):

bharat-thotakura:
(minor note: qiskit notebook check not passing because of [IBM maintenance](https://quantum.ibm.com/) -- expected to pass later)

PR #1036, opened 2024-08-23
migrate superstaq_backend to `BackendV2`

richrines1:
`BackendV1` was officially deprecated with qiskit 1.2.0.

(also updates a few other instances of deprecated behavior)

PR #1055, opened 2024-09-04
Add custom repr to `qss.SuperstaqBackend`

bharat-thotakura:
Adds a similar repr returned by  `qiskit.providers.BackendV1` before https://github.com/Infleqtion/client-superstaq/pull/1036


qiskit-superstaq/qiskit_superstaq/superstaq_backend.py
```diff
@@ -55,6 +55,9 @@ def __eq__(self, other: Any) -> bool:
 
         return self._provider == other._provider and self.target_info() == other.target_info()
 
+    def __repr__(self) -> str:
+        return f"{self.__class__.__name__}('{self.name}')"
```
richrines1, line 59: i think prior to #1036 the repr also had angle braces:
```suggestion
        return f"<{self.__class__.__name__}('{self.name}')>"
```
(this is also consistent with what qiskit-ibm-runtime does now, for example `<IBMBackend('ibm_brisbane')>`)

bharat-thotakura: This was completely intentional from me. Purely a style preference of mine, but I thought it looked better without the angle braces (similar to how cirq-superstaq prints `Target` objects without such braces). But I suppose this might be an area to maintain qiskit parity than css style parity.

richrines1: ah yeah fair, i don't love them either. i do think they're maybe a little useful here though, in that otherwise `SuperstaqBackend('...')` looks like evaluable code.. imo if we break with the qiskit style we should make it actually evaluable (which will require updating the provider repr too i think)

bharat-thotakura: >... if we break with the qiskit style we should make it actually evaluable (which will require updating the provider repr too i think)

I'm not too hung up on it, so will leave any further change for a future PR (perhaps for `qiskit-superstaq >= 0.6.0`)

richrines1 (approved):
lgtm!

Issue #1045, opened 2024-08-29
Change return types of `qiskit_superstaq.custom_gates` `__array__` methods


pgoiporia:
```suggestion
    def __array__(self, dtype: type | None = None) -> npt.NDArray[np.complex_]:
```
(this should be fixed for all the existing gates too)

_Originally posted by @richrines1 in https://github.com/Infleqtion/client-superstaq/pull/1044#discussion_r1735395924_

bharat-thotakura:
Seems like a sub-issue of https://github.com/Infleqtion/client-superstaq/issues/94

PR #1062, opened 2024-09-11
Update typing of `__array__` dunder methods

bharat-thotakura:
Fixes https://github.com/Infleqtion/client-superstaq/issues/1045

### DIFF
--- a/qiskit-superstaq/qiskit_superstaq/custom_gates.py
+++ b/qiskit-superstaq/qiskit_superstaq/custom_gates.py
@@ -66,7 +66,7 @@ class AceCR(qiskit.circuit.Gate):
         qc.rzx(-self.params[0] / 2, 0, 1)
         self.definition = qc
 
-    def __array__(self, dtype: type | None = None) -> npt.NDArray[np.complex_]:
+    def __array__(self, dtype: type | None = None) -> npt.NDArray[np.generic]:
         """Returns an array for the AceCR gate."""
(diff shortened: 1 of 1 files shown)


qiskit-superstaq/qiskit_superstaq/custom_gates.py
```diff
@@ -66,7 +66,7 @@ def _define(self) -> None:
         qc.rzx(-self.params[0] / 2, 0, 1)
         self.definition = qc
 
-    def __array__(self, dtype: type | None = None) -> npt.NDArray[np.bool_]:
+    def __array__(self, dtype: type | None = None) -> npt.NDArray[np.complex_]:
```
richrines1, line 69: probably doesn't matter but technically this might be more correct, given that passing `dtype=` can change the array type:
```suggestion
    def __array__(self, dtype: type | None = None) -> npt.NDArray[np.generic]:
```
(there's also probably a more clever way to handle this using a typevar and/or overloading but it's probably too complicated to be worthwhile)

bharat-thotakura: Using `np.generic` seems to be the path of least resistance so I updated everywhere else accordingly. I also updated `dtype` to use the broader union of [objects that can be coerced into a `dtype`](https://numpy.org/doc/stable/reference/typing.html#numpy.typing.DTypeLike)

richrines1 (approved):
lgtm!

PR #1065, opened 2024-09-16
Add functionality for RB

cdbf1:
Adds the option to pass no interleaving gate to the IRB module in which case vanilla RB is performed


dowusu-antwi (approved):
LGTM

Issue #1066, opened 2024-09-17
replace `pulse_sequence.draw()` with timeline drawer in all ibm notebooks


richrines1:
currently we return qiskit.pulse.Sequence objects in addition to pulse gate circuits, however (1) we only use them for these diagrams, and (2) they aren't be supported by the newest ibm devices. it would be good to remove them from our docs so we can start phasing them out entirely

richrines1:
closed by #1067

PR #1067, opened 2024-09-17
Use timeline drawer for IBM optimization notebooks

bharat-thotakura:
Addresses part of https://github.com/Infleqtion/client-superstaq/issues/1066


richrines1 (approved):
lgtm!

PR #1072, opened 2024-09-23
fix qcvv progress bars when using cli

richrines1:
currently some progress bars only work correctly in a notebook environment. this is fixed by importing from `tqdm.auto` instead of `tqdm.notebook`

also moves `patch_tqdm` to qcvv/conftest.py to avoid repetition in each test file


cdbf1 (approved):
I didn't know tqdm.auto was a thing - this is great thanks!

LGTM

PR #1074, opened 2024-09-24
allow qcvv routines to be initialized with a random seed

richrines1:
(for reproducible circuit generation/simulation without side effects)

### DIFF
--- a/supermarq-benchmarks/supermarq/qcvv/xeb.py
+++ b/supermarq-benchmarks/supermarq/qcvv/xeb.py
@@ -289,7 +289,7 @@ class XEB(BenchmarkingExperiment[XEBResults]):
         Returns:
             A dictionary of the probability of each bitstring.
         """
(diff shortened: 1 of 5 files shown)


supermarq-benchmarks/supermarq/qcvv/xeb.py
cdbf1, line 1: On line 292 (github won't let me add a comment there) should we add
```
sim = cirq.Simulator(seed=self._rng)
```

I don' think this will have any material effect as the simulator is used to compute the exact probabilities, but for consistency it might be worth adding it anyway.

cdbf1 (approved):
Minor comment, but you might not want to implement it. Otherwise looks good

supermarq-benchmarks/supermarq/qcvv/xeb.py
richrines1: good call, done.

PR #1075, opened 2024-09-24
Fix uninitialised attributes in QCVV samples

### LINKED ISSUE #1071: QCVV sample repr fails if any component hasn't been generated yet
### What is happening?

`print(experiment.samples)` fails for qcvv experiments if any sample hasn't been both run and compiled, because `repr(sample)` will try to access missing attributes. this is pretty annoying when working in a python shell, which tries to print the repr if i just enter e.g. `experiment.samples[0]`

i think the easy (and imo best?) fix would be to have the missing attributes default to None instead of making them [`field`s with `init=False`](https://github.com/Infleqtion/client-superstaq/blob/ca80aad0c8bc72ddc4756c9d70673aef98944c59/supermarq-benchmarks/supermarq/qcvv/base_experiment.py#L40).

```python
exp = supermarq.qcvv.IRB()
exp.prepare_experiment(2, [2])
print(exp.samples)  # AttributeError: 'Sample' object has no attribute 'probabilities'

exp.run_with_simulator()
print(exp.samples)  # AttributeError: 'Sample' object has no attribute 'compiled_circuit'
```

### What should happen?

i should be able to print the samples, even if they're not ready yet

### Environment

- **client-superstaq version**: 0.5.27
- **Python version**: 3.12

### Any additional context?

_No response_

cdbf1:
Closes #1071

### DIFF
--- a/supermarq-benchmarks/supermarq/qcvv/xeb.py
+++ b/supermarq-benchmarks/supermarq/qcvv/xeb.py
@@ -16,8 +16,9 @@ from __future__ import annotations
 
 import itertools
 import random
+import warnings
 from collections.abc import Iterable, Sequence
-from dataclasses import dataclass, field
+from dataclasses import dataclass
 
 import cirq
 import numpy as np
@@ -35,9 +36,9 @@ from supermarq.qcvv import BenchmarkingExperiment, BenchmarkingResults, Sample
 class XEBSample(Sample):
     """The samples used in XEB experiments."""
 
-    target_probabilities: dict[str, float] = field(init=False)
+    target_probabilities: dict[str, float] | None = None
     """The target probabilities obtained through a noiseless simulator"""
-    sample_probabilities: dict[str, float] = field(init=False)
+    sample_probabilities: dict[str, float] | None = None
     """The sample probabilities obtained from the chosen target"""
 
     def sum_target_probs_square(self) -> float:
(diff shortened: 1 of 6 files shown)


supermarq-benchmarks/supermarq/qcvv/xeb.py
```diff
@@ -247,21 +248,27 @@ def _process_probabilities(
...
 
         for sample in tqdm.notebook.tqdm(samples, desc="Evaluating circuits"):
             sample.target_probabilities = self._simulate_sample(sample)
 
         records = []
         for sample in samples:
             target_probabilities = {
-                f"p({key})": value for key, value in sample.target_probabilities.items()
+                f"p({key})": value
+                for key, value in sample.target_probabilities.items()  # type: ignore[union-attr]
```
richrines1, line 267: is it possible to avoid the type: ignore pragmas? e.g. with
```suggestion
                for key, value in sample.target_probabilities.items() if sample.target_probabilities is not None
```
or by combining this w/ the previous loop

cdbf1: The logic meant that this case was never actually reached, hence the type: ignore. But adding the `if _ is not None` works much nicer!

richrines1 (approved):
lgtm!

PR #1076, opened 2024-09-24
allow instantiation of qcvv experiments without superstaq credentials

richrines1:
currently instantiating a qcvv experiment will fail if a `css.Service` can't be constructed, even if we don't require superstaq functionality

### DIFF
--- a/supermarq-benchmarks/supermarq/qcvv/base_experiment.py
+++ b/supermarq-benchmarks/supermarq/qcvv/base_experiment.py
@@ -189,12 +189,20 @@ class BenchmarkingExperiment(ABC, Generic[ResultsT]):
         self._samples: Sequence[Sample] | None = None
         """The attribute to store the experimental samples in."""
 
-        self._service: css.service.Service = css.service.Service(**kwargs)
+        self._service_kwargs = kwargs
+        self._service: css.Service | None = None
         """The superstaq service for submitting jobs."""
 
     ##############
     # Properties #
     ##############
+    @property
+    def superstaq_service(self) -> css.Service:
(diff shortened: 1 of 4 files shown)


supermarq-benchmarks/supermarq/qcvv/base_experiment.py
```diff
@@ -189,12 +189,20 @@ def __init__(
         self._samples: Sequence[Sample] | None = None
         """The attribute to store the experimental samples in."""
 
-        self._service: css.service.Service = css.service.Service(**kwargs)
+        self._service_kwargs = kwargs
+        self._service: css.Service | None = None
```
cdbf1, line 193: Given the property `superstaq_service` which creates a service when needed, what is the point of having the `_service` attribute which is always initialized to `None`?

If we want to give the user an option to use some prebuilt service, then I suggest we have a `service` kwarg which defaults to None and overrides any other service kwargs that the user provides.

richrines1: the idea was that `.superstaq_service` used this attribute to store the service the first time it's referenced, and then returned it in subsequent references so the service only gets instantiated once. not sure why i didn't just use `@cached_property` though - updated.

(agree we might think about allowing the user to pass in a prebuilt service as well though - was mostly trying to avoid any changes in functionality in this pr)

cdbf1: Ah yeah, thats much nicer!

cdbf1 (approved):
lgtm

Issue #1059, opened 2024-09-10
`get_user_info()` endpoint broken


richrines1:
### What is happening?

two bugs breaking `_SuperstaqClient.get_user_info()`:
1. it points to the wrong url (the endpoint should be `"/user_info"` and not [`"/get_user_info"`](https://github.com/Infleqtion/client-superstaq/blob/b512cf185b62591544e3332b0eef79651f9892c2/general-superstaq/general_superstaq/superstaq_client.py#L267)
2. even after fixing the above, it always raises an exception due to a malformed request

### How can we reproduce the issue?

```python
service = css.Service()
_ = service._client.get_user_info()
```
raises
```
SuperstaqServerException: <!DOCTYPE html>
<html lang=en>
  <meta charset=utf-8>
  <meta name=viewport content="initial-scale=1, minimum-scale=1, width=device-width">
  <title>Error 400 (Bad Request)!!1</title>
  <style>
    *{margin:0;padding:0}html,code{font:15px/22px arial,sans-serif}html{background:#fff;color:#222;padding:15px}body{margin:7% auto 0;max-width:390px;min-height:180px;padding:30px 0 15px}* > body{background:url(//www.google.com/images/errors/robot.png) 100% 5px no-repeat;padding-right:205px}p{margin:11px 0 22px;overflow:hidden}ins{color:#777;text-decoration:none}a img{border:0}@media screen and (max-width:772px){body{background:none;margin-top:0;max-width:none;padding-right:0}}#logo{background:url(//www.google.com/images/branding/googlelogo/1x/googlelogo_color_150x54dp.png) no-repeat;margin-left:-5px}@media only screen and (min-resolution:192dpi){#logo{background:url(//www.google.com/images/branding/googlelogo/2x/googlelogo_color_150x54dp.png) no-repeat 0% 0%/100% 100%;-moz-border-image:url(//www.google.com/images/branding/googlelogo/2x/googlelogo_color_150x54dp.png) 0}}@media only screen and (-webkit-min-device-pixel-ratio:2){#logo{background:url(//www.google.com/images/branding/googlelogo/2x/googlelogo_color_150x54dp.png) no-repeat;-webkit-background-size:100% 100%}}#logo{display:inline-block;height:54px;width:150px}
  </style>
  <a href=//www.google.com/><span id=logo aria-label=Google></span></a>
  <p><b>400.</b> <ins>That’s an error.</ins>
[...]

PR #1079, opened 2024-10-01
Fix/user info

cdbf1:
Hopefully fixes https://github.com/Infleqtion/client-superstaq/issues/1059

I think issue was that something on the server side wasn't happy with using json body for get requests. This PR switches to including get request query in the URL string.

Also updates the route to the correct `/user_info`

### DIFF
--- a/general-superstaq/general_superstaq/superstaq_client.py
+++ b/general-superstaq/general_superstaq/superstaq_client.py
@@ -704,7 +704,8 @@ class _SuperstaqClient:
 
         Args:
             endpoint: The endpoint to perform the GET request on.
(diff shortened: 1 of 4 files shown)


general-superstaq/general_superstaq/superstaq_client.py
vtomole, line 1: Looks like `def get_user_info` doesn't take `id` as input. Can we handle that in this PR?

cdbf1: Good call, just added.

vtomole (approved):
LGTM

Issue #1084, opened 2024-10-03
`supermarq/qcvv/irb_test.py::test_random_two_qubit_clifford` test flakes


bharat-thotakura:
The initial deploy for [`v0.5.28`](https://github.com/Infleqtion/client-superstaq/actions/runs/11167141937) failed twice due to a pytest & coverage check flake: [Attempt #1 run](https://github.com/Infleqtion/client-superstaq/actions/runs/11167141937/job/31042699696), [Attempt #2 run](https://github.com/Infleqtion/client-superstaq/actions/runs/11167141937/attempts/2). See full details below:

<details>
 <summary>
run_checks / Pytest and Coverage check (3.10) failure</summary>
Run checks/coverage_.py
============================= test session starts ==============================
platform linux -- Python 3.10.15, pytest-8.3.3, pluggy-1.5.0
Using --randomly-seed=3333475222
rootdir: /home/runner/work/client-superstaq/client-superstaq
configfile: pyproject.toml
plugins: socket-0.7.0, randomly-3.15.0, xdist-3.6.1, nbmake-1.5.4, cov-5.0.0
collected 561 items

qiskit-superstaq/qiskit_superstaq/superstaq_job_test.py ................ [  2%]
......                                                                   [  3%]
supermarq-benchmarks/supermarq/plotting_test.py ....                     [  4%]
supermarq-benchmarks/supermarq/benchmark_test.py ..                      [  4%]
general-superstaq/general_superstaq/resource_estimate_test.py .          [  5%]
supermarq-benchmarks/supermarq/qcvv/base_experiment_test.py ............ [  7%]
.....................                                                    [ [11](https://github.com/Infleqtion/client-superstaq/actions/runs/11167141937/job/31042699696#step:6:12)%]
supermarq-benchmarks/supermarq/converters_test.py ..                     [ 11%]
qiskit-superstaq/qiskit_superstaq/superstaq_backend_test.py ............ [ 13%]
....                                                                     [ 14%]
qiskit-superstaq/qiskit_superstaq/validation_test.py .                   [ 14%]
qiskit-superstaq/qiskit_superstaq/serialization_test.py ................ [ 17%]
........................................................................ [ 30%]
..................................................                       [ 39%]
supermarq-benchmarks/supermarq/benchmarks/qaoa_fermionic_swap_proxy_test.py . [ 39%]
.                                                                        [ 39%]
[...]

PR #1085, opened 2024-10-04
Fix the flakey two qubit IRB test

cdbf1:
Fixes https://github.com/Infleqtion/client-superstaq/issues/1084

Cause: The numpy random number generator didn't seem to be patched properly.

Solution: fix the patch and also path the single qubit Clifford generator so that it outputs a fixed value (not strictly necessary but leads to a more predictable test)


supermarq-benchmarks/supermarq/qcvv/irb_test.py
```diff
@@ -82,9 +81,14 @@ def test_irb_random_clifford() -> None:
 
 
 def test_random_two_qubit_clifford() -> None:
-    with patch("numpy.random.default_rng", side_effect=np.random.default_rng) as rng:
-        rng.return_value.integers.side_effect = range(20)
-        exp = IRB()
+    exp = IRB()
+    # Mock the random number generator
+    exp._rng = (mock_rng := MagicMock())
+    mock_rng.integers.side_effect = range(20)
```
richrines1, line 87: can we just find a random seed that works instead of mocking anything?

supermarq-benchmarks/supermarq/qcvv/irb_test.py
```diff
@@ -82,9 +81,14 @@ def test_irb_random_clifford() -> None:
 
 
 def test_random_two_qubit_clifford() -> None:
```
vtomole, line 83: There's [too many mocks](https://github.com/Infleqtion/server-superstaq/issues/1244) going on here. Let's delete this test. There are quality end-to-end tests that are getting us the 100% coverage we need.

vtomole (approved):
LGTM

PR #1091, opened 2024-10-11
factor out shared code in qaoa proxies

richrines1:
most of the code in QAOAFermionicSwapProxy is identical to that of QAOAVanillaProxy, so instead of repeating it we can inherit


bharat-thotakura (approved):
LGTM

PR #1096, opened 2024-10-29
Fix spurious colon

cdbf1:
Fixes the spurious colons generated in the API summary in the docs. Achieved by hiding the `colon` class in our `.css`.

Supercedes https://github.com/Infleqtion/client-superstaq/pull/944


vtomole (approved):
LGTM

Issue #727, opened 2023-09-06
Need a backend to retrieve an old job


stephanielee9:
### What is happening?

Currently you need to define a backend to retrieve an old job (though the backend does not need to match the original backend the circuit was submitted on - I originally submitted on `ibmq_qasm_simulator`)

### How can we reproduce the issue?

<img width="745" alt="image" src="https://github.com/Infleqtion/client-superstaq/assets/72408904/e7f18757-af80-442b-83aa-d5fc70aac2d5">

### What should happen?

One way could be to retrieve the old job through provider instead of backend

### Environment

- **Superstaq version**: 
- **general-superstaq version**: 
- **cirq-superstaq version**: 
- **qiskit-superstaq version**: 0.4.14
- **Operating system**: MacOS
- **Python version**: 
- **Any other relevant dependencies**:

vtomole:
>  retrieve the old job through provider instead of backend

Sounds good to me. Long term I think we need to phase out `backend`. This is justified given that qiskit is not 1.0 so we don't have to adhere to qiskit's UX.

natibek:
@richrines1 are you still working on this? If not, I can help out.

natibek:
@richrines1 would this be solved by the new interface?

richrines1:
@natibek sorry missed this! honestly forgot about this issue, so definitely not working on it :) will assign you (i think we'll need the same thing regardless of any interface updates)

bharat-thotakura:
Closing as addressed by #1099

PR #1099, opened 2024-10-29
Retrieve old jobs

natibek:
Addressing #727. For `qiskit_superstaq`, this enables users to retrieve a job without specifying a backend using the `qss.SuperstaqProvider` object.


vtomole (commented):
Did you get rid of the other ways to retrieve old jobs in this PR?

richrines1:
> Did you get rid of the other ways to retrieve old jobs in this PR?

are we sure we want to drop backend.retrieve_jobs? this is technically a breaking change (though not sure how much use it gets so maybe doesn't matter)

qiskit-superstaq/qiskit_superstaq/superstaq_provider.py
```diff
@@ -164,6 +164,22 @@ def backends(
...
+            job_id: The UUID of the job. Jobs are assigned these numbers by the server during the
+            creation of the job.
+
+        Returns:
+            A `qss.SuperstaqJob` which can be queried for status or results.
+
+        Raises:
+            ~gss.SuperstaqServerException: If there was an error accessing the API.
+        """
+        job = self._client.fetch_jobs([job_id])
```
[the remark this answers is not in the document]
natibek: We opted to fetch one job to match the implementation in `cirq_superstaq`. Should that also be able to retrieve multi-circuit jobs?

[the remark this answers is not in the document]
richrines1: `css.Service.get_job()` accepts comma-separated job ids, e.g. this works for me:
```python
service = css.Service()
job = service.create_job(
    [
        cirq.Circuit(cirq.measure(cirq.q(0))),
        cirq.Circuit(cirq.X(cirq.q(0)), cirq.measure(cirq.q(0)))
    ],
    target="ss_unconstrained_simulator",
    method="dry-run",
)
print(job.job_id())  # will have a comma, e.g. "bc58eb54-5787-4328-bd45-a53f0cd609c,66f63a82-822d-43f0-8c1d-dce1923ec67"

new_job = service.get_job(job.job_id())
print(new_job.counts(0))  # {"0": 1000}
print(new_job.counts(1))  # {"1": 1000}
```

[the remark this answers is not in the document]
natibek: My bad! I looked at the docstring for `get_job()` and it seemed like it only took a single job id.


```python
  def get_job(self, job_id: str) -> css.job.Job:
        """Gets a job that has been created on the Superstaq API.

        Args:
            job_id: The UUID of the job. Jobs are assigned these numbers by the server during the
            creation of the job.

        Returns:
            A `css.Job` which can be queried for status or results.

        Raises:
            ~gss.SuperstaqServerException: If there was an error accessing the API.
        """
        return css.job.Job(client=self._client, job_id=job_id)

```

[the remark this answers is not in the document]
richrines1: ah yeah that's confusing lol. i think the idea was that a job id with commas is still just a single "job id", but for a multi-circuit job?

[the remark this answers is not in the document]
natibek: I think that makes sense.

natibek:
> > Did you get rid of the other ways to retrieve old jobs in this PR?
> 
> are we sure we want to drop backend.retrieve_jobs? this is technically a breaking change (though not sure how much use it gets so maybe doesn't matter)

I can try looking for instances of its use. The only one I found through test failures was in a notebook. @vtomole what do you think?

vtomole:
From dev meeting: Add deprecation warning for usage and remove later.

qiskit-superstaq/qiskit_superstaq/superstaq_provider.py
```diff
@@ -164,6 +164,30 @@ def backends(
             superstaq_backends.append(self.get_backend(backend.target))
         return superstaq_backends
 
+    def retrieve_job(self, job_id: str) -> qss.SuperstaqJob:
```
vtomole, line 167: Change to `get_job` for `cirq-superstaq` parity

qiskit-superstaq/qiskit_superstaq/superstaq_backend.py
```diff
@@ -172,17 +172,21 @@ def run(
...
             job_id: The UUID of the job. Jobs are assigned these numbers by the server during the
             creation of the job.
-
         Returns:
             A `qss.SuperstaqJob` which can be queried for status or results.
-
         Raises:
             ~gss.SuperstaqServerException: If there was an error accessing the API.
         """
+        warnings.warn(
```
vtomole, line 183: Use Cirq's deprecation infrastructure. [Here's an example](https://github.com/quantumlib/Cirq/blob/d1b0430d27faa75efe933f8dc4bfe1a08b8fed09/cirq-core/cirq/devices/noise_utils.py#L99)

natibek: Is it worth having our own deprecation warning decorator in `gss`?

vtomole: It is not.

natibek: What version should we say is the deadline for deprecation?

vtomole: Let's do `0.5.40`. I made it up.

natibek: Double checking. This is the warning body we will have 

`f'{qualname} was used but is deprecated.\n'
                f'It will be removed in cirq {deadline}.\n'
              f'{fix}\n'`. 
             Given that the message will state that the function will be deprecated for a certain cirq version, should we use it for `qss`?

vtomole: Ah I see. What we have now is fine then.

vtomole (approved):
LGTM

PR #1102, opened 2024-11-14
Refresh some notebooks with newer run

bharat-thotakura:
Notable updates are:

1. Refreshes listed targets from `get_targets()`
2. Refreshes Sqale notebooks (e.g., use "mpl" for `qiskit` circuit visualization)
3. Updates IBM hardware targets used in DFE notebooks (& updates identity gate call in `"dfe_qss"`)


natibek (commented):
Should there be a section on Multiple Circuit Submission for `sqorpius_compile_css.ipynb` and `sqorpius_compile_qss.ipynb`?

bharat-thotakura:
> Should there be a section on Multiple Circuit Submission for `sqorpius_compile_css.ipynb` and `sqorpius_compile_qss.ipynb`?

Yeah, that's a fair observation. Given that we already highlight single- and multi-circuit compilation, I think it might be redundant to mention it again separately. Instead, I've moved around the cell order and show a job submission with both input types. 

I think further overhauls might be outside the scope of this PR's original intention (moreover, we plan to enhance the documentation more in the upcoming months so maybe we can think about additional standardization then), but I have tried to clean up the markdown cells a bit more for now

natibek (approved):
LGTM. I am getting similar test failures on another PR. I don't understand why though. I will try to figure it out.

bharat-thotakura:
Not immediately obvious what the root cause of the CI test failures, but I think the issue stems from the recently released qiskit==1.3.0 (its just strange that it only affects Python 3.9 coverage check and the MacOS check)

PR #1106, opened 2024-11-25
Initial adaptations for Server API version 0.3.0

cdbf1:
In order to prepare to transition to v0.3.0 of the Superstaq API, this PR adds a temporary `_SuperstaqClient` object (named `_SuperstaqClient_v0_3_0` which adheres to the new API and a `css.job._Job` object to work with it (an equivelant update for qiskit-superstaq will follow). This also involves adding a `gss._models` module containing the send/receive data structures that are needed for the new API. 

This PR does not affect the current operation of any existing parts of client superstaq. A warning has also been added to indicate to users that using `v0.3.0` is not stable and that they should continue using `v.0.2.0` for now.


vtomole:
From Dev meeting: Move models to a separate PR to be merged in so server can import.

cdbf1:
Updated with testing for the `general-superstaq`. Still requires testing for `qiskit-superstaq`

PR #1107, opened 2024-11-26
K body noise reconstruction

cdbf1:
Add K-Body noise reconstruction to the QCVV framework


singular-value:
@cdbf1 can you remind me what the status of this PR is? Were we waiting some feedback from end user?

cdbf1:
> @cdbf1 can you remind me what the status of this PR is? Were we waiting some feedback from end user?

K-body noise reconstruction has fallen off the radar a little bit due to other priorities. Viet has been pushing forward the Cycle-benchmarking implementation when he has the time and once that is done we can re-visit k-body noise reconstruction (since it uses cycle benchmarking as a subroutine)

singular-value:
Got it, thanks.
________________________________

PR #1111, opened 2024-11-29
Implementing `get_my_targets()`

natibek:
Clientside changes for the new `get_my_targets()` endpoint that returns only `ss_unconstrained_simulator` if it is a FREE_TRIAL user otherwise filters targets with `supports_submit= True`, `available= True` and `retired= False`.


vtomole (commented):
From dev meeting: This makes sense, but no need to add a server-side endpoint for it. Implement logic in server.

vtomole:
@natibek  Test with latest merge to server and let us know if it still works.

natibek:
I tested it again with a `FREE_TRIAL` user and a `SUPER_TECH` user. It is working as expected for both.

vtomole:
I don't see how it's working from reading the implementation though. It looks almost similar to `def get_targets` except the type definition of `json_dict` is different.

natibek:
The 'accessible' filter is what distinguishes targets as usable or not. The 'get_my_targets()' endpoint calls the server side endpoint with only this filter. The server side endpoint then returns only the targets that are accessible by the user given their role.

vtomole (approved):
LGTM!

PR #1121, opened 2024-12-11
Fixing headers in EeroQ demo notebooks

pgoiporia:
I accidentally set the subheadings to be headings, so the readthedocs contents list is inaccurate.
image


bharat-thotakura (approved):
LGTM

PR #1123, opened 2024-12-11
Add models for new server I/O

cdbf1:
Adds data models that will be needed for v0.3.0 of the superstaq server

### DIFF
--- a/general-superstaq/general_superstaq/_models.py
+++ b/general-superstaq/general_superstaq/_models.py
@@ -1,0 +1,291 @@
+"""Data models used when communicating with the superstaq server."""
+
+# pragma: no cover
+from __future__ import annotations
+
+import datetime
+import uuid
+from enum import Enum
+from typing import Any
+
+import pydantic
+
+
+class JobType(str, Enum):
+    """The different types of jobs that can be submitted through Superstaq."""
+
+    DEVICE_SUBMISSION = "device_submission"
+    """A job that involves submitting circuits to live quantum hardware (device)."""
+    DEVICE_SIMULATION = "device_simulation"
+    """A job that involves simulating a circuit as if it was on some live hardware (device).
+    May or may not include noise (depending on options dict)."""
+    SIMULATION = "simulation"
+    """Simulation of a given quantum circuit but agnostic to any specific device."""
+    COMPILE = "compile"
+    """A job that only involves compiling a circuit for some given device."""
+    CONVERT = "convert"
+    """A job that only involves converting circuits between e.g. qiskit and cirq"""
+
+
+class SimMethod(str, Enum):
+    """The different simulation methods that are available."""
+
+    STATEVECTOR = "statevector"
+    """Simulation calculates the wavefunction of the state at the end of the circuit."""
+    NOISE_SIM = "noise-sim"
+    """The simulation applies noise. If used with `device_simulation` this applies realistic device
+    noise otherwise the noise needs to be provided by the user (via the options dict)."""
+    SIM = "sim"
+    """Samples the circuits without any noise."""
+
+
+class CircuitType(str, Enum):
+    """The different languages that are recognized by Superstaq."""
+
+    CIRQ = "cirq"
+    QISKIT = "qiskit"
+    QASM_STRS = "qasm_strs"
+
+
+class CircuitStatus(str, Enum):
+    """The possible statuses of a job/circuit."""
+
+    RECEIVED = "received"
+    """The job has been received (and accepted) to the server and is awaiting further action."""
+    AWAITING_COMPILE = "awaiting_compile"
+    """The job is waiting for a worker to compile."""
+    AWAITING_SUBMISSION = "awaiting_submission"
+    """The job is waiting for a worker to submit the circuit to an external device."""
+    AWAITING_SIMULATION = "awaiting_simulation"
+    """The job is waiting for a worker to simulate."""
+    AWAITING_CONVERSION = "awaiting_conversion"
+    """The job is waiting for a worker to convert."""
+    RUNNING = "running"
+    """The job is being run by a worker."""
+    COMPLETED = "completed"
+    """The job is completed."""
+    FAILED = "failed"
+    """The job failed. A reason should be stored in the job."""
+    CANCELLED = "cancelled"
+    """The job was cancelled. A reason should be stored in the job."""
+    UNRECOGNIZED = "unrecognized"
+    """Something has gone wrong! (Treated as terminal)"""
+    PENDING = "pending"
+    """When a job has been submitted to an external provider but that provider has
+    not yet run the job."""
+    DELETED = "deleted"
+    """When a job has been deleted."""
+
+
+TERMINAL_CIRCUIT_STATES = [
+    CircuitStatus.COMPLETED,
+    CircuitStatus.FAILED,
+    CircuitStatus.CANCELLED,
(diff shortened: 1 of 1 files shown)


general-superstaq/general_superstaq/_models.py
```diff
@@ -0,0 +1,291 @@
+"""Data models used when communicating with the superstaq server."""
+
+# pragma: no cover
```
vtomole, line 3: This will need to be removed once client-side starts using these models.

vtomole (dismissed):
LGTM

general-superstaq/general_superstaq/_models.py
vtomole: Remind me why these aren't server side for now? https://github.com/Infleqtion/client-superstaq/pull/1106/files doesn't use them either.

cdbf1: https://github.com/Infleqtion/client-superstaq/pull/1106/files does use the models when handling server responses (it validates that the response is in the expected format). See for instance gss.client._SuperstaqClient_v0_3_0.create_job()`
 
 I'm adding them to client side now so that I can import them server side.

cdbf1: Because for now there are no tests that use any of these models (because they are just data structures)

vtomole (approved):
LGTM

general-superstaq/general_superstaq/_models.py
```diff
@@ -0,0 +1,291 @@
...
+]
+
+UNSUCCESSFUL_CIRCUIT_STATES = [
+    CircuitStatus.CANCELLED,
+    CircuitStatus.FAILED,
+    CircuitStatus.DELETED,
+]
+
+
+class DefaultPydanticModel(pydantic.BaseModel, use_enum_values=True, extra="forbid"):
```
richrines1, line 95: we might want to rethink `extra="forbid"` - it'd be nice to add new fields server-side and not have it break existing clients

general-superstaq/general_superstaq/_models.py
```diff
@@ -0,0 +1,291 @@
...
+    target: str
+    circuits: str
+    circuit_type: CircuitType
+    compiled_circuits: str | None = pydantic.Field(default=None)
+    compiled_circuit_type: CircuitType | None = pydantic.Field(default=None)
+    shots: int = pydantic.Field(default=0, ge=0)
+    dry_run: bool = pydantic.Field(default=False)
+    sim_method: SimMethod | None = pydantic.Field(default=None, validate_default=True)
+    priority: int = pydantic.Field(default=0)
+    options_dict: dict[str, Any] = pydantic.Field(default={})
```
richrines1, line 139: this might be a good time to think about retiring the options dict. it was originally mostly intended as a stopgap so that we could make things flexible as we refined our interfaces, now that we're defining these models we should maybe to start breaking things into their own fields/models (and maybe just use for `extra="allow"` for experimental/wip/non-publicized fields?)

(not sure how we want to organize things though - we could put everything in the top level model, or keep a single `options` field w/ a model for possible arguments, or start separating e.g. compilation options/user credentials/simulation parameters into their own models...

general-superstaq/general_superstaq/_models.py
```diff
@@ -0,0 +1,291 @@
...
+
+    name: str | None = pydantic.Field(None)
+    """New user name."""
+    role: str | None = pydantic.Field(None)
+    """New user role."""
+    balance: float | None = pydantic.Field(None)
+    """New user balance"""
+
+
+class TargetInfoModel(DefaultPydanticModel):
```
richrines1, line 251:
```suggestion
class TargetModel(DefaultPydanticModel):
```
(we use "target_info" for info about the hardware itself, e.g. coupling maps, gate sets, qubits...)

general-superstaq/general_superstaq/_models.py
```diff
@@ -0,0 +1,291 @@
...
+    """Target allows circuit compilation."""
+    available: bool
+    """Target is currently available."""
+    retired: bool
+    """Target is retired."""
+    simulator: bool
+    """Target is simulator"""
+
+
+class GetTargetsFilterModel(DefaultPydanticModel):
```
richrines1, line 270: could this and the above model be combined?

general-superstaq/general_superstaq/_models.py
```diff
@@ -0,0 +1,291 @@
...
+    num_qubits: list[int]
+    shots: list[int]
+    dry_run: bool
+    submission_timestamp: datetime.datetime
+    last_updated_timestamp: list[datetime.datetime | None]
+    initial_logical_to_physicals: list[str | None]
+    final_logical_to_physicals: list[str | None]
+
+
+class NewJob(DefaultPydanticModel):
```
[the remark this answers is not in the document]
cdbf1: This is intended to replace the "verbatim=True" option. On the server, if the `compiled_circuits` is provided then the server skips the compiling step

general-superstaq/general_superstaq/_models.py
cdbf1: This is top of my to-do list once the initial server is up and running. As per my comment above about extra="forbid", the client side models should be fully enumerated.

I imagine that the best approach will be to have different models for different types of jobs etc., set up in some hierarchical way this will ensure that everything gets validated automatically by the server. This is part of the job refactor to-do list

For now I suggest we keep the `options_dict` field to maintain compatibility with existing Superstaq method.

general-superstaq/general_superstaq/_models.py
cdbf1: Although the field names are the same they represent different things so I feel they should remain as separate models. Also on server side some logic is added to the to the FilterModel

general-superstaq/general_superstaq/_models.py
cdbf1: My feeling is that for ease of use, self-documentation and secure coding, any I/O models should be fully enumerated. I agree it's a little less flexible though

general-superstaq/general_superstaq/_models.py
```diff
@@ -0,0 +1,291 @@
...
+    """The data model for submitting new jobs."""
+
+    job_type: JobType
+    target: str
+    circuits: str
+    circuit_type: CircuitType
+    compiled_circuits: str | None = pydantic.Field(default=None)
```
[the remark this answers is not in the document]
cdbf1: I think this is needed server-side to play nicely with fastapi.

general-superstaq/general_superstaq/_models.py
richrines1, line 95: two other options that are nice imo: `validate_assignment=True` and `validate_default=True`

general-superstaq/general_superstaq/_models.py
richrines1: yep totally agree, just wanted to flag it in case it had any implications for how we go about organizing things

general-superstaq/general_superstaq/_models.py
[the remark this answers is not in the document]
richrines1: ah ok. imo it would be nicer to just include the verbatim flag and a single `circuits` field, unless there is some case in which we'd need both

general-superstaq/general_superstaq/_models.py
cdbf1: From dev meeting: Set `extra=ignore` for now.

PR #1127, opened 2024-12-13
support qiskit 1.3.*

richrines1:
(still lots of deprecation warnings to work out though)


qiskit-superstaq/qiskit_superstaq/serialization.py
```diff
@@ -52,6 +52,55 @@
...
+) -> qiskit.circuit.library.MCPhaseGate:
+    """The `ctrl_state` argument was added to `MCPhaseGate` in Qiskit 1.1.0."""
+
+    if qiskit.__version__.split(".")[:2] >= ["1", "1"]:
+        return qiskit.circuit.library.MCPhaseGate(
+            lam, num_ctrl_qubits=num_ctrl_qubits, label=label, ctrl_state=ctrl_state
+        )
+
+    gate = qiskit.circuit.library.MCPhaseGate(lam, num_ctrl_qubits=num_ctrl_qubits, label=label)
+    gate.ctrl_state = ctrl_state
```
bharat-thotakura, line 69: Curious, will overwriting the `ctrl_state` attribute (which seems to default to $2^\text{num-ctrl-qubits} - 1$) if not provided) cause serialization issues in `qiskit>=1.1.*`? Of course, it does seem much better to pass it as an input arg if available

bharat-thotakura (approved):
LGTM

qiskit-superstaq/qiskit_superstaq/serialization.py
richrines1: conveniently no - `ControlledGate.ctrl_state` is implemented as a property with a [setter](https://github.com/Qiskit/qiskit/blob/053ee324eef0b872e21e1d5e8816ecc0e3c4057f/qiskit/circuit/controlledgate.py#L212) implementing the correct parsing logic, so `gate.ctrl_state = None`, `gate.ctrl_state = 2**n- 1`, `gate.ctrl_state = "111..."` all do the same thing

PR #1128, opened 2024-12-13
Minor fix to v0.3.0 I/O models

cdbf1:
Correcting a small typing mistake and fixing naming.


dowusu-antwi (approved):
lgtm!

general-superstaq/general_superstaq/_models.py
```diff
@@ -264,13 +265,13 @@ class UserQuery(DefaultPydanticModel):
     """Model for querying the database to retrieve users. Use of lists implied logical OR. Providing
     multiple fields (e.g. name and email) implies logical AND."""
 
-    name: list[str] | None = pydantic.Field(None)
+    name: Sequence[str] | None = pydantic.Field(None)
```
richrines1, line 268: why Sequence? i thought pydantic should convert/cast the input so generic types shouldn't be necessary

cdbf1: On server side the this class gets subclassed to introduce a checks for enums not exposed to the client side. Therefore the type needs to be covariant - hence Sequence. Strictly speaking not required for all fields but I thought it would be neater to apply to all.

richrines1: ah ok makes sense!

PR #1131, opened 2024-12-19
[testing] use nbval instead of nbmake for notebook testing

richrines1:


pgoiporia:
Any reason we abandoned this? Seems like nbval has 2x the stars that nbmake does. I'm happy to pick this up.

richrines1:
> Any reason we abandoned this? Seems like nbval has 2x the stars that nbmake does. I'm happy to pick this up.

iirc when i stopped nbval was failing due to a missing/invalid `execution_count` in a couple notebook cells. This should be super easy to correct though - it should just require rerunning notebooks or manually editing the notebook json (but instead i think i went down a rabbit hole of auto-standardizing execution counts/notebook metadata lol, which i still think would be nice but definitely isn't necessary for this pr)

PR #1132, opened 2024-12-20
update max sharpe notebook

richrines1:
we exclude this one from pytest but include it in our docs, and currently it won't run correctly (due to changes in yfinance i think?)


docs/source/apps/max_sharpe_ratio_optimization.ipynb
```diff
@@ -81,10 +81,7 @@
     "import numpy as np\n",
```
bharat-thotakura, line 81: <pre class="ql-syntax" spellcheck="false">Line #14.        %pip install --quiet yfinance
</pre><p>&gt;due to changes in yfinance i think</p><p> </p><p>Would you recommend requiring a certain version or higher of <code>yfianance</code> ? Like the current latest version or something </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Infleqtion/client-superstaq/pull/1132/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

bharat-thotakura, line 81: <pre class="ql-syntax" spellcheck="false">Line #11.    df0 = df
</pre><p>It looks the original df = table[0] from the previous cell isn't used again and is purely for illustration (case in point, I was having trouble connecting to Wikipedia at the time of this comment but was able to run the notebook by ignoring the previous cell). </p><p> </p><p>If so, I think we can get rid of this line and not have any assignment in the previous cell? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Infleqtion/client-superstaq/pull/1132/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

bharat-thotakura, line 81: <pre class="ql-syntax" spellcheck="false">Line #12.    print("Creating objective function...")
</pre><p>Minor, optional nit: Can we add one more print statement at the end of the cell to just validate the cell executed fully? Otherwise it may look like it's still 'creating' fwiw</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Infleqtion/client-superstaq/pull/1132/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

bharat-thotakura (approved):
Thanks for looking at this! LGTM % comments/questions 

(was able to run the notebook locally without issue ... except the cell that calls Wikipedia, but I'm having trouble with Wikipedia outside of this notebook anyway so most likely not universal)

docs/source/apps/max_sharpe_ratio_optimization.ipynb
richrines1: yep great point.

ending up moving the specific requirements (including versions) to gss/example-requirements to better align with what we do for css/qss

richrines1: yep, done. found a few other unnecessary assignments/imports as well

richrines1: done.

PR #1134, opened 2024-12-23
Usage & documentation of `accessible` filter & `get_my_targets()`

bharat-thotakura:
Fixes https://github.com/Infleqtion/client-superstaq/issues/1133 but more broadly updates documentation to mention the new `accessible` filter and `get_my_targets()` method.


natibek (approved):
LGTM

Issue #799, opened 2023-09-29
add option to retrieve stored ibm credentials


richrines1:
### What is happening?

i can instantiate a `QiskitRuntimeService()` / `IBMProvider()` without any arguments because my credentials are already saved in a dot file. it would be nice if i could automatically use the same credentials when i instantiate a `qss.SuperstaqProvider` or `css.Service()`, instead of having to pass everything in manually

imo this shouldn't happen by default though - maybe we could either (1) allow the user to pass in a provider object/runtime service instead of individual credentials, (2) add an argument like `use_stored_ibmq_credentials=False`, or (3) some combination of these and a flag allowing the automatic reading of ibm configs that we put in our own config file?

PR #1136, opened 2024-12-30
Add option to retrieve stored ibm credentials #799

natibek:
Adds the option to retrieve stored ibm credentials (#799) by either:

1. setting the `use_stored_ibm_credentials` parameter to `True` or
2. passing an `IBMProvider()` object instance to the `ibmq_provider` parameter. 

First the `use_stored_ibm_credentials` flag is checked then whether an `ibmq_provider` is provided. Finally, if neither are set, checks if `ibmq_token`, `ibmq_instance`, and `ibmq_channel` have been set manually.

### DIFF
--- a/general-superstaq/general_superstaq/superstaq_client.py
+++ b/general-superstaq/general_superstaq/superstaq_client.py
@@ -25,6 +25,7 @@ from collections.abc import Callable, Mapping, Sequence
 from typing import Any, TypeVar
 
 import requests
+from qiskit_ibm_provider import IBMProvider
 
 import general_superstaq as gss
 
@@ -57,6 +58,8 @@ class _SuperstaqClient:
         ibmq_token: str | None = None,
         ibmq_instance: str | None = None,
         ibmq_channel: str | None = None,
+        ibmq_provider: IBMProvider | None = None,
+        use_stored_ibmq_credentials: bool = False,
         **kwargs: Any,
     ) -> None:
         """Creates the SuperstaqClient.
@@ -111,17 +114,29 @@ class _SuperstaqClient:
         }
         self.session = requests.Session()
 
-        if ibmq_channel and ibmq_channel not in ("ibm_quantum", "ibm_cloud"):
-            raise ValueError("ibmq_channel must be either 'ibm_cloud' or 'ibm_quantum'.")
-
         if cq_token:
             kwargs["cq_token"] = cq_token
-        if ibmq_token:
-            kwargs["ibmq_token"] = ibmq_token
-        if ibmq_instance:
-            kwargs["ibmq_instance"] = ibmq_instance
-        if ibmq_channel:
-            kwargs["ibmq_channel"] = ibmq_channel
+
+        if use_stored_ibmq_credentials:
+            ibmq_provider = IBMProvider()
+
+        if ibmq_provider:
+            ibmq_active_account = ibmq_provider.active_account()
+            assert (
+                ibmq_active_account
+            ), "No active account found for provided `IBMProvider` instance."
+            kwargs["ibmq_token"] = ibmq_active_account.get("token")
+            kwargs["ibmq_instance"] = ibmq_active_account.get("instance")
+            kwargs["ibmq_channel"] = ibmq_active_account.get("channel")
+        else:
+            if ibmq_token:
+                kwargs["ibmq_token"] = ibmq_token
+            if ibmq_instance:
+                kwargs["ibmq_instance"] = ibmq_instance
+            if ibmq_channel and ibmq_channel not in ("ibm_quantum", "ibm_cloud"):
+                raise ValueError("ibmq_channel must be either 'ibm_cloud' or 'ibm_quantum'.")
+            elif ibmq_channel:
+                kwargs["ibmq_channel"] = ibmq_channel
 
         self.client_kwargs = kwargs
 
--- a/cirq-superstaq/cirq_superstaq/service.py
+++ b/cirq-superstaq/cirq_superstaq/service.py
@@ -31,6 +31,7 @@ import cirq_superstaq as css
 
 if TYPE_CHECKING:
     from _typeshed import SupportsItems
+    from qiskit_ibm_provider import IBMProvider
 
 
 def _to_matrix_gate(matrix: npt.ArrayLike) -> cirq.MatrixGate:
@@ -138,6 +139,8 @@ class Service(gss.service.Service):
         ibmq_token: str | None = None,
         ibmq_instance: str | None = None,
         ibmq_channel: str | None = None,
+        ibmq_provider: IBMProvider | None = None,
+        use_stored_ibmq_credentials: bool = False,
         **kwargs: object,
     ) -> None:
         """Creates the Service to access Superstaq's API.
@@ -187,6 +190,8 @@ class Service(gss.service.Service):
             ibmq_token=ibmq_token,
             ibmq_instance=ibmq_instance,
             ibmq_channel=ibmq_channel,
+            ibmq_provider=ibmq_provider,
+            use_stored_ibmq_credentials=use_stored_ibmq_credentials,
             **kwargs,
         )
 
--- a/general-superstaq/general_superstaq/superstaq_client_test.py
+++ b/general-superstaq/general_superstaq/superstaq_client_test.py
@@ -215,6 +215,73 @@ def test_superstaq_client_validate_email_error(
         _ = client.create_job({"Hello": "World"}, target="ss_example_qpu")
 
 
+class MockIBMProvider:
+    """MockIBMProvider is a mock class to represent an `IBMProvider`."""
+
+    def __init__(
+        self,
+        token: str | None = "ibmq_token",
+        instance: str = "instance",
+        channel: str = "ibm_quantum",
+    ) -> None:
+        self.token = token
+        self.instance = instance
+        self.channel = channel
+
+    def active_account(self) -> dict[str, str] | None:
+        """Mock of the `IBMProvider.active_account` function.
+
+        Returns:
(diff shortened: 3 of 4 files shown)


general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -111,17 +114,29 @@ def __init__(
...
             kwargs["cq_token"] = cq_token
-        if ibmq_token:
-            kwargs["ibmq_token"] = ibmq_token
-        if ibmq_instance:
-            kwargs["ibmq_instance"] = ibmq_instance
-        if ibmq_channel:
-            kwargs["ibmq_channel"] = ibmq_channel
+
+        if use_stored_ibmq_credentials:
+            ibmq_provider = IBMProvider()
```
richrines1, line 121: for runtime i think this will be something like
```suggestion
            ibmq_account = qiskit_ibm_runtime.accounts.AccountManager.get(
                channel=ibmq_channel, name=ibmq_name
            )
```
qiskit runtime allows multiple accounts to be saved to be saved under different names, so we might want to add an argument to allow users to specify an `ibmq_name` too?

general-superstaq/requirements.txt
```diff
@@ -1,4 +1,5 @@
 numpy>=1.21.0
 pydantic>=1.10.7
+qiskit-ibm-provider==0.11.0
```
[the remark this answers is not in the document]
richrines1: i think in this case i'd just go with `ModuleNotFoundError`

natibek:
@richrines1, 

```bash 
The user requested pydantic==1.10.7
    checks-superstaq 0.5.33 depends on pydantic>=1.10.7
    qiskit-ibm-runtime 0.34.0 depends on pydantic<2.10 and >=2.5.0

```

should I change the `qiskit-ibm-runtime` version? Would removing the requirement for the version let `pip` install a version that does not cause conflicts?

richrines1:
> should I change the `qiskit-ibm-runtime` version? Would removing the requirement for the version let `pip` install a version that does not cause conflicts?

not 100% sure it will work but try `qiskit-ibm-runtime>=0.21.0`

general-superstaq/requirements.txt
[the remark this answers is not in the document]
natibek: @cdbf1 `qiskit-superstaq` already has it as an optional dependency under the `examples` optional dependency. Is it bad design to have the same optional dependency under `qiskit-superstaq[ibm_credentials]` as well?

vtomole:
From dev meeting: Remove runtime requirement and from `gss` and read from the IBM config files ourselves.

vtomole:
@natibek Is this ready for another review?

natibek:
@vtomole Yes, it.

vtomole (approved):
LGTM.

@bharat-thotakura  Can you check to make sure this works on Windows? The checks pass, but just want to make sure.

bharat-thotakura:
@vtomole, unsure about what to check on Windows (see https://github.com/Infleqtion/client-superstaq/issues/873)? I would assume it works in WSL2 like on bare Ubuntu

general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -920,6 +933,53 @@ def __repr__(self) -> str:
...
+        home_dir.joinpath(".qiskit"),
+        data_dir.joinpath("super.tech"),
+        data_dir.joinpath("coldquanta"),
+        home_dir.joinpath(".super.tech"),
+        home_dir.joinpath(".coldquanta"),
+    ]:
+        path = directory.joinpath("qiskit-ibm.json")
+        if path.is_file():
+            config = json.load(open(path))
+            if ibm_name not in config:
```
richrines1, line 963: if the user doesn't pass in a name we should handle it closer to what qiskit does instead of assuming a particular name. i think this means:
1. look for any account they've marked as default (by checking `credentials.get("is_default_account")`)
2. look for accounts named `["default-ibm-cloud", "default-ibm-quantum", "default"]` (in that order i think)
3. look for the last account with the channel "ibm-cloud"
4. return the last account in the config

general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -920,6 +933,53 @@ def __repr__(self) -> str:
...
+        Dictionary containing the ibm token, channel, and instance (if available).
+    """
+    data_dir = pathlib.Path(os.getenv("XDG_DATA_HOME", "~/.local/share")).expanduser()
+    home_dir = pathlib.Path.home()
+    for directory in [
+        home_dir.joinpath(".qiskit"),
+        data_dir.joinpath("super.tech"),
+        data_dir.joinpath("coldquanta"),
+        home_dir.joinpath(".super.tech"),
+        home_dir.joinpath(".coldquanta"),
```
richrines1, line 958: we probably only need to look in the `.qiskit` directory

general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -111,11 +117,18 @@ def __init__(
         }
         self.session = requests.Session()
 
+        if cq_token:
+            kwargs["cq_token"] = cq_token
+
+        if use_stored_ibmq_credentials:
+            config = gss.superstaq_client.read_ibm_credentials(ibmq_name or "default-ibm-quantum")
```
richrines1, line 124: we should not assume a name here - instead if `read_ibm_credentials()` should accept None (see prior comment)
```suggestion
            config = read_ibm_credentials(ibmq_name)
```

vtomole:
@bharat-thotakura Oh! I didn't know that that this doesn't work on Windows. NVM.

general-superstaq/general_superstaq/superstaq_client.py
richrines1: assuming this came from `gss.superstaq_client` - the first one checks the XDG data directory (e.g. `~/.local/share/super.tech`), second one looks for a hidden folder in the home directory (`~/.super.tech`)

general-superstaq/general_superstaq/superstaq_client.py
richrines1: alternatively, if we're not given a name maybe we should just check for `credentials.get("is_default_account")`, and if that doesn't work throw an error to avoid doing something the user might not expect (unless there is exactly one account in the config, in which case we should just return that)

general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -920,6 +933,57 @@ def __repr__(self) -> str:
...
+            if len(config) == 1:
+                ibmq_name = list(config.keys())[0]
+            elif "default-ibm-quantum" not in config:
+                raise ValueError(
+                    "Multiple accounts found but none are marked default with ",
+                    "`default-ibm-quantum`. Please provide the name of the ",
+                    "account to retrieve.",
+                )
+            else:
+                ibmq_name = "default-ibm-quantum"
```
richrines1, line 967: please check for the "is_default_account" item in the configs themselves, instead of this name. e.g.
```suggestion
            elif any(creds.get("is_default_account") for creds in config.values()):
                ibmq_name = next(name for name in config if config[name].get("is_default_account"))
            else:
                raise ValueError(
                    "Multiple accounts found but none are marked default with ",
                    "`default-ibm-quantum`. Please provide the name of the ",
                    "account to retrieve.",
                )
```

general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -920,6 +933,55 @@ def __repr__(self) -> str:
...
+            raise KeyError(
+                f"No account credentials saved under the name {ibmq_name}"
+                f" in the config file found at {path}."
+            )
+
+        credentials = config.get(ibmq_name)
+        if any(key not in credentials for key in ["token", "channel"]):
+            raise KeyError(
+                "`token` and/or `channel` keys missing from credentials for the account",
+                f" under the name {ibmq_name} in the file {path}.",
```
richrines1, line 977: nits: wrap line at 100 and add quotes:
```suggestion
                "`token` and/or `channel` keys missing from credentials for the account under the "
                f"name '{ibmq_name}' in the file '{path}'.",
```

general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -920,6 +933,55 @@ def __repr__(self) -> str:
...
+            else:
+                raise ValueError(
+                    "Multiple accounts found but none are marked as default.",
+                    " Please provide the name of the account to retrieve.",
+                )
+        elif ibmq_name not in config:
+            raise KeyError(
+                f"No account credentials saved under the name {ibmq_name}"
+                f" in the config file found at {path}."
+            )
```
richrines1, line 971:
```suggestion
            raise KeyError(
            raise KeyError(
                f"No account credentials saved under the name '{ibmq_name}' in the config file "
                f"found at '{path}'."
            )
```

general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -920,6 +933,55 @@ def __repr__(self) -> str:
...
+        credentials = config.get(ibmq_name)
+        if any(key not in credentials for key in ["token", "channel"]):
+            raise KeyError(
+                "`token` and/or `channel` keys missing from credentials for the account",
+                f" under the name {ibmq_name} in the file {path}.",
+            )
+
+        return credentials
+
+    raise FileNotFoundError(f"The `qiskit-ibm.json` file was not found in {config_dir}.")
```
richrines1, line 982:
```suggestion
    raise FileNotFoundError(f"The `qiskit-ibm.json` file was not found in '{config_dir}'.")
```

richrines1 (commented):
lg % formatting nits!

richrines1 (commented):
lg % formatting nits!

Issue #761, opened 2023-09-12
Document "advanced" features


stephanielee9:
Related: https://github.com/Infleqtion/client-superstaq/issues/523. We should have a notebook(s) or include this info in current notebooks that is the next step to basics. Things we can include:
- [ ] `dry-run` (explain what it means also)
- [ ] any options we're missing
- [ ] BYO IBM token when we have it
- [ ] explanation on how we use Runtime
- [ ] `verbatim`

PR #1137, opened 2025-01-07
remove redundant examples

richrines1:
the examples/ subdirectories in css and qss have been superseded by the docs/ directory


bharat-thotakura (commented):
We currently don't have a notebook to showcase the `resource_estimate()` method on read-the-docs. Do we want a full notebook for it or is the [API reference](https://superstaq.readthedocs.io/en/latest/autoapi/general_superstaq/resource_estimate/index.html) and docstring sufficient? Alternatively, we could have a dedicated 'advanced/general superstaq features' notebook as a sequel to [basics](https://superstaq.readthedocs.io/en/latest/get_started/basics/basics.html) & [account_info](https://superstaq.readthedocs.io/en/latest/get_started/access_info/access_info.html) that highlights things like `resource_estimate`, and features like in https://github.com/Infleqtion/client-superstaq/issues/761

richrines1:
i think we decided way back that resource_estimate() itself didn't really warrant a demo yet? (the endpoint is pretty under-developed atm, and will probably change substantially before it's useful)

that said i do really like the idea of adding an "advanced features" demo before the device-specific stuff. it also seems like the "basics" notebooks themselves could be extended a bit - currently they only showcase submission, but imo should at least demo compilation as well

bharat-thotakura (approved):
Ah I see, I wasn't aware of that decision. But yeah, I agree with the reasoning given the current state of `resource_estimate`

PR #1140, opened 2025-01-10
Format TOML files with taplo

perlinm:
`taplo fmt <file>` will format TOML files.

Maybe I should add https://pypi.org/project/taplo/ and use it in `checks/format_.py`?  Might not be worth the extra dependency though.

I mostly want this change because I want to format my TOML files in other projects (in particular, `pyproject.toml`), but that causes `checks/configs.py` to fail 🙂


bharat-thotakura (approved):
LGTM

PR #1147, opened 2025-01-20
Add figure saving to QCVV library

cdbf1:
Adds functionality to save figures generated during the `analyze()` step in the QCVV framework


supermarq-benchmarks/supermarq/qcvv/irb.py
```diff
@@ -405,9 +414,19 @@ def average_error_per_clifford_std(self) -> float:
...
+        self,
+        filename: str | None = None,
+    ) -> None:
+        """Plot the exponential decay of the circuit fidelity with cycle depth.
+
+        Args:
+            filename: Optional argument providing a filename to save the plots to. Defaults to None,
+                indicating not to save the plot.
+        """
         super()._plot_results()
```
richrines1, line 427: predates this pr, but this should just be
```suggestion
        self._plot_results()
```
the only reason to use super() would be if `_plot_results()` was overridden in this subclass (and we wanted to use the original version)

(same occurs a few other places as well)

supermarq-benchmarks/supermarq/qcvv/xeb_test.py
```diff
@@ -129,13 +130,13 @@ def test_build_xeb_circuit(xeb_experiment: XEB) -> None:
     }
 
 
-def test_xeb_analyse_results(xeb_experiment: XEB) -> None:
+def test_xeb_analyse_results(tmp_path: pathlib.Path, xeb_experiment: XEB) -> None:
```
richrines1, line 133: oh wow somehow never knew pytest provides `tmp_path`, that's super nice

supermarq-benchmarks/supermarq/qcvv/irb.py
```diff
@@ -351,6 +358,8 @@ def plot_results(self) -> None:
             alpha=0.5,
             color="tab:orange",
         )
+        if filename is not None:
+            plt.savefig(filename)
```
richrines1, line 362: fwiw locally i usually have to use
```suggestion
            plt.savefig(filename, bbox_inches="tight")
```
to prevent weird margins/cutoffs around labels/etc. Not sure if it makes sense to use here though?

richrines1: also does this work?
```suggestion
            figure = plot.get_figure()
            figure.savefig(filename, bbox_inches="tight"))
```
(to avoid relying on the global mpl state)

richrines1 (commented):
lg % a couple of questions (and a nit unrelated to these changes)

supermarq-benchmarks/supermarq/qcvv/irb.py
cdbf1: Yeah not sure how that slipped through

supermarq-benchmarks/supermarq/qcvv/irb.py
cdbf1: > also does this work?
> 
> (to avoid relying on the global mpl state)

Good shout

supermarq-benchmarks/supermarq/qcvv/xeb.py
```diff
@@ -169,15 +177,24 @@ def plot_results(self) -> None:
         ax_2.plot(x, y, color="tab:red", linewidth=2)
         ax_2.fill_between(x, y_m, y_p, alpha=0.2, color="tab:red")
 
+        if filename is not None:
+            name, extension = os.path.splitext(filename)
+            plot_1.savefig(name + "_ray_plot" + extension, bbox_inches="tight")
+            plot_2.savefig(name + "_circuit_fidelity_decay" + extension, bbox_inches="tight")
```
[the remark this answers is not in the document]
cdbf1: Yeah I see your point. To facilitate returning a single figure to the user I've now put both plots on a single figure so they can be saved together with a single name.

supermarq-benchmarks/supermarq/qcvv/irb.py
```diff
@@ -405,9 +426,31 @@ def average_error_per_clifford_std(self) -> float:
...
+    ) -> plt.Figure:
+        """Plot the exponential decay of the circuit fidelity with cycle depth.
+
+        Args:
+            filename: Optional argument providing a filename to save the plots to. Defaults to None,
+                indicating not to save the plot.
+
+        Returns:
+            A single matplotlib figure containing the RB decay plot and the corresponding
+            fit.
```
richrines1, line 441: nit
```suggestion
            A single matplotlib figure containing the RB decay plot and the corresponding fit.
```

supermarq-benchmarks/supermarq/qcvv/xeb.py
```diff
@@ -116,9 +116,20 @@ def _analyze(self) -> None:
...
+    ) -> plt.Figure:
         """Plot the experiment data and the corresponding fits.
 
+        Args:
+            filename: Optional argument providing a filename to save the plots to. Defaults to None,
+                indicating not to save the plot.
+
+        Returns:
+            A single matplotlib figure containing both the linear fit per cycle depth
+            and the decay with cycle depth.
```
richrines1, line 131: nit
```suggestion
            A single matplotlib figure containing both the linear fit per cycle depth and the decay
            with cycle depth.
```

supermarq-benchmarks/supermarq/qcvv/xeb.py
```diff
@@ -215,6 +238,8 @@ def plot_speckle(self) -> None:
         fig.colorbar(
             mpl.cm.ScalarMappable(norm, cmap), ax=axs, orientation="vertical", label="Probability"
         )
+        if filename is not None:
+            fig.savefig(filename, bbox_inches="tight")
```
richrines1, line 242: for consistency should probably return `fig` here as well
```suggestion
            fig.savefig(filename, bbox_inches="tight")

        return fig
```

richrines1 (approved):
lg % nits!

PR #1156, opened 2025-02-21
allow qubit specification for qcvv experiments

richrines1:
uses `op.qubits` instead of `LineQubit.range(<num_qubits>)` when an Operation is passed instead of a Gate in XEB/IRB


supermarq-benchmarks/supermarq/qcvv/base_experiment.py
```diff
@@ -338,13 +338,18 @@ def __init__(
             _samples: Optional list of samples to construct the experiment from
             kwargs: Additional kwargs passed to the Superstaq service object.
         """
-        self.qubits = cirq.LineQubit.range(num_qubits)
+        self.qubits: Sequence[cirq.Qid]
```
bharat-thotakura, line 341: Minor point to update the docstring of this `__init__`

bharat-thotakura (approved):
LGTM

supermarq-benchmarks/supermarq/qcvv/base_experiment.py
richrines1: ah good catch, thought i did this already. updated.

PR #1169, opened 2025-03-24
Don't pass in `css.AceCR` to `daily_integration_test.py`

vtomole:
Not passing in `css.AceCR` cause they can't be serialized anymore.

### DIFF
--- a/cirq-superstaq/cirq_superstaq/daily_integration_test.py
+++ b/cirq-superstaq/cirq_superstaq/daily_integration_test.py
@@ -26,11 +26,10 @@ def service() -> css.Service:
 
 
 def test_ibmq_compile(service: css.Service) -> None:
-    qubits = cirq.LineQubit.range(4)
+    qubits = cirq.LineQubit.range(2)
     circuit = cirq.Circuit(
-        css.AceCRMinusPlus(qubits[0], qubits[1]),
-        css.AceCRMinusPlus(qubits[1], qubits[2]),
-        css.AceCRMinusPlus(qubits[2], qubits[3]),
+        cirq.H(qubits[0]),
+        cirq.CX(qubits[0], qubits[1]),
     )
 
     out = service.ibmq_compile(circuit, target="ibmq_brisbane_qpu")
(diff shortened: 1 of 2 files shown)


cirq-superstaq/cirq_superstaq/daily_integration_test.py
```diff
@@ -26,11 +26,10 @@ def service() -> css.Service:
 
 
 def test_ibmq_compile(service: css.Service) -> None:
-    qubits = cirq.LineQubit.range(4)
+    qubits = cirq.LineQubit.range(2)
     circuit = cirq.Circuit(
-        css.AceCRMinusPlus(qubits[0], qubits[1]),
```
bharat-thotakura, line 31: I think `css.AceCR` does not need to be deleted necessarily since it, by itself, does not seem to be the cause of https://github.com/Infleqtion/client-superstaq/issues/1166. That is, I'm able to run the following snippet (with `qiskit==1.4.2`): 
```python
qubits = cirq.LineQubit.range(4)
circuit = cirq.Circuit(
    css.AceCRMinusPlus(qubits[0], qubits[1]),
    css.AceCRMinusPlus(qubits[1], qubits[2]),
    css.AceCRMinusPlus(qubits[2], qubits[3]),
)

out = service.ibmq_compile(circuit, target="ibmq_brisbane_qpu")
assert isinstance(out.circuit, cirq.Circuit)
```
as expected. It's specifically the `.pulse_gate_circuit` attribute that's broken in lines like:
https://github.com/Infleqtion/client-superstaq/blob/c328209815d0ae9a5a6f744afc1c603d24350727/cirq-superstaq/cirq_superstaq/daily_integration_test.py#L38-L39
and, 
https://github.com/Infleqtion/client-superstaq/blob/c328209815d0ae9a5a6f744afc1c603d24350727/cirq-superstaq/cirq_superstaq/daily_integration_test.py#L47-L50

So we could create a separate test function to test pulse gate circuits with a different circuit till the underlying issue is fixed. I was able to run the [IBM notebooks](https://superstaq.readthedocs.io/en/latest/optimizations/ibm/ibm.html) which use `.pulse_gate_circuit(s)` so it seems isolated to cases such as an input circuit containing a ` css.AceCRMinusPlus`

vtomole: Serialization fails when you compile. 
image

See https://github.com/Infleqtion/client-superstaq/actions/runs/14063246968/job/39379156558

bharat-thotakura: As an aside, it seems like this issue can be fixed by `qiskit.qpy.load(buf, trust_payload=True)`?

vtomole: Only changed the circuits for `qiskit_superstaq` tests.

vtomole: Why factor it out? The test passes I think https://github.com/Infleqtion/client-superstaq/actions/runs/14068868143/job/39398018037

bharat-thotakura: Ah nevermind then

bharat-thotakura (approved):
LGTM (with filing an issue for the `qss` serialization problem?)

vtomole:
Filed https://github.com/Infleqtion/client-superstaq/issues/1170

Issue #1171, opened 2025-03-26
Update pydantic requirements


cdbf1:
### What is happening?

Our current requirements in general-superstaq currently require pydantic>1.7 , but our minimum dependency checks only check down to 2.10.6 and from testing with the system repo it looks like general-superstaq fails for pydantic<2.0.

### How can we reproduce the issue?

-

### What should happen?

Update `requirements.txt` with `pydantic>=2.10.6`

### Environment

-

bharat-thotakura:
I think the issue might not stem from `general-superstaq` since its min-dependency coverage check [does test against `pydantic==1.10.7`](https://github.com/Infleqtion/client-superstaq/actions/runs/14070793681/job/39404272628#:~:text=0.2.3%20pycodestyle%2D2.12.1-,pydantic%2D1.10.7,-pyflakes%2D3.2.0%20pygments) afaict. The other min-dependency coverage checks default to using the latest (or was 3 hours ago of this comment) `v2.10.6` since they only check against the minimum versions of their own respective `*-requirements.txt` file.

cdbf1:
> I think the issue might not stem from `general-superstaq` since its min-dependency coverage check [does test against `pydantic==1.10.7`](https://github.com/Infleqtion/client-superstaq/actions/runs/14070793681/job/39404272628#:~:text=0.2.3%20pycodestyle%2D2.12.1-,pydantic%2D1.10.7,-pyflakes%2D3.2.0%20pygments) afaict. The other min-dependency coverage checks default to using the latest (or was 3 hours ago of this comment) `v2.10.6` since they only check against the minimum versions of their own respective `requirements.txt` file

Ah interesting - thanks for digging into that! I have a hunch that the general-superstaq might fail with the old version of pydantic if we added checks for the `_models.py` module as I think the email validator relies on v2. I'll have a play and see!

bharat-thotakura:
Done in https://github.com/Infleqtion/client-superstaq/pull/1172/commits/fe8cd59910d22ea2bf0fbcc943351c2a7941d00b

PR #1172, opened 2025-03-28
Add models test

cdbf1:
Investigating https://github.com/Infleqtion/client-superstaq/issues/1171

What I found is as follows:
1. Pydantic V1 is quite out of date now and only has security patch updates. I suggest we move away from it (hence updated the requirements.txt to a much more recent  version of pydantic. 
2. In order for pydantic to play nicely with older versions of python (3.8 & 3.9) I needed to change how the default model config was implemented and add a package `eval_type_backport` which allows pydantic to recognize modern type hinting when running on older python versions (e.g. `list` rather than `typing.List` and `|` rather than `typing.Union`). Unfortunately pydantic does not seem to use `__future__annotations` natively.
3. I've now added a single test that checks whether pydantic can load models as expected. When we get round to implementing the updated client for the new server, these tests can be filled out to ensure coverage.

### DIFF
--- a/general-superstaq/general_superstaq/_models_test.py
+++ b/general-superstaq/general_superstaq/_models_test.py
@@ -7,7 +7,8 @@ import pytest
 from general_superstaq import _models
 
 """This test file will be updated once the new server (v0.3.0) client is implemented.
-For now, a single test is sufficient to check that `pydantic` is loading models and working as expected.
+For now, a single test is sufficient to check that `pydantic` is loading models and working as
+expected.
 """
 
 
--- a/general-superstaq/general_superstaq/_models_test.py
+++ b/general-superstaq/general_superstaq/_models_test.py
@@ -9,3 +9,4 @@ def test_user_token_response() -> None:
         email="cameron.booker@infleqtion.com",
         token="abc",
     )
+    
--- a/general-superstaq/general_superstaq/_models_test.py
+++ b/general-superstaq/general_superstaq/_models_test.py
@@ -1,0 +1,11 @@
+# pylint: disable=missing-function-docstring,missing-class-docstring
+from __future__ import annotations
+
+from general_superstaq import _models
+
+
+def test_user_token_response() -> None:
+    _models.UserTokenResponse(
+        email="cameron.booker@infleqtion.com",
+        token="abc",
+    )
--- a/general-superstaq/general_superstaq/_models_test.py
+++ b/general-superstaq/general_superstaq/_models_test.py
@@ -1,7 +1,14 @@
 # pylint: disable=missing-function-docstring,missing-class-docstring
 from __future__ import annotations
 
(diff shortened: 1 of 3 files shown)


general-superstaq/requirements.txt
```diff
@@ -1,4 +1,5 @@
+eval_type_backport>=0.2.2
 numpy>=1.21.0
-pydantic>=1.10.7
+pydantic[email]>=2.10.6
```
bharat-thotakura, line 3: We should try to find a lower compatible version of `pydantic` since this version requirement may conflict for a user wanting to install the current version of `qiskit-ibm-runtime` (`==0.37.0`) and `qiskit-superstaq` (which relies on `general-superstaq`). This is because `qiskit-ibm-runtime` needs: `pydantic [required: >=2.5.0,<2.10]`

cdbf1: Ah yeah good spot!

cdbf1: Good call. Unfortunately the pydantic update that allows backward compatibility with python 3.8 typing is only available in >2.10.6
(PR here: https://github.com/pydantic/pydantic/pull/8209/commits)

bharat-thotakura: Just did some testing locally, and I think we might be in luck with just requiring `>=2.6.0`

general-superstaq/general_superstaq/_models_test.py
```diff
@@ -0,0 +1,23 @@
+# pylint: disable=missing-function-docstring,missing-class-docstring
+from __future__ import annotations
+
+import pydantic
+import pytest
+
+from general_superstaq import _models
+
+"""This test file will be updated once the new server (v0.3.0) client is implemented.
+For now a single test is sufficient to check that pydandic is loading models and working as expected
```
bharat-thotakura, line 10: typo nit:
```suggestion
For now, a single test is sufficient to check that `pydantic` is loading models and working as expected.
```

bharat-thotakura (approved):
LGTM

(aside comment: given the necessity of `pydantic v2` is primarily for the `v0.3.0` client, I'd assumed we'd mark that with `v0.6` of `gss, css, qss` etc, so I wonder if we hold off on merging this till that version? Of course, if this is immediately beneficial, I don't have a strong reason to stop as long as we don't have strict compatibility issue -- which I think we try with `>=2.6.0`. The only one that might have crossover is `pyquil` (though really, `qcs-api-client`) wanting `pydantic<2.0`. However, we've dropped such dependencies a while ago, so I don't see that being a particular issue)

general-superstaq/general_superstaq/_models_test.py
cdbf1: Yeah I think holding off until we can mark v0.6 is a good shout

PR #1179, opened 2025-04-16
Support enabling `COVERAGE_CORE=sysmon` in `checks-superstaq`

bharat-thotakura:


checks-superstaq/checks_superstaq/coverage_.py
```diff
@@ -54,6 +61,14 @@ def run(
         return 0
 
     coverage_args = []
+    if parsed_args.sysmon and sys.version_info.minor >= 12:
+        os.environ["COVERAGE_CORE"] = "sysmon"
+        if parsed_args.branch:
+            warnings.warn(
+                "Note: Efficient branch coverage with `COVERAGE_CORE=sysmon` may require "
+                "additional configuration."
+            )
```
richrines1, line 70: would it make sense to put this warning text in the argument's `help=` string text and then assume that if someone passes `--sysmon` they know what they're doing (instead of displaying a warning every time)

bharat-thotakura: Sounds good, done https://github.com/Infleqtion/client-superstaq/pull/1179#:~:text=%E2%80%A6-,6bc39df,-This%20branch%20has

richrines1 (approved):
lgtm!

Issue #1180, opened 2025-04-16
Failing `cirq_superstaq/ops/qudit_gates.py::test_qutrit_cz_pow_gate` with `cirq-core==1.5.0`


bharat-thotakura:
### What is happening?

Currently, the pytest checks are failing on the following test:
```python
    def test_qutrit_cz_pow_gate() -> None:
        cirq.testing.assert_eigengate_implements_consistent_protocols(
            css.QutritCZPowGate,
            setup_code="import cirq_superstaq as css, sympy",
            ignore_decompose_to_default_gateset=True,
        )

        shifted_cz3 = css.QutritCZPowGate(global_shift=0.3)

        assert css.CZ3 == css.QutritCZPowGate()
>       assert css.CZ3_INV == css.CZ3**-1 == css.CZ3**2
E       assert (css.CZ3 ** -1) == (css.CZ3 ** 2)
E        +  where css.CZ3 = css.CZ3
E        +  and   css.CZ3 = css.CZ3

cirq-superstaq/cirq_superstaq/ops/qudit_gates_test.py:157: AssertionError
```
presumably due to some change in `cirq-core==1.5.0`

### How can we reproduce the issue?

Run `pytest cirq_superstaq/cirq_superstaq/ops/qudit_gates_test.py` from directory root with `cirq-core==1.5.0` installed

### What should happen?

The test should be updated to pass for all supported `cirq-core` versions

### Environment

- **general-superstaq version**:  0.5.39
- **cirq-superstaq version**:  0.5.39
- **Operating system**:  Ubuntu
- **Python version**: 3.13.3

vtomole:
Dev meeting hypothesis: Might be a qudit issue that passed by Cirq.

PR #1182, opened 2025-04-21
Override equality comparisons for `cirq.EigenGate`'s post `cirq-core==1.5.0`

bharat-thotakura:
Fixes: https://github.com/Infleqtion/client-superstaq/issues/1180

This PR makes the following updates for compatibility with `cirq-core==1.5.0`:
- Updates the signature of `_equal_up_to_global_phase_` to be compatible with `cirq.Gate`.
- Overrides `_value_equality_values_` to follow the `cirq-core<1.5.0` implementation for `css.QutritCZPowGate()` (for now).
- Overrides `_value_equality_approximate_values_` to follow the `cirq-core<1.5.0` implementation for `css.VirtualZPowGate()` (for now).
- `mypy` typing issues in `supermarq-benchmarks/supermarq/qcvv/irb.py` raised after `cirq-core==1.5.0` upgrade

### DIFF
--- a/cirq-superstaq/cirq_superstaq/ops/qudit_gates.py
+++ b/cirq-superstaq/cirq_superstaq/ops/qudit_gates.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import abc
 from collections.abc import Sequence, Set
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
 import cirq
 import numpy as np
@@ -11,6 +11,9 @@ import sympy
 from cirq.ops.common_gates import proper_repr
 
 import cirq_superstaq as css
+
+if TYPE_CHECKING:
+    from types import NotImplementedType
 
 
 def _subscript(n: int) -> str:
@@ -39,7 +42,9 @@ class QuditSwapGate(cirq.Gate, cirq.InterchangeableQubitsGate):
     def _value_equality_values_(self) -> int:
         return self.dimension
 
-    def _equal_up_to_global_phase_(self, other: Any, atol: float) -> bool | None:
+    def _equal_up_to_global_phase_(
+        self, other: Any, atol: float = 1e-8
+    ) -> NotImplementedType | bool:
         if isinstance(other, QuditSwapGate):
             return other.dimension == self.dimension
 
@@ -140,7 +145,9 @@ class BSwapPowGate(cirq.EigenGate, cirq.InterchangeableQubitsGate):
 
         return [(0.0, projector_rest), (0.5, projector_p), (-0.5, projector_n)]
 
-    def _equal_up_to_global_phase_(self, other: Any, atol: float) -> bool | None:
+    def _equal_up_to_global_phase_(
+        self, other: Any, atol: float = 1e-8
+    ) -> NotImplementedType | bool:
         """Workaround for https://github.com/quantumlib/Cirq/issues/5980."""
 
         if not isinstance(other, BSwapPowGate):
@@ -231,7 +238,9 @@ class QutritCZPowGate(cirq.EigenGate, cirq.InterchangeableQubitsGate):
     def _value_equality_values_(self) -> tuple[float, float]:
         return self._canonical_exponent, self._global_shift
 
-    def _equal_up_to_global_phase_(self, other: Any, atol: float) -> bool | None:
+    def _equal_up_to_global_phase_(
+        self, other: Any, atol: float = 1e-8
+    ) -> NotImplementedType | bool:
         """Workaround for https://github.com/quantumlib/Cirq/issues/5980."""
 
         if isinstance(other, QutritCZPowGate):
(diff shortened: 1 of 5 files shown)


cirq-superstaq/cirq_superstaq/ops/qudit_gates.py
```diff
@@ -198,14 +214,34 @@ def _qid_shape_(self) -> tuple[int, int]:
...
+                self._canonical_exponent_cached = self._exponent
+            elif cirq.is_parameterized(self._exponent):
+                self._canonical_exponent_cached = self._exponent
+                if isinstance(self._exponent, sympy.Number):
+                    self._canonical_exponent_cached = float(self._exponent)
+            else:
+                self._canonical_exponent_cached = self._exponent % period
+        return self._canonical_exponent_cached
+
+    def _value_equality_values_(self) -> tuple[float | None, float]:
```
richrines1, line 239: so it looks like this [does](https://github.com/quantumlib/Cirq/blob/197c83a87fd8184f9ced0a111e50c08f768ed0da/cirq-core/cirq/ops/eigen_gate.py#L306) have a default implementation in `cirq.EigenGate`). however our test asserting `css.CZ3**-1 == css.CZ3**2` still fails due to a rounding error:
```python
print((css.CZ3**2)._value_equality_values_())
print((css.CZ3**-1)._value_equality_values_())
```
prints
```
(cirq.PeriodicValue(0.0, 2), cirq.PeriodicValue(1.3333333333333333, 2), cirq.PeriodicValue(0.6666666666666665, 2))
(cirq.PeriodicValue(0.0, 2), cirq.PeriodicValue(1.3333333333333335, 2), cirq.PeriodicValue(0.6666666666666667, 2))
```

not sure what the best workaround is though. imo it would be good to stick as close as possible to the default implementation (i.e. return a tuple of PeriodicValues). notably this issue also affects some cirq builtin gates, e.g.
```python
assert cirq.XPowGate(dimension=3) ** 2 == cirq.XPowGate(dimension=3) ** -1
```
passes on `cirq<1.5.0` but fails on newer versions - so maybe we can just remove that assertion for now? (and instead assert `cirq.approx_eq(css.CZ3**-1, css.CZ3**2)`, which still passes)

supermarq-benchmarks/supermarq/stabilizers.py
```diff
@@ -133,11 +133,12 @@ def construct_stabilizer(
     pauli_basis = []
     start_idx = 0
     for i, row in enumerate(echelon_matrix.T):
-        if 1 in row[start_idx:]:
+        matrix_row = np.array(row)
```
richrines1, line 136: nit (avoids unnecessary copies)
```suggestion
        matrix_row = np.asarray(row)
```

cirq-superstaq/cirq_superstaq/ops/qudit_gates.py
```diff
@@ -321,9 +357,16 @@ def _value_equality_values_(self) -> tuple[object, ...]:
         return (*super()._value_equality_values_(), self._dimension, self._level)
 
     def _value_equality_approximate_values_(self) -> tuple[object, ...]:
-        return (*super()._value_equality_approximate_values_(), self._dimension, self._level)
+        period = self._period()
+        if not period or cirq.is_parameterized(self._exponent):
+            exponent = self._exponent
+        else:
+            exponent = cirq.PeriodicValue(self._exponent, period)
+        return (exponent, self._global_shift, self._dimension, self._level)
```
richrines1, line 365: afaict everything works correctly if we just use
```suggestion
        return self._value_equality_values_()
```

cirq-superstaq/cirq_superstaq/ops/qudit_gates.py
```diff
@@ -198,14 +214,34 @@ def _qid_shape_(self) -> tuple[int, int]:
     def _eigen_components(self) -> list[tuple[float, npt.NDArray[np.float64]]]:
         eigen_components = []
         exponents = np.kron(range(self.dimension), range(self.dimension)) % self.dimension
-        for x in sorted(set(exponents)):
+        int_exponents = {int(i) for i in exponents}
+        for x in sorted(int_exponents):
```
richrines1, line 218: does this work?
```suggestion
        for x in sorted(np.unique(exponents)):
```

bharat-thotakura: The issue was just the following `mypy` error:
```bash
error: Value of type variable "SupportsRichComparisonT" of "sorted" cannot be "signedinteger[Any]"  [type-var]
```

cirq-superstaq/cirq_superstaq/ops/qudit_gates.py
bharat-thotakura: Thanks for looking into this more! If we aren't losing any underlying functionality across the `cirq-core` versions, I would agree to stick to the default implementation. 

I'll leave it to a future PR to consider an override if more beneficial

richrines1 (approved):
lgtm!

PR #1178, opened 2025-04-02
Support qiskit `v2.0.0`

bharat-thotakura:
Fixes: https://github.com/Infleqtion/client-superstaq/issues/1177

PR #1183, opened 2025-04-21
Dropping Python 3.8 support (+ adding README badges)

bharat-thotakura:
This PR officially drops Python 3.8 support for all `client-superstaq` packages based on the need to drop Python 3.8 in https://github.com/Infleqtion/client-superstaq/pull/1178 (namely, supporting Qiskit `v2.0.0` requires `qiskit-superstaq`'s minimum qiskit support to be `v1.3.0`  (which has no Python 3.8 support) for QPY reasons).

Given Python 3.8 had reached [EOL](https://devguide.python.org/versions/) on 2024-10-07, this PR extends the requirement to at least Python 3.9 moving forward.

As a bonus, this PR includes new README badges to highlight the supported Python versions, as well as offering other useful at-a-glance information.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -105,7 +105,7 @@ jobs:
     name: Pytest and Coverage check
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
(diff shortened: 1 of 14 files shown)


richrines1 (approved):
lgtm! (and just disabled the 3.8 test requirement)

richrines1:
> Just flagging that we may want to make 'Pytest and Coverage check (3.13) ' required now for the sake of consistency

good call, done.

PR #1185, opened 2025-04-25
Ignore `eval_type_backport` if `python_version > '3.9'`

bharat-thotakura:
This PR makes `eval_type_backport` somewhat optional in the spirit of minimal package dependencies.

It also serves as a reminder to remove the requirement when Python 3.9 support is dropped.


richrines1 (approved):
lgtm!

PR #1184, opened 2025-04-25
Ruff linting fixes for `css`, `gss`, `qss`, and `checks`

bharat-thotakura:
This PR applies various fixes highlighted by `ruff check` (with select rules) to enforce better consistency. It also preps `client-superstaq` to use `ruff` linting moving forward to replace the current `pylint` + `flake8` set-up. 

Linting updates for `supermarq-benchmarks` are intended to be done in a separate PR.

PR #1188, opened 2025-04-29
Use `ruff` in place of `black`, `flake8`, and `pylint` in `client-superstaq`

bharat-thotakura:
With the fixes in https://github.com/Infleqtion/client-superstaq/pull/1184 and https://github.com/Infleqtion/client-superstaq/pull/1187, this PR transitions the CI and the local `checks/` directory to use `ruff` in place of formatting with `black` (and `isort`), `checks/pylint_.py`, `checks/flake8_.py`

It also enables all `"RUF"` (except `"RUF001"`, `"RUF002"`), `“FURB”`, `“PGH”`,  `"PYI"` (& `"NPY201"`, `"PT018"`, `"SIM114"`, `"SIM300"`) rules as a bonus


ruff.toml
richrines1, line 1: can we not put all this in `pyproject.toml`?

bharat-thotakura: I wasn't sure if that would impact other projects using `checks-superstaq` since this repos `pyproject.toml` uses `checks-pyproject.toml` which is a template for the package?

richrines1: i think it should be ok - it should only affect the results of the configs_.py check, which others can either ignore or take as motivation to start making the same changes (once we're done making ruff-related changes it might be nice to make a pr in qldpc before deploying though)

richrines1: that said maybe we shouldn't remove the flake8/black/pylint parts of the existing pyproject.toml until we remove the corresponding checks

richrines1 (approved):
lgtm!

bharat-thotakura:
Will merge after the Flake8 check is dropped, but also wanted to mention and follow up on:
>once we're done making ruff-related changes it might be nice to make a pr in qldpc before deploying though

I tried enabling all the rules from this PR in the `qldpc` library, and it raised:
```
78 errors (of which 22 fixable with the `--fix` option (39 hidden fixes can be enabled with the `--unsafe-fixes` option).
```
of non-docstring, print, and commented-out code errors. 

Ultimately though, I think the config present in [its `pyproject.toml`](https://github.com/Infleqtion/qLDPC/blob/af2e3df919c459402456da909ddf793e745d9e92/pyproject.toml#L57-L60) will take precedence and these errors won't be raised (at least, when checked against an editable-install of `checks-superstaq` based off this branch).

richrines1:
> after the Flake8 check is dropped

sorry forgot about this - done!

PR #1190, opened 2025-05-01
clean up qcvv json resolution

richrines1:
removes the custom `_from_json_dict_` methods from qcvv classes


cdbf1:
Ahh nice, thats much cleaner thanks!

PR #1192, opened 2025-05-07
Client versioning

vietphamngoc:
Pull request to support client side job refactor

### DIFF
--- a/general-superstaq/general_superstaq/superstaq_client.py
+++ b/general-superstaq/general_superstaq/superstaq_client.py
@@ -791,8 +791,8 @@ class _BaseSuperstaqClient(ABC):
             if isinstance(json_content, dict) and set(json_content.keys()).intersection(
                 {"message", "details"}
             ):
-                alternative = json_content.get("details", "")
-                message = json_content.get("message", alternative)
+                alternative: str = json_content.get("details", "")
+                message: str = json_content.get("message", alternative)
             else:
                 message = str(response.text)
 


dowusu-antwi (commented):
Just a couple of comments on the tests:

general-superstaq/general_superstaq/superstaq_client_test.py
```diff
@@ -24,34 +27,75 @@
...
 
 
-def test_superstaq_client_args() -> None:
+@pytest.mark.parametrize("api_version", ["v0.2.0", "v0.3.0"])
+def test_superstaq_client_args(api_version: str) -> None:
     client = gss.superstaq_client._SuperstaqClient(
         client_name="general-superstaq",
         remote_host="http://example.com",
         api_key="to_my_heart",
+        api_version=api_version,
```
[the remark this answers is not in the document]
vietphamngoc: I believe it can be replaced by a fixture by using a fixture factory but given that these "non-standard" clients are only used in a few instances I didn't think it was needed.
Happy to add the fixture factory if I'm wrong.

general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -10,27 +10,94 @@
...
+RECOGNISED_CIRCUIT_TYPES = Literal[_models.CircuitType.CIRQ, _models.CircuitType.QISKIT]
+"""The circuit types that are currently implemented within the SuperstaqClient."""
+
+
+class _API_Version(str, enum.Enum):
+    V0_2_0 = "v0.2.0"
+    V0_3_0 = "v0.3.0"
+
+
+class _versioned_method:
```
vtomole, line 50: Camel space your class names.

vtomole (dismissed):
From dev meeting: Better method is to create a new private client for version 3 that `Service` can call if they want to use it instead of version 2.

general-superstaq/general_superstaq/superstaq_client.py
dowusu-antwi: @vietphamngoc don't forget to camel space this (and `_API_Version` above)

general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -10,27 +10,94 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+# mypy: disable-error-code="empty-body, operator"
```
vtomole, line 14: Remove this comment

cdbf1:
LGTM % one minor comment (which I'm happy if we want to ignore)

general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -954,6 +910,643 @@ def __repr__(self) -> str:
...
+        method: str = "sim_anneal",
+        max_solutions: int | None = 1000,
+        *,
+        qaoa_depth: int = 1,
+        rqaoa_cutoff: int = 0,
+        dry_run: bool = False,
+        random_seed: int | None = None,
+        **kwargs: object,
+    ) -> dict[str, str]:
+        gss.validation.validate_target(target)
```
cdbf1, line 1059: Can we add any validation that is common between both clients into the base methods?

cdbf1 (approved):

general-superstaq/general_superstaq/superstaq_client.py
vietphamngoc: Added some common validations. Happy to discuss the best way to do it.

PR #1196, opened 2025-05-21
Version bump to 0.5.42

bharat-thotakura:
Deploying for https://github.com/Infleqtion/client-superstaq/pull/1178


bharat-thotakura:
(checks should pass after https://github.com/Infleqtion/client-superstaq/pull/1197)

PR #1199, opened 2025-05-27
Adding barriers between cycles

vietphamngoc:
Adding barriers to prevent the compiler from reducing the circuits to `Id`.


supermarq-benchmarks/supermarq/qcvv/base_experiment.py
```diff
@@ -508,7 +508,17 @@ def _interleave_op(
...
         interleaved_circuit = circuit.copy()
         interleaved_circuit.batch_insert(
-            [(k, operation) for k in range(len(circuit) - int(not include_final), 0, -1)]
+            [
+                (
+                    k,
+                    [
+                        css.barrier(*interleaved_circuit.all_qubits()),
+                        operation,
+                        css.barrier(*interleaved_circuit.all_qubits()),
```
richrines1, line 517:
```suggestion
                        css.barrier(*self.qubits),
                        operation,
                        css.barrier(*self.qubits),
```

richrines1 (commented):
one request but otherwise lg!

supermarq-benchmarks/supermarq/qcvv/base_experiment.py
richrines1: (sorry just realized this also requires making `_interleave_op` not a staticmethod (which imo it shouldn't be anyway))

richrines1: interesting, yeah maybe (thought ideally it would be nice if we could make the same method work for more protocols)

probably not for this pr though - would love to get this in so i can build on it :)

richrines1:
@vietphamngoc is this ready to merge?

PR #1201, opened 2025-05-28
TESTING: numpy pin

vtomole:


vtomole:
My local python 3.12 was broken.

Issue #1103, opened 2024-11-14
Pytest Windows flake for `supermarq-benchmarks\supermarq\qcvv\xeb_test.py`


natibek:
### What is happening?

Pytest Windows flakes with error `_tkinter.TclError: invalid command name "tcl_findLibrary"` for `supermarq-benchmarks\supermarq\qcvv\xeb_test.py`.

<details>
<summary>supermarq-benchmarks\supermarq\qcvv\xeb_test.py </summary>

```bash

============================= test session starts =============================
platform win32 -- Python 3.12.7, pytest-8.3.3, pluggy-1.5.0
Using --randomly-seed=2676077198
rootdir: D:\a\client-superstaq\client-superstaq
configfile: pyproject.toml
plugins: nbmake-1.5.4, cov-6.0.0, randomly-3.16.0, socket-0.7.0, xdist-3.6.1
created: 2/2 workers
2 workers [561 items]

........................................................................ [ 12%]
.....s.................................................................. [ 25%]
..............................................................F......... [ 38%]
........................................................................ [ 51%]
........................................................................ [ 64%]
........................................................................ [ 77%]
........................................................................ [ 89%]
Matplotlib is building the font cache; this may take a moment.
........s................................................                [100%]
================================== FAILURES ===================================
__________________________ test_xeb_analyse_results ___________________________
[gw0] win32 -- Python 3.12.7 C:\hostedtoolcache\windows\Python\3.12.7\x64\python.exe

xeb_experiment = <supermarq.qcvv.xeb.XEB object at 0x00000248E6459EE0>

    def test_xeb_analyse_results(xeb_experiment: XEB) -> None:
        xeb_experiment._samples = MagicMock()
        # Choose example data to give perfect fit with fidelity=0.95
        xeb_experiment._raw_data = pd.DataFrame(
            [
                {
                    "cycle_depth": 1,
                    "circuit_depth": 3,
                    "sum_p(x)p(x)": 0.3,
                    "sum_p(x)p^(x)": 0.95**1 * 0.3,
                },
                {
                    "cycle_depth": 1,
                    "circuit_depth": 3,
                    "sum_p(x)p(x)": 0.5,
                    "sum_p(x)p^(x)": 0.95**1 * 0.5,
                },
                {
                    "cycle_depth": 5,
                    "circuit_depth": [11](https://github.com/Infleqtion/client-superstaq/actions/runs/11825271337/job/32948670957#step:6:12),
[...]

PR #1202, opened 2025-05-29
parallel xeb

richrines1:
qcvv updates to support parallel XEB experiments:
* allows XEB interleaved layers to contain multiple gates on any number of qubits
* adds indexing to QCVVResults objects, so that e.g. `results[q0, q1]` will return a new results object for analysis on just those qubits


supermarq-benchmarks/supermarq/qcvv/base_experiment_test.py
```diff
@@ -163,6 +163,108 @@ def test_results_init(
...
     assert results.num_qubits == 2
+    assert results.qubits == (cirq.q(0), cirq.q(1))
+
+
+def test_results_getitem(
+    abc_experiment: ExampleExperiment,
+) -> None:
+    q0, q1 = abc_experiment.qubits
+
+    results = abc_experiment.run_with_callable(lambda c: {"00": 10, "01": 10})
```
dowusu-antwi, line 174: Does the experiment expect counts (`"00":10, "01": 10`) and then convert it into a probability distribution (`{"00": 0.5, "01": 0.5}`)?

dowusu-antwi (approved):
lgtm

dowusu-antwi:
Can we close https://github.com/Infleqtion/client-superstaq/issues/1103 now?

supermarq-benchmarks/supermarq/qcvv/base_experiment_test.py
richrines1: it will accept either probabilities or counts (and in the latter case automatically converts to a probability distribution). But that's not what's being tested here so tbh not sure why i did it this way..

updated to instantiate the results object directly instead of this weird use of `run_with_callable`

PR #1207, opened 2025-06-04
Added warning attribute to `JobCancellationResults`

dowusu-antwi:
This is to fix a server-side issue for today's deploy: a server-side test expects `JobCancellationResults` to have a `warnings` attribute for catching a failed attempted at cancelling a job that already has a status.


dowusu-antwi:
Closing because the fix was server-side.

PR #1208, opened 2025-06-06
fix `backend.target` failures for missing target_info

richrines1:
fixes `backend.target` error when the backend's `target_info` doesn't contain timing constraints


vtomole (approved):
LGTM

PR #1210, opened 2025-06-06
support custom gates in `SuperstaqBackend.target`

richrines1:
currently `backend.target` fails for devices with custom gates, with:
```
KeyError: 'The specified basis gate: <name> is not present in the standard gate names or a provided custom_name_mapping'
```

### DIFF
--- a/qiskit-superstaq/qiskit_superstaq/conftest.py
+++ b/qiskit-superstaq/qiskit_superstaq/conftest.py
@@ -1,8 +1,8 @@
 from __future__ import annotations
 
 import general_superstaq as gss
-import general_superstaq.testing
 import pytest
+from general_superstaq.testing import RETURNED_TARGETS
 
 import qiskit_superstaq as qss
 
@@ -31,7 +31,7 @@ class MockSuperstaqClient(gss.superstaq_client._SuperstaqClient):
         Returns:
             A list of Superstaq targets matching all provided criteria.
         """
-        return gss.testing.RETURNED_TARGETS
+        return RETURNED_TARGETS
 
     def target_info(self, target: str, **kwargs: object) -> dict[str, object]:
         """Mocks a request to the /target_info endpoint.


qiskit-superstaq/qiskit_superstaq/conftest.py
```diff
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import general_superstaq as gss
+import general_superstaq.testing
```
bharat-thotakura: Is this import to replace `gss.testing.RETURNED_TARGETS`? If not, I think we can remove it:
```suggestion
```

qiskit-superstaq/qiskit_superstaq/conftest.py
```diff
@@ -82,6 +83,7 @@ def target_info(self, target: str, **kwargs: object) -> dict[str, object]:
                         ["x", [0], 3.5555555555555554e-08, "s"],
                         ["x", [1], 3.5555555555555554e-08, "s"],
                         ["x", [2], 3.5555555555555554e-08, "s"],
+                        ["gr", [0, 1, 2, 3], 1.5e-5, "s"],
```
bharat-thotakura, line 86: I don't think we include custom gates in `"gate_durations"` server-side in `target_info()`. Does that cause issues in initializing `qiskit.transpiler.Target`?

qiskit-superstaq/qiskit_superstaq/superstaq_backend.py
bharat-thotakura, line 1: Thoughts on overriding the `num_qubits` property: 
```
    @property
    def num_qubits(self) -> int:
        target_info = self.target_info()
        return target_info["num_qubits"]
```
It seems like you have to jump over a lot of hoops to retrieve a simple backend property

qiskit-superstaq/qiskit_superstaq/conftest.py
richrines1: nope (e.g. it's not included in the mock target info in `test_target_with_custom_gates`) - i just added it here to check/show that it's possible

qiskit-superstaq/qiskit_superstaq/conftest.py
richrines1: `gss.testing` is used in `conftest.py` but not loaded by default, so without it coverage fails for some specific sets of tests - e.g.:
```
./checks/coverage_.py qiskit-superstaq/qiskit_superstaq/conftest.py qiskit-superstaq/qiskit_superstaq/superstaq_backend.py 
```
(which i ran into using `./checks/coverage_.py -i` when prepping this pr)

also maybe an weird use case, but e.g. the following will also fail without the import:
```python
import qiskit_superstaq.conftest
mock_provider = qiskit_superstaq.conftest.MockSuperstaqProvider()
_ = mock_provider.get_targets()
```

qiskit-superstaq/qiskit_superstaq/superstaq_backend.py
richrines1: yeah great idea, just added.

qiskit-superstaq/qiskit_superstaq/conftest.py
bharat-thotakura: Oh I didn't know that, thanks for the example!

richrines1: (just changed it to `from general_superstaq.testing import RETURNED_TARGETS` and used `return RETURNED_TARGETS` in that test to hopefully make this less opaque)

PR #1209, opened 2025-06-06
bump version to v0.5.44

richrines1:
(for #1208)

PR #1211, opened 2025-06-06
`qss` backend properties regression in `daily_integration_test.py`

bharat-thotakura:
Skipping integration tests like `test_submit_to_sqale_qubit_sorting` allowed some bugs like https://github.com/Infleqtion/client-superstaq/pull/1209 and https://github.com/Infleqtion/client-superstaq/pull/1210 to go under the radar. This PR adds some tests to check such functionality daily

### DIFF
--- a/qiskit-superstaq/qiskit_superstaq/daily_integration_test.py
+++ b/qiskit-superstaq/qiskit_superstaq/daily_integration_test.py
@@ -35,10 +35,13 @@ def test_backends(provider: qss.SuperstaqProvider) -> None:
         retired=False,
         accessible=True,
     )
+    all_backends = provider.backends()
     assert ibmq_backend_info in result
     assert provider.get_backend("ibmq_brisbane_qpu").name == "ibmq_brisbane_qpu"
-    assert len(provider.backends()) == len(result)
+    assert provider.get_backend("cq_sqale_qpu").target_info()["target"] == "cq_sqale_qpu"
+    assert len(all_backends) == len(result)
     assert all(target in result for target in filtered_result)
+    assert all(backend.target.num_qubits is not None for backend in all_backends)
 
 
 def test_ibmq_compile(provider: qss.SuperstaqProvider) -> None:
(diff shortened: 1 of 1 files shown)


cirq-superstaq/cirq_superstaq/daily_integration_test.py
```diff
@@ -177,6 +177,16 @@ def test_get_targets(service: css.Service) -> None:
     assert ibmq_target_info in result
     assert aqt_target_info in result
     assert all(target in result for target in filtered_result)
+    for gss_target in result:
+        target_name = gss_target.target
+        try:
+            assert (
+                service.target_info(target_name)["target"] == target_name
+            )  # Checks `.target_info()` is retrievable
+        except Exception as e:
```
richrines1: why do we need to catch these? this is usually run automatically so we probably won't notice the prints if the test passes

bharat-thotakura: My thought was the integration test failing and making an issue every single time a provider API is down might be too much, but having a debug check to review could be helpful. However, if this is rarer than I'd initially imagined I'm also happy to let the test fail normally

bharat-thotakura: Ended up deciding to revert the exception catching in [df6f89d](https://github.com/Infleqtion/client-superstaq/pull/1211/commits/df6f89d53ae5aa0a315220e9b02a484a17166348)

qiskit-superstaq/qiskit_superstaq/superstaq_backend.py
```diff
@@ -103,6 +103,11 @@ def target(self) -> qiskit.transpiler.Target:
             "dd": qss.DDGate,
             "gr": qiskit.circuit.library.GR,
             "iccx_o0": qss.AQTiCCXGate,
+            "GPI": qiskit.circuit.library.UGate,
+            "GPI2": qiskit.circuit.library.UGate,
+            "prx": qiskit.circuit.library.UGate,
```
vtomole, line 108: Make a low-pri issue to use a qiskit R gate instead of a U gate. Similiar for GPI and GPI 2

vtomole (approved):
LGTM

qiskit-superstaq/qiskit_superstaq/superstaq_backend.py
bharat-thotakura: Done: https://github.com/Infleqtion/client-superstaq/issues/1223

PR #1214, opened 2025-06-11
Ruff is now the default and sole tool for formatting and linting.

KerryZhou1024:
Removes support and configuration for black, isort, flake8, and pylint.
Local runs seem successful.


checks-superstaq/checks_superstaq/all_.py
```diff
@@ -38,6 +35,8 @@ def run(*args: str) -> int:
         """
     )
 
+    # "--ruff" flag removed
```
vtomole:
```suggestion
```

checks/all_.py
```diff
@@ -6,7 +6,9 @@
 import checks_superstaq as checks
 
 if __name__ == "__main__":
-    args = [*sys.argv[1:], "--ruff"]
+    args = sys.argv[1:]  # Removed the automatic "--ruff" addition as ruff is
```
vtomole:
```suggestion
    args = sys.argv[1:]
```

checks/all_.py
```diff
@@ -6,7 +6,9 @@
...
+    # now the default for lints and formats
```
vtomole:
```suggestion
```

checks-superstaq/checks_superstaq/all_.py
```diff
@@ -75,25 +69,23 @@ def run(*args: str) -> int:
...
-    if parsed_args.ruff:
-        checks_failed |= ruff_format_.run(
-            *args_to_pass, exit_on_failure=exit_on_failure, silent=True
-        )
-        checks_failed |= ruff_lint_.run(*args_to_pass, exit_on_failure=exit_on_failure, silent=True)
-    else:
-        checks_failed |= format_.run(*args_to_pass, exit_on_failure=exit_on_failure, silent=True)
-        checks_failed |= flake8_.run(*args_to_pass, exit_on_failure=exit_on_failure, silent=True)
-        checks_failed |= pylint_.run(*args_to_pass, exit_on_failure=exit_on_failure, silent=True)
+    exit_on_failure_formats_lints = not (parsed_args.force_formats or parsed_args.force_all)
```
vtomole: Why did you switch this from `exit_on_failure` to `exit_on_failure_formats_lints`?

checks-superstaq/checks_superstaq/all_.py
bharat-thotakura, line 1: Given that `flake8_`, `format_`, and `pylint_` are no longer used, I think we can delete those files from `check-superstaq/` (along with the `pylint` plugin files since `pylint` is no longer in the `requirements.txt`).

Additionally, now that `ruff` is the only formatter and linter, we don't need to distinguish the file names, so following https://github.com/Infleqtion/client-superstaq/pull/1016, we can now rename `ruff_format` -> `format_` and `ruff_lint_` -> `lint_`

checks-superstaq/requirements.txt
bharat-thotakura, line 1: Thoughts on bumping the minimum `ruff` version now that it is a core dependency @vtomole, @richrines1? The latest `v0.11.*` is generally more robust than the earlier `v0.5.6` and still maintains Python 3.7+ compatiblity

bharat-thotakura:
This PR also makes a breaking change in dropping the `--ruff` flag. We should give a heads-up to `checks-superstaq` users before deploying, or potentially still silently accept the `--ruff` temporarily with a deprecation warning

checks-superstaq/requirements.txt
richrines1: imo it's fine to require `~=0.11.0`

KerryZhou1024:
Ready for review!

vtomole (approved):
LGTM

PR #1215, opened 2025-06-12
Adding support for new v0.3.0 Job standard

vietphamngoc:
API v0.3.0 requires a different Job structure for css and qss.

### DIFF
--- a/cirq-superstaq/cirq_superstaq/daily_integration_test.py
+++ b/cirq-superstaq/cirq_superstaq/daily_integration_test.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import os
+from typing import TYPE_CHECKING, cast
 
 import cirq
 import general_superstaq as gss
@@ -12,6 +13,9 @@ import qiskit
 from general_superstaq import ResourceEstimate
 
 import cirq_superstaq as css
+
+if TYPE_CHECKING:
+    from cirq_superstaq.job import Job
 
 
 @pytest.fixture
@@ -301,6 +305,8 @@ def test_job(service: css.Service) -> None:
     multi_job = service.create_job(
         [circuit, circuit_alt], target="ibmq_brisbane_qpu", repetitions=10, method="dry-run"
     )
+    job = cast("Job", job)
+    multi_job = cast("Job", multi_job)
 
     job_id = job.job_id()  # To test for https://github.com/Infleqtion/client-superstaq/issues/452
     multi_job_id = multi_job.job_id()
--- a/cirq-superstaq/cirq_superstaq/job.py
+++ b/cirq-superstaq/cirq_superstaq/job.py
@@ -25,6 +25,7 @@ import cirq
 import general_superstaq as gss
 from cirq._doc import document
 from general_superstaq import _models
+from general_superstaq.superstaq_client import _SuperstaqClientV3
 
 import cirq_superstaq as css
 
@@ -501,7 +502,7 @@ class JobV3:  # pragma: no cover
     If a job is canceled or deleted, only the job id and the status remain valid.
     """
 
-    def __init__(self, client: gss.superstaq_client._SuperstaqClientV3, job_id: uuid.UUID) -> None:
+    def __init__(self, client: gss.superstaq_client._SuperstaqClientV3, job_id: str) -> None:
         """Constructs a `Job`.
 
         Users should not call this themselves. If you only know the `job_id`, use `fetch_jobs`
@@ -512,7 +513,7 @@ class JobV3:  # pragma: no cover
             job_id: Unique identifier for the job.
         """
         self._client = client
-        if self._client.api_version != "v0.3.0":
+        if not isinstance(self._client, _SuperstaqClientV3):
             raise NotImplementedError(
                 "JobV3 job can only be used with v0.3.0 of the Superstaq API."
             )
@@ -601,7 +602,15 @@ class JobV3:  # pragma: no cover
         """
         return self._job_id
 
-    def status(self, index: int | None = None) -> str:
+    @overload
+    def status(self, index: int) -> _models.CircuitStatus: ...
+
+    @overload
+    def status(self, index: None = None) -> list[_models.CircuitStatus]: ...
+
+    def status(
+        self, index: int | None = None
+    ) -> _models.CircuitStatus | list[_models.CircuitStatus]:
         """Gets the current status of the job.
 
         If the current job is in a non-terminal state, this will update the job and return the
--- a/cirq-superstaq/cirq_superstaq/job.py
+++ b/cirq-superstaq/cirq_superstaq/job.py
@@ -493,7 +493,7 @@ class Job:
 
 
 @cirq.value_equality(unhashable=True)
-class JobV3:  # pragma: no cover
+class JobV3:
     """***NEW API STANDARD*** A job created on the Superstaq API.
 
     Note that this is mutable, when calls to get status or results are made the job updates itself
@@ -502,7 +502,7 @@ class JobV3:  # pragma: no cover
     If a job is canceled or deleted, only the job id and the status remain valid.
     """
 
-    def __init__(self, client: gss.superstaq_client._SuperstaqClientV3, job_id: str) -> None:
+    def __init__(self, client: gss.superstaq_client._SuperstaqClientV3, job_id: uuid.UUID) -> None:
         """Constructs a `Job`.
 
         Users should not call this themselves. If you only know the `job_id`, use `fetch_jobs`
@@ -514,13 +514,10 @@ class JobV3:  # pragma: no cover
         """
         self._client = client
         if not isinstance(self._client, _SuperstaqClientV3):
-            raise NotImplementedError(
-                "JobV3 job can only be used with v0.3.0 of the Superstaq API."
-            )
+            raise ValueError("JobV3 job can only be used with v0.3.0 of the Superstaq API.")
         self._overall_status = _models.CircuitStatus.RECEIVED
         self._job_data: _models.JobData
         self._job_id = job_id
-        self._refresh_job()
 
     def _refresh_job(self) -> None:
         """If the last fetched job is not terminal, gets the job from the API."""
@@ -548,7 +545,6 @@ class JobV3:  # pragma: no cover
             _models.CircuitStatus.AWAITING_COMPILE,
             _models.CircuitStatus.AWAITING_SUBMISSION,
             _models.CircuitStatus.AWAITING_SIMULATION,
-            _models.CircuitStatus.AWAITING_SUBMISSION,
             _models.CircuitStatus.PENDING,
             _models.CircuitStatus.RUNNING,
             _models.CircuitStatus.FAILED,
@@ -602,15 +598,7 @@ class JobV3:  # pragma: no cover
         """
         return self._job_id
 
-    @overload
-    def status(self, index: int) -> _models.CircuitStatus: ...
-
-    @overload
-    def status(self, index: None = None) -> list[_models.CircuitStatus]: ...
-
-    def status(
-        self, index: int | None = None
-    ) -> _models.CircuitStatus | list[_models.CircuitStatus]:
+    def status(self, index: int | None = None) -> _models.CircuitStatus:
         """Gets the current status of the job.
 
         If the current job is in a non-terminal state, this will update the job and return the
@@ -658,6 +646,8 @@ class JobV3:  # pragma: no cover
             ~gss.SuperstaqServerException: If unable to get the status of the job
                 from the API.
         """
+        if not hasattr(self, "_job_data"):
+            self._refresh_job()
         return self._job_data.target
 
     @overload
@@ -682,6 +672,8 @@ class JobV3:  # pragma: no cover
             ~gss.SuperstaqServerException: If unable to get the status of the job
                 from the API.
         """
+        if not hasattr(self, "_job_data"):
+            self._refresh_job()
         num_qubits = []
         if index is None:
             to_check = list(range(self._job_data.num_circuits))
@@ -704,28 +696,9 @@ class JobV3:  # pragma: no cover
         Returns:
             The number of repetitions for this job.
         """
+        if not hasattr(self, "_job_data"):
+            self._refresh_job()
         return self._job_data.shots[0]
-
-    @overload
-    def _get_circuits(self, circuit_type: str, index: int) -> cirq.Circuit: ...
-
-    @overload
-    def _get_circuits(self, circuit_type: str, index: None = None) -> list[cirq.Circuit]: ...
-
-    def _get_circuits(
-        self, circuit_type: str, index: int | None = None
-    ) -> cirq.Circuit | list[cirq.Circuit]:
-        """Retrieves the corresponding circuit(s) to `circuit_type`.
-
-        Args:
-            circuit_type: The kind of circuit(s) to retrieve. Either "input_circuit" or
-                "compiled_circuit".
-            index: The index of the circuit to retrieve.
-
-        Returns:
-            A single circuit or list of circuits.
-        """
-        raise NotImplementedError
 
     @overload
     def compiled_circuits(self, index: int) -> cirq.Circuit: ...
@@ -746,6 +719,8 @@ class JobV3:  # pragma: no cover
         Returns:
             A single compiled circuit or list of compiled circuits.
         """
+        if not hasattr(self, "_job_data"):
+            self._refresh_job()
         if all(c is None for c in self._job_data.compiled_circuits):
             raise RuntimeError(f"The job {self._job_id} has no compiled circuits.")
 
@@ -781,6 +756,8 @@ class JobV3:  # pragma: no cover
         Returns:
             A single input circuit or list of submitted input circuits.
         """
+        if not hasattr(self, "_job_data"):
+            self._refresh_job()
         circuits = [
             css.deserialize_circuits(self._job_data.input_circuits[k])[0]
             for k in range(self._job_data.num_circuits)
@@ -789,22 +766,6 @@ class JobV3:  # pragma: no cover
         if index is None:
             return circuits
         return circuits[index]
-
-    def pulse_gate_circuits(self, index: int | None = None) -> Any:
-        """Gets the pulse gate circuit(s) returned by this job.
-
-        Args:
-            index: An optional index of the pulse gate circuit to retrieve.
-
-        Returns:
-            A `qiskit.QuantumCircuit` pulse gate circuit or list of `qiskit.QuantumCircuit` pulse
-            gate circuits.
-
-        Raises:
-            ValueError: If the job was not run on an IBM pulse device.
-            ValueError: If the job's target does not use pulse gate circuits.
-        """
-        raise NotImplementedError
 
     @overload
     def counts(
@@ -897,17 +858,18 @@ class JobV3:  # pragma: no cover
         Returns:
             A dictionary containing updated job information.
         """
-        self._refresh_job()
+        if not hasattr(self, "_job_data"):
+            self._refresh_job()
         return self._job_data.model_dump()
 
     def __str__(self) -> str:
         return f"Job with job_id={self.job_id()}"
 
     def __repr__(self) -> str:
-        return f"css.Job(client={self._client!r}, job_id={self.job_id()!r})"
+        return f"css.JobV3(client={self._client!r}, job_id={self.job_id()!r})"
 
     def _value_equality_values_(self) -> tuple[uuid.UUID, dict[str, Any] | None]:
-        if self._job_data is None:
+        if not hasattr(self, "_job_data"):
             return self._job_id, None
         return self._job_id, self._job_data.model_dump()
 
--- a/cirq-superstaq/cirq_superstaq/job.py
+++ b/cirq-superstaq/cirq_superstaq/job.py
@@ -22,6 +22,7 @@ from typing import Any, overload
 
 import cirq
 import general_superstaq as gss
+import general_superstaq._models as _models
 from cirq._doc import document
 
 import cirq_superstaq as css
@@ -60,7 +61,7 @@ class Job:
         "data associated with it beyond an id and a status.",
     )
 
-    def __init__(self, client: gss.superstaq_client._BaseSuperstaqClient, job_id: str) -> None:
+    def __init__(self, client: gss.superstaq_client._SuperstaqClient, job_id: str) -> None:
         """Constructs a `Job`.
 
         Users should not call this themselves. If you only know the `job_id`, use `fetch_jobs`
@@ -77,6 +78,7 @@ class Job:
 
     def _refresh_job(self) -> None:
         """If the last fetched job is not terminal, gets the job from the API."""
+
         jobs_to_fetch: list[str] = []
 
         for job_id in self._job_id.split(","):
@@ -84,7 +86,7 @@ class Job:
                 jobs_to_fetch.append(job_id)
 
         if jobs_to_fetch:
-            result = self._client.fetch_jobs(jobs_to_fetch)
+            result = self._client.fetch_jobs(jobs_to_fetch, _models.CircuitType.CIRQ)
             self._job.update(result)
 
         self._update_status_queue_info()
@@ -99,6 +101,7 @@ class Job:
             jobs are still running (even if some are done), we report 'Running' as the
             overall status of the entire batch.
         """
+
         job_id_list = self._job_id.split(",")  # Separate aggregated job ids
 
         status_occurrence = {self._job[job_id]["status"] for job_id in job_id_list}
@@ -469,7 +472,7 @@ class Job:
     def __repr__(self) -> str:
         return f"css.Job(client={self._client!r}, job_id={self.job_id()!r})"
 
-    def _value_equality_values_(self) -> tuple[str, dict[str, Any]]:
+    def _value_equality_values_(self) -> tuple[str, dict[str, Any] | None]:
         return self._job_id, self._job
 
     def __getitem__(self, index: int) -> css.Job:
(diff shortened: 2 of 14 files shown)


cirq-superstaq/cirq_superstaq/daily_integration_test.py
```diff
@@ -301,6 +305,8 @@ def test_job(service: css.Service) -> None:
     multi_job = service.create_job(
         [circuit, circuit_alt], target="ibmq_brisbane_qpu", repetitions=10, method="dry-run"
     )
+    job = cast("Job", job)
+    multi_job = cast("Job", multi_job)
```
dowusu-antwi: Why is this needed? Once we officially migrate, will `service.create_job` return objects of this type by default?

cirq-superstaq/cirq_superstaq/job_test.py
```diff
@@ -148,6 +235,26 @@ def test_job_fields(job: css.Job, job_dict: dict[str, object]) -> None:
...
+    job_result = modifiy_job_result(
+        job_dictV3,
+        compiled_circuits=[css.serialize_circuits(circuit)],
+    )
+
+    mock_get.return_value.json.return_value = {str(uuid.UUID(int=42)): job_result}
+
+    assert jobV3.job_id() == uuid.UUID(int=42)
+    assert jobV3.target() == "ss_unconstrained_simulator"
+    assert jobV3.num_qubits(index=0) == 2
```
dowusu-antwi, line 252: If we are reusing the `jobV3` fixture which returns a two-qubit Bell state circuit, why do we use a new one-qubit circuit here? (If the compiled circuit being completely different is the intent, then I'm not sure what this is testing.)

cirq-superstaq/cirq_superstaq/job_test.py
```diff
@@ -158,6 +265,16 @@ def test_target(job: css.Job, job_dict: dict[str, object]) -> None:
...
 
 
+@mock.patch("requests.Session.get")
+def test_targetV3(mock_get: mock.MagicMock, jobV3: css.Job, job_dictV3: dict[str, object]) -> None:
+    mock_get.return_value.json.return_value = {str(uuid.UUID(int=42)): job_dictV3}
+    assert jobV3.target() == "ss_unconstrained_simulator"
+    mock_get.assert_called_once()
+
+    # Shouldn't need to retrieve anything now that `job._job_data` is populated:
+    assert jobV3.target() == "ss_unconstrained_simulator"
```
dowusu-antwi, line 275: What is this testing? (I see this pattern in other tests as well.)

cirq-superstaq/cirq_superstaq/job_test.py
```diff
@@ -391,13 +669,29 @@ def test_job_str_repr_eq(job: css.Job) -> None:
     assert not job == 1
 
 
+def test_job_str_repr_eqV3(jobV3: css.JobV3) -> None:
+    assert str(jobV3) == f"Job with job_id={uuid.UUID(int=42)}"
+    setup_code = "import cirq_superstaq as css\nimport general_superstaq as gss\n"
+    setup_code += "from uuid import UUID"
+    cirq.testing.assert_equivalent_repr(jobV3, setup_code=setup_code)
+    assert not jobV3 == 1
```
dowusu-antwi: Is this `assert` testing whether `cirq.testing.assert_equivalent_repr` passes? If so, could those two lines be combined into something like `assert cirq.testing.assert_equivalent_repr`?

dowusu-antwi (commented):
Some comments on the first couple of test files:

cirq-superstaq/cirq_superstaq/job_test.py
vietphamngoc: The `jobV3` fixture returns an empty job (the `job_data` attribute is not populated yet). This test checks whether the different methods return the right content corresponding to the related field in `job_data`. In that sense, the contents of the fields don't really have to be consistent with one another. The one qubit circuit was simply used for convenience but I can change that.

cirq-superstaq/cirq_superstaq/job_test.py
vietphamngoc: Once `_job_data` is populated, a call to `.target()` or any other similar methods shouldn't fetch the information from the server anymore. The first call to `.target()` should trigger a `get` request and the second should fetch from `_job_data`. The unique `get` request is tested explicitly with `mock_get.assert_called_once()` and implicitly with the second call to `.target()` as it should fail (I think) if it were to make a request (absence of side_effect).

cirq-superstaq/cirq_superstaq/job_test.py
vietphamngoc: The first `testing_equivalent_repr` will test the `__repr__` method while the `assert` statement will test the `_value_equality_values` method.

cirq-superstaq/cirq_superstaq/daily_integration_test.py
vietphamngoc: At the moment `service.create_job` either returns `Job` objects or `JobV3` objects. I have added these lines to remove the ambiguity as mypy can't seem to infere that a `"v0.2.0"` service will return `Job` objects. 
Once we migrate, the ambiguity should disappear as we replace the `Job` class with `JobV3` in which case these lines can be removed.
I agree that it is not ideal so I'm open to suggestions as to how we can do without casting.

general-superstaq/general_superstaq/superstaq_client_test.py
```diff
@@ -652,7 +652,7 @@ def test_superstaq_client_fetch_jobs(
         assert response == expected_result
 
         mock_get.assert_called_with(
-            f"http://example.com/{api_version}/client/job?job_id={uuid.UUID(int=0)}",
+            f"http://example.com/{api_version}/client/job/cirq?job_id={uuid.UUID(int=0)}",
```
dowusu-antwi, line 686: Why are we adding `cirq` to the routes? These are for job submission, right?

general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -1486,20 +1486,27 @@ def compile(self, json_dict: dict[str, str]) -> dict[str, str]:
...
-
         if circuit_type == _models.CircuitType.CIRQ:
+            compile_dict = {
+                "initial_logical_to_physicals": json.dumps(
+                    job_data["initial_logical_to_physicals"]
+                ),
+                "final_logical_to_physicals": json.dumps(job_data["final_logical_to_physicals"]),
+            }
+            compile_dict["physical_qubits"] = "[" + ", ".join(job_data["physical_qubits"]) + "]"
+            compile_dict["logical_qubits"] = "[" + ", ".join(job_data["logical_qubits"]) + "]"
```
vietphamngoc:
```suggestion
            init_log_to_phys = []
            final_log_to_phys = []
            for i in range(len(job_data["initial_logical_to_physicals"])):
                logical = json.loads(job_data["logical_qubits"][i])
                physical = json.loads(job_data["physical_qubits"][i])

                init_log_to_phys.append({json.dumps(logical[k]): json.dumps(physical[v]) for k, v in job_data["initial_logical_to_physicals"][i].items()})
                final_log_to_phys.append({json.dumps(logical[k]): json.dumps(physical[v]) for k, v in job_data["final_logical_to_physicals"][i].items()})

            compile_dict = {
                "initial_logical_to_physicals": "["
                + ", ".join(json.dumps(dico) for dico in init_log_to_phys)
                + "]",
                "final_logical_to_physicals": "["
                + ", ".join(json.dumps(dico) for dico in final_log_to_phys)
                + "]",
            }
```
This way we shouldn't need to change css.compiler_output

cirq-superstaq/cirq_superstaq/job.py
```diff
@@ -489,6 +492,400 @@ def __getitem__(self, index: int) -> css.Job:
...
+            _models.CircuitStatus.COMPILING,
+            _models.CircuitStatus.SIMULATING,
+            _models.CircuitStatus.PENDING,
+            _models.CircuitStatus.RUNNING,
+            _models.CircuitStatus.FAILED,
+            _models.CircuitStatus.CANCELLED,
+            _models.CircuitStatus.UNRECOGNIZED,
+            _models.CircuitStatus.COMPLETED,
+            _models.CircuitStatus.DELETED,
+        )
```
dowusu-antwi: Nit: Can this be made a class constant so we're not recreating it every status update?

cirq-superstaq/cirq_superstaq/job.py
```diff
@@ -489,6 +492,400 @@ def __getitem__(self, index: int) -> css.Job:
...
+        Users should not call this themselves. If you only know the `job_id`, use `fetch_jobs`
+        on `css.Service`.
+
+        Args:
+            client: The client used for calling the API.
+            job_id: Unique identifier for the job.
+        """
+        self._client = client
+        if not isinstance(self._client, _SuperstaqClientV3):
+            raise ValueError("JobV3 job can only be used with v0.3.0 of the Superstaq API.")
```
dowusu-antwi, line 535: Is this needed given the type annotation for `client` in the initializer parameters?

cirq-superstaq/cirq_superstaq/job.py
```diff
@@ -489,6 +492,400 @@ def __getitem__(self, index: int) -> css.Job:
...
+
+        Args:
+            index: An optional index of the specific sub-job to cancel.
+            kwargs: Extra options needed to fetch jobs.
+
+        Raises:
+            ~gss.SuperstaqServerException: If unable to get the status of the job
+                from the API or cancellations were unsuccessful.
+        """
+        self._client.cancel_jobs([self._job_id], **kwargs)
```
dowusu-antwi: Does the client check that the job isn't in a terminal state or should that be done here? E.g., like in `def counts`. Also, the `index` parameter isn't otherwise used here.

cirq-superstaq/cirq_superstaq/job.py
```diff
@@ -489,6 +492,400 @@ def __getitem__(self, index: int) -> css.Job:
...
+
+    def repetitions(self) -> int:
+        """Gets the number of repetitions requested for this job.
+
+        Returns:
+            The number of repetitions for this job.
+        """
+        if not hasattr(self, "_job_data"):
+            self._refresh_job()
+        return self._job_data.shots[0]
```
dowusu-antwi: Is it possible to request different shot counts for different circuits in the same job? If so we should rethink how this is defined.

cirq-superstaq/cirq_superstaq/job.py
```diff
@@ -489,6 +492,400 @@ def __getitem__(self, index: int) -> css.Job:
...
+
+        for temp_status in status_priority_order:
+            if temp_status in status_occurrence:
+                self._overall_status = temp_status
+                return
+
+    def _check_if_unsuccessful(self, index: int | None = None) -> None:
+        """Helper method to check if the current circuit status has any failure.
+
+        TODO: Separate out the different circuits within the job.
```
dowusu-antwi: It looks like this is done? AFAICT it produces adds an error message for each failed circuit, so the TODO can be removed if that's what it refer to.

dowusu-antwi (commented):
A few other comments:

cirq-superstaq/cirq_superstaq/daily_integration_test.py
richrines1: in general i prefer e.g. `assert isinstance(job, css.Job)` to using `cast`. mypy will treat them the same way, but the assertion will also generate a runtime error if our assumption is wrong

qiskit-superstaq/qiskit_superstaq/superstaq_backend.py
```diff
@@ -171,14 +173,16 @@ def run(
...
+            # Make a virtual job_id that aggregates all of the individual jobs
+            # into a single one that comma-separates the individual jobs.
+            job_id = ",".join(result["job_ids"])
+            return qss.SuperstaqJob(self, job_id)
+        else:
+            job_id_v3 = cast("uuid.UUID", result["job_id"])
+            return qss.SuperstaqJobV3(self, job_id_v3)
 
-    def retrieve_job(self, job_id: str) -> qss.SuperstaqJob:
+    def retrieve_job(self, job_id: str | uuid.UUID) -> qss.SuperstaqJob | qss.SuperstaqJobV3:
```
richrines1, line 196: this method was deprecated a while ago - do we need it to support JobV3?

cirq-superstaq/cirq_superstaq/job.py
vietphamngoc: Not really needed given that a job is created by `css.Service` which creates a `JobV3` only for`_SuperstaqClientV3` . It was just to add a double check but can be removed.

cirq-superstaq/cirq_superstaq/job.py
vietphamngoc: That's a good point, I changed it to a class constant.

cirq-superstaq/cirq_superstaq/job.py
vietphamngoc: At the moment, we can only run the circuits in a same job with the same number of shots.

general-superstaq/general_superstaq/superstaq_client_test.py
vietphamngoc: This allows to retrieve any circuits, regardless of the type they have been submitted with. If a circuit has been submitted with qiskit, it will be converted to a cirq circuit.

cirq-superstaq/cirq_superstaq/job.py
vietphamngoc: You are right, removed the TODO

qiskit-superstaq/qiskit_superstaq/superstaq_backend.py
vietphamngoc: Not really, but I thought that as long as it has not been removed, might as well support `JobV3`. I can remove the `jobV3` support if it doesn't make sense or remove the method altogether.

cirq-superstaq/cirq_superstaq/job.py
vietphamngoc: The check is done server side. You are right, I removed the `index` argument.

cirq-superstaq/cirq_superstaq/job.py
dowusu-antwi: Got it. I don't know if we would ever need this capability, but we can discuss at dev meeting.

cirq-superstaq/cirq_superstaq/daily_integration_test.py
vietphamngoc: Thanks I'll change to your suggestion.

general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -219,7 +222,7 @@ def fetch_jobs(
         self,
         job_ids: Sequence[str] | Sequence[uuid.UUID],
         **kwargs: object,
-    ) -> dict[str, dict[str, str]]:
+    ) -> dict[str, dict[str, Any]]:
```
dowusu-antwi, line 237: I think we've wanted to get rid of `Any`s in the past so that we always have good type checking. @richrines1 is this still true? We should be able to use `Union` of whatever type options we expect here. Unless we won't be sure precisely what types to expect until job refactor is complete? (I thought server-side decisions had been all but finalized.)

Also, see elsewhere in this file (if we end up replacing `Any`).

cirq-superstaq/cirq_superstaq/job.py
dowusu-antwi: This was sitting in my drafts, but from last week's dev discussion: supporting different shot counts would be a nice to have but not necessary for this PR.

cirq-superstaq/cirq_superstaq/compiler_output.py
```diff
@@ -170,6 +170,7 @@ def read_json(json_dict: dict[str, Any], circuits_is_list: bool) -> CompilerOutp
     final_logical_to_physicals: list[dict[cirq.Qid, cirq.Qid]] = list(
         map(dict, cirq.read_json(json_text=json_dict["final_logical_to_physicals"]))
     )
+
```
dowusu-antwi, line 173: nit: Why was this added? Is it to pass formatting checks?

cirq-superstaq/cirq_superstaq/daily_integration_test.py
dowusu-antwi: Has this been updated? I still see the `cast` for both what's been pushed.

cirq-superstaq/cirq_superstaq/job_test.py
dowusu-antwi: Got it, no need to change it then.

cirq-superstaq/cirq_superstaq/job_test.py
dowusu-antwi: I think the `mock_get.assert_called_once()` would need to be moved to after line 275. If I'm understanding `MagicMock` correctly, then this would still pass even if it's called again in line 275 via `jobv3.target()`.

Alternatively, we could have something like:
```python
assert jobV3.target() == "ss_unconstrained_simulator"
mock_get.call_count == 1
assert jobV3.target() == "ss_unconstrained_simulator"
mock_get.call_count == 1
```

cirq-superstaq/cirq_superstaq/job.py
```diff
@@ -489,6 +492,399 @@ def __getitem__(self, index: int) -> css.Job:
...
+        _models.CircuitStatus.PENDING,
+        _models.CircuitStatus.RUNNING,
+        _models.CircuitStatus.FAILED,
+        _models.CircuitStatus.CANCELLED,
+        _models.CircuitStatus.UNRECOGNIZED,
+        _models.CircuitStatus.COMPLETED,
+        _models.CircuitStatus.DELETED,
+    )
+
+    def __init__(self, client: gss.superstaq_client._SuperstaqClientV3, job_id: uuid.UUID) -> None:
```
richrines1: can we make this more flexible, e.g.
```suggestion
    def __init__(self, client: gss.superstaq_client._SuperstaqClientV3, job_id: str | uuid.UUID) -> None:
```
and use something like
```python
self._job_id = job_id if isinstance(job_id, uuid.UUID) else uuid.UUID(job_id)
```
below?

cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -390,7 +393,13 @@ def get_job(self, job_id: str) -> css.job.Job:
         Raises:
             ~gss.SuperstaqServerException: If there was an error accessing the API.
         """
-        return css.job.Job(client=self._client, job_id=job_id)
+        if isinstance(self._client, _SuperstaqClient):
+            job_id = cast("str", job_id)
+            return css.Job(client=self._client, job_id=job_id)
+        else:
+            # mypy doesn't narrow hence the ignore
+            job_id = cast("uuid.UUID", job_id)
```
richrines1: assuming prior suggestion (re: allowing `JobV3` to accept strings), this shouldn't be necessary
```suggestion
```

qiskit-superstaq/qiskit_superstaq/superstaq_backend.py
```diff
@@ -198,7 +202,12 @@ def retrieve_job(self, job_id: str) -> qss.SuperstaqJob:
             DeprecationWarning,
             stacklevel=2,
         )
-        return qss.SuperstaqJob(self, job_id)
+        if isinstance(self._provider._client, _SuperstaqClient):
+            job_id = cast("str", job_id)
+            return qss.SuperstaqJob(self, job_id)
+        else:
+            job_id = cast("uuid.UUID", job_id)
+            return qss.SuperstaqJobV3(self, job_id)
```
richrines1:
```suggestion
            return qss.SuperstaqJob(self, str(job_id))
        else:
            return qss.SuperstaqJobV3(self, job_id)
```

cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -390,7 +393,13 @@ def get_job(self, job_id: str) -> css.job.Job:
         Raises:
             ~gss.SuperstaqServerException: If there was an error accessing the API.
         """
-        return css.job.Job(client=self._client, job_id=job_id)
+        if isinstance(self._client, _SuperstaqClient):
+            job_id = cast("str", job_id)
+            return css.Job(client=self._client, job_id=job_id)
```
richrines1:
```suggestion
            return css.Job(client=self._client, job_id=str(job_id))
```

general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -1481,23 +1486,57 @@ def compile(self, json_dict: dict[str, str]) -> dict[str, str]:
...
+                + ", ".join(json.dumps(q_map) for q_map in qiskit_initial_log_to_phys)
+                + "]",
+                "final_logical_to_physicals": "["
+                + ", ".join(json.dumps(q_map) for q_map in qiskit_final_log_to_phys)
+                + "]",
+            }
             compile_dict["qiskit_circuits"] = "[" + ", ".join(job_data["compiled_circuits"]) + "]"
             return compile_dict
 
+        elif circuit_type == _models.CircuitType.CIRQ:
```
richrines1: this doesn't seem to be working correctly - `compiler_output.final_logical_to_physicals` is giving me e.g.
```
{
    '{"cirq_type": "GridQubit", "row": 0, "col": 0}': '{"cirq_type": "GridQubit", "row": 0, "col": 0}',
    '{"cirq_type": "GridQubit", "row": 1, "col": 0}': '{"cirq_type": "GridQubit", "row": 1, "col": 0}',
    ...
}
```
when it should be a dictionary of cirq.Qids

ideally this conversion shouldn't be happening here - `_SuperstaqClient` should just return raw data objects, and the language-specific caller (e.g. `css.Service` or `qss.SuperstaqBackend`) should handle the conversion to the correct types

cirq-superstaq/cirq_superstaq/job_test.py
```diff
@@ -170,6 +287,19 @@ def test_num_qubits(job: css.Job, job_dict: dict[str, object]) -> None:
     assert job.num_qubits() == [2]
 
 
+@mock.patch("requests.Session.get")
+def test_num_qubitsV3(
+    mock_get: mock.MagicMock, jobV3: css.Job, job_dictV3: dict[str, object]
+) -> None:
+    mock_get.return_value.json.return_value = {str(uuid.UUID(int=42)): job_dictV3}
+    assert jobV3.num_qubits(index=0) == 2
+    mock_get.assert_called_once()
```
dowusu-antwi: See thread from [above comment](https://github.com/Infleqtion/client-superstaq/pull/1215/files/0d4dd9030f3bc9c8978f93bb9c87379941549cad#r2150696803).

cirq-superstaq/cirq_superstaq/job_test.py
```diff
@@ -180,6 +310,18 @@ def test_repetitions(job: css.Job, job_dict: dict[str, object]) -> None:
     assert job.repetitions() == 1
 
 
+@mock.patch("requests.Session.get")
+def test_repetitionsV3(
+    mock_get: mock.MagicMock, jobV3: css.Job, job_dictV3: dict[str, object]
+) -> None:
+    mock_get.return_value.json.return_value = {str(uuid.UUID(int=42)): job_dictV3}
+    assert jobV3.repetitions() == 1
+    mock_get.assert_called_once()
```
dowusu-antwi: See [above comment](https://github.com/Infleqtion/client-superstaq/pull/1215/files/0d4dd9030f3bc9c8978f93bb9c87379941549cad#r2150696803).

cirq-superstaq/cirq_superstaq/job.py
dowusu-antwi: Has this change been pushed?

cirq-superstaq/cirq_superstaq/job.py
dowusu-antwi: Sounds good -- we can leave it until we officially deprecate `v0.2.0`.

cirq-superstaq/cirq_superstaq/job.py
dowusu-antwi: I also see the `index` parameter in the code -- it's possible that I'm not seeing the most up-to-date 'Files Changed'.

dowusu-antwi (commented):
This looks about ready to go to me. I just had a few comments and follow-ups, and after also responding to Rich's comments we can merge.

cirq-superstaq/cirq_superstaq/job.py
vietphamngoc: It has, but weirdly enough it doesn't show here...

cirq-superstaq/cirq_superstaq/job.py
vietphamngoc: It has been removed in the last push but doesn't show here.

cirq-superstaq/cirq_superstaq/job.py
vietphamngoc: that's a good idea.

cirq-superstaq/cirq_superstaq/daily_integration_test.py
vietphamngoc: I forgot some in the code. Should all be removed now.

cirq-superstaq/cirq_superstaq/job_test.py
vietphamngoc: That's a good idea, makes the test clearer.

cirq-superstaq/cirq_superstaq/service.py
vietphamngoc: Done

cirq-superstaq/cirq_superstaq/compiler_output.py
vietphamngoc: I'm not sure I understand the question here.

general-superstaq/general_superstaq/superstaq_client.py
vietphamngoc: It should be fixed now, returning dictionary of `cirq.Qids`. 
The reason we are doing this "processing" work here is to maintain compatibility between v2 and v3 for `CompilerOutput` but I can move this to the relevant place in `qss` and `css`

cirq-superstaq/cirq_superstaq/job.py
dowusu-antwi: @richrines1 do you know what's going on here? I'm not sure why updates aren't showing up here:
https://github.com/Infleqtion/client-superstaq/blob/2d4ee0282428ee95a5710eafa7a0a6c7522bc9f1/cirq-superstaq/cirq_superstaq/job.py#L186

cirq-superstaq/cirq_superstaq/compiler_output.py
dowusu-antwi: I'm referring to the newline on 173 -- there are sometimes weird newline issues that mess with format checks, although I guess that's typically happened at the end of files. So we always try to minimize the introduction of newlines where it isn't necessary.

vietphamngoc: I see, sorry. No I just find the code more readable with this newline but happy to remove it.

dowusu-antwi: Ah, fair point -- that's also always something to keep in mind. To be fair I'm being a bit nitpicky. FWIW we'll probably remove the following line `pulse_gate_circuits = None` at some point anyway, since IBM doesn't support pulse calibrations anymore.

cirq-superstaq/cirq_superstaq/job.py
```diff
@@ -489,6 +492,403 @@ def __getitem__(self, index: int) -> css.Job:
...
+    def _check_if_unsuccessful(self, index: int | None = None) -> None:
+        """Helper method to check if the current circuit status has any failure.
+
+        Args:
+            index: The index of the specific job status.
+
+        Raises:
+            gss.SuperstaqUnsuccessfulJobException: If a failure status is found in the job.
+        """
+        status = self.status()
```
richrines1:
```suggestion
        status = self.status(index)
```
(if `index` is not None, we should only throw an error if the corresponding task failed)

vietphamngoc: Done

general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -1487,23 +1492,62 @@ def compile(self, json_dict: dict[str, str]) -> dict[str, str]:
...
+                for q_map in job_data["final_logical_to_physicals"]
+            ]
+
+            compile_dict = {
+                "initial_logical_to_physicals": "["
+                + ", ".join(json.dumps(q_map) for q_map in qiskit_initial_log_to_phys)
+                + "]",
+                "final_logical_to_physicals": "["
+                + ", ".join(json.dumps(q_map) for q_map in qiskit_final_log_to_phys)
+                + "]",
```
richrines1: does this do the same thing?
```suggestion
                "initial_logical_to_physicals": json.dumps(qiskit_initial_log_to_phys),
                "final_logical_to_physicals": json.dumps(qiskit_final_log_to_phys),
```

cirq-superstaq/cirq_superstaq/job.py
```diff
@@ -489,6 +491,402 @@ def __getitem__(self, index: int) -> css.Job:
...
+                    _get_marginal_counts(counts, qubit_indices) for counts in counts_list
+                ]
+            return counts_list
+
+        gss.validation.validate_integer_param(index, min_val=0)
+        if self._job_data.counts[index] is None:
+            raise RuntimeError(f"Circuit {index} of job {self._job_id} does not have any counts.")
+        single_counts: dict[str, int] = self._job_data.counts[index]  # type: ignore[assignment]
+        # Type checking does not recognise that the above error catches the case when the counts
+        # are None.
```
richrines1: does this work without the pragma?
```suggestion
        single_counts = self._job_data.counts[index]
        if single_counts is None:
            raise RuntimeError(f"Circuit {index} of job {self._job_id} does not have any counts.")
```

cirq-superstaq/cirq_superstaq/job.py
```diff
@@ -489,6 +491,402 @@ def __getitem__(self, index: int) -> css.Job:
...
+
+        Args:
+            client: The client used for calling the API.
+            job_id: Unique identifier for the job.
+        """
+        self._client = client
+        if not isinstance(self._client, _SuperstaqClientV3):
+            raise ValueError("JobV3 job can only be used with v0.3.0 of the Superstaq API.")
+        self._overall_status = _models.CircuitStatus.RECEIVED
+        self._job_data: _models.JobData
```
richrines1: can we make this 
```suggestion
        self._job_data: _models.JobData | None = None
```
and then use `if self._job_data is None` instead of all the hasattrs below?

general-superstaq/general_superstaq/superstaq_client.py
richrines1: yeah ideally we should avoid Any when possible (with the possible exception of json content that's just bound for pydantic validation anyway). A more specific annotation is definitely nice, but in most cases like this we can also just use `object` instead (which will force us to validate types as needed downstream - in this case i think this would just require updating the `if` statement in `qss.SuperstaqProvider.get_job()` to something like
```python
        if isinstance(target, str) and all(target == val["target"] for val in jobs.values()):
```
)

general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -1487,23 +1492,62 @@ def compile(self, json_dict: dict[str, str]) -> dict[str, str]:
...
+            }
             compile_dict["qiskit_circuits"] = "[" + ", ".join(job_data["compiled_circuits"]) + "]"
```
richrines1: not sure where the issue is, but for some reason `qss.SuperstaqBackend.compile().circuits` only gives me a single circuit when i compile multiple?

general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -1487,23 +1492,62 @@ def compile(self, json_dict: dict[str, str]) -> dict[str, str]:
...
+
+            compile_dict = {
+                "initial_logical_to_physicals": "["
+                + ", ".join(q_map for q_map in cirq_initial_log_to_phys)
+                + "]",
+                "final_logical_to_physicals": "["
+                + ", ".join(q_map for q_map in cirq_final_log_to_phys)
+                + "]",
+            }
+            compile_dict["cirq_circuits"] = "[" + ", ".join(job_data["compiled_circuits"]) + "]"
```
richrines1: i think this can be simplified a bit to avoid manually editing json strings, e.g.:
```suggestion
                cirq_initial_log_to_phys.append(
                    [
                        [logical[k], physical[v]]
                        for k, v in job_data["initial_logical_to_physicals"][i].items()
                    ]
                )
                cirq_final_log_to_phys.append(
                    [
                        [logical[k], physical[v]]
                        for k, v in job_data["final_logical_to_physicals"][i].items()
                    ]
                )

            compile_dict = {
                "initial_logical_to_physicals": json.dumps(cirq_initial_log_to_phys),
                "final_logical_to_physicals": json.dumps(cirq_final_log_to_phys),
                "cirq_circuits": json.dumps(list(map(json.loads, job_data["compiled_circuits"]))),
            }
```

general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -1487,23 +1492,62 @@ def compile(self, json_dict: dict[str, str]) -> dict[str, str]:
         # Join circuits together in json string - TODO: make this neater.
         # Note mypy does not recognize that the above checks ensure there are no None's anywhere,
         # hence the ignored [arg-type]'s
```
richrines1: is this comment relevant anymore?

cirq-superstaq/cirq_superstaq/job.py
```diff
@@ -489,6 +491,402 @@ def __getitem__(self, index: int) -> css.Job:
...
+
+        Returns:
+            A single compiled circuit or list of compiled circuits.
+        """
+        if not hasattr(self, "_job_data"):
+            self._refresh_job()
+        if all(c is None for c in self._job_data.compiled_circuits):
+            raise RuntimeError(f"The job {self._job_id} has no compiled circuits.")
+
+        if any(c is None for c in self._job_data.compiled_circuits):
```
richrines1: if an index is passed in we should only throw an error if that circuit is missing

(ditto in qss)

cirq-superstaq/cirq_superstaq/job_test.py
```diff
@@ -132,6 +211,14 @@ def test_cancel(job: css.Job, multi_circuit_job: css.Job) -> None:
         assert mock_post.call_count == 3
 
 
+@mock.patch("requests.Session.put")
```
richrines1, line 214: love all the mocking of requests themselves instead of local code :)

qiskit-superstaq/qiskit_superstaq/superstaq_job.py
```diff
@@ -11,11 +11,14 @@
...
 from __future__ import annotations
 
+import time
+import uuid
 from collections.abc import Sequence
 from typing import Any, overload
 
 import general_superstaq as gss
 import qiskit
+from general_superstaq import _models
```
richrines1: now that it's used in main code let's import models in `gss.__init__` and use `gss._models` everywhere

cirq-superstaq/cirq_superstaq/job.py
```diff
@@ -489,6 +491,402 @@ def __getitem__(self, index: int) -> css.Job:
...
+                )
+            time.sleep(polling_seconds)
+            time_waited_seconds += polling_seconds
+
+        # Check to see if unsuccessful
+        self._check_if_unsuccessful(index)
+
+        if index is None:
+            if any(c is None for c in self._job_data.counts):
+                raise RuntimeError(f"Job {self._job_id} does not have counts for all circuits.")
```
richrines1: i don't think `RuntimeError` is the right exception class for these - maybe `SuperstaqException`? (ditto elsewhere)

general-superstaq/general_superstaq/superstaq_client.py
richrines1: makes sense, i think this is fine for now

general-superstaq/general_superstaq/superstaq_client.py
vietphamngoc: It does, thanks!

cirq-superstaq/cirq_superstaq/job.py
vietphamngoc: Seems like it's working

cirq-superstaq/cirq_superstaq/job.py
vietphamngoc: We can do that, but then we get a lot of situations where mypy doesn't get that `self._job_data` is not `None` anymore. I'll see how to fix this

general-superstaq/general_superstaq/superstaq_client.py
vietphamngoc: No, removed it

cirq-superstaq/cirq_superstaq/job.py
vietphamngoc: Good point!

cirq-superstaq/cirq_superstaq/job.py
vietphamngoc: That's a good point. Updated.

qiskit-superstaq/qiskit_superstaq/superstaq_job.py
vietphamngoc: Done

general-superstaq/general_superstaq/superstaq_client.py
vietphamngoc: It had to do with the deserialization of the returned circuits. Put in a fix.

general-superstaq/general_superstaq/superstaq_client.py
vietphamngoc: Nice suggestion! Thanks!

vietphamngoc:
The switch from `Any` to `object` might take some work for the type checker to be happy. Given that we'd like to have this ready for Friday next week, can we merge this branch and I'll create another PR for the switch to `object`?

vtomole:
@richrines1  to review before Friday.

general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -1476,33 +1497,67 @@ def compile(self, json_dict: dict[str, str]) -> dict[str, str]:
...
+            ]
+            qiskit_final_log_to_phys = [
+                [[k, v] for k, v in q_map.items()] for q_map in final_logical_to_physicals
+            ]
+
+            compile_dict = {
+                "initial_logical_to_physicals": json.dumps(qiskit_initial_log_to_phys),
+                "final_logical_to_physicals": json.dumps(qiskit_final_log_to_phys),
+            }
+            compile_dict["qiskit_circuits"] = "[" + ", ".join(compiled_circuits) + "]"
```
richrines1: it would be nice to get rid of the rest of the manual json string editing. i believe this should work after subsequent suggestions:
```suggestion
            "qiskit_circuits": json.dumps(compiled_circuits),
        }
```

qiskit-superstaq/qiskit_superstaq/compiler_output.py
```diff
@@ -184,19 +184,28 @@ def __eq__(self, other: object) -> bool:
...
 
     Returns:
         A `CompilerOutput` object with the compiled circuit(s) and (if applicable to this target)
         corresponding pulse gate circuit(s).
     """
-    compiled_circuits = qss.serialization.deserialize_circuits(json_dict["qiskit_circuits"])
+    if api_version == "v0.2.0":
+        compiled_circuits = qss.serialization.deserialize_circuits(json_dict["qiskit_circuits"])
+    else:
+        serialized_circuits = json_dict["qiskit_circuits"][1:-1].split(", ")
```
richrines1: (assuming prior suggestion)
```suggestion
        serialized_circuits = json.loads(json_dict["qiskit_circuits"])
```

qiskit-superstaq/qiskit_superstaq/compiler_output_test.py
```diff
@@ -196,6 +196,40 @@ def test_read_json_empty_circuit() -> None:
     assert out.pulse_gate_circuits[1].op_start_times == []
 
 
+def test_read_jsonV3() -> None:
+    qc = qiskit.QuantumCircuit(2)
+    qc.h(0)
+    qc.cx(0, 1)
+
+    json_dict = {
+        "qiskit_circuits": "[" + qss.serialization.serialize_circuits(qc) + "]",
```
richrines1: (assuming prior)
```suggestion
        "qiskit_circuits": json.dumps([qss.serialization.serialize_circuits(qc)]),
```

qiskit-superstaq/qiskit_superstaq/compiler_output_test.py
```diff
@@ -196,6 +196,40 @@ def test_read_json_empty_circuit() -> None:
...
+    assert out.final_logical_to_physical == {0: 0, 1: 1}
+
+    json_dict = {
+        "qiskit_circuits": (
+            "["
+            + qss.serialization.serialize_circuits(qc)
+            + ", "
+            + qss.serialization.serialize_circuits(qc)
+            + "]"
+        ),
```
richrines1: (assuming prior)
```suggestion
        "qiskit_circuits": json.dumps(2 * [qss.serialize_circuits(qc)]),
```

general-superstaq/general_superstaq/superstaq_client_test.py
```diff
@@ -1368,7 +1385,7 @@ def test_superstaq_client_compile_v3(
             "target": "ss_example_qpu",
             "provider_id": ["ss"],
             "num_circuits": 1,
-            "compiled_circuits": ["compiled world"],
+            "compiled_circuits": ['"compiled world"'],
```
richrines1, line 1389: to get this to pass for both cirq and qiskit we'd need the serialized value to change i think. alternatively this should work for both though:
```suggestion
            "compiled_circuits": [],
```

general-superstaq/general_superstaq/superstaq_client_test.py
```diff
@@ -1409,14 +1426,14 @@ def test_superstaq_client_compile_v3(
...
-        f"http://example.com/{client_v3.api_version}/client/job?job_id={job_id}",
+        f"http://example.com/{client_v3.api_version}/client/job/cirq?job_id={job_id}",
         headers=EXPECTED_HEADERS[client_v3.api_version],
         verify=False,
     )
     assert compilation_results == {
-        f"{circuit_type}_circuits": "[compiled world]",
-        "initial_logical_to_physicals": "[{0: 0}]",
-        "final_logical_to_physicals": "[{0: 0}]",
+        f"{circuit_type}_circuits": '["compiled world"]',
```
richrines1: (assuming prior)
```suggestion
        f"{circuit_type}_circuits": "[]",
```

richrines1 (commented):
mostly looks great! only remaining concern is the handling of serialized compiled circuits for qiskit, which i think can be corrected via the following 6 suggested edits:

qiskit-superstaq/qiskit_superstaq/compiler_output.py
vietphamngoc: Interesting, I remember we tried using `json.dumps` and `json.loads` but it didn't behave as expected. Seems to be working now. Thanks!

PR #1217, opened 2025-06-12
add offset to XEB fit plot

richrines1:
currently we drop the fit's y-intercept from the fidelity decay plot, resulting in a fit that doesn't align with the data - for example:
image

after correction, the same plot looks much more reasonable:
image


vtomole (approved):
LGTM

PR #1219, opened 2025-06-13
add reference for experimental characterization in ACES tutorial

KerryZhou1024:


vtomole (commented):
Actually can we just only reference the experimental paper? The experimental paper cites the Flammia paper in the abstract.

KerryZhou1024:
Should I include the original paper in the two notebooks or should I only reference the experimental one? So far I include both papers for the two notebooks but only the experimental one in rst file.

vtomole (approved):
LGTM

PR #1226, opened 2025-06-23
add phase_by magic method to `StrippedCZGate`

richrines1:


cirq-superstaq/cirq_superstaq/ops/qubit_gates_test.py
```diff
@@ -151,6 +151,9 @@ def test_stripped_cz_gate() -> None:
     assert gate**1 == gate
     assert gate**2 == css.ParallelGates(z_exp_gate, z_exp_gate)
 
+    assert cirq.phase_by(gate, 1.23, 0) == gate
```
pgoiporia, line 154: I'm a bit confused here. Does adding phase to the gate not matter? How are they equal?

richrines1: `cirq.phase_by(gate, phase, qid)` is supposed to \*conjugate\* `gate` by `cirq.Z(q)**phase`, which has no effect in this case because this gate commutes with Z (so `Z(q)**phase * stripped_cz * Z(q)**-phase == stripped_cz`)

(note this is also the same as what cirq [does in `CZPowGate`](https://github.com/quantumlib/Cirq/blob/7978a27d4579295343a462958d6dbfdbc0ae9401/cirq-core/cirq/ops/common_gates.py#L1072-L1073))

pgoiporia: Ah ok I see

pgoiporia (approved):
LGTM

cirq-superstaq/cirq_superstaq/ops/qubit_gates_test.py
pgoiporia: I guess I'm a bit confused why this magic method exists. Shouldn't it be fairly trivial to conjugate any gate?

richrines1: a few reasons i can think of:
* sometimes the returned gate type is different from the input - for example `phase_by(cirq.X, ...)` will return a `cirq.PhasedXPowGate`. This is very easy to support if gates can just implement their own `._phase_by_`
* this can also be much more efficient, especially for larger gates (otherwise we'd probably have to expand the gate's unitary every time and do some linear algebra to determine what the conjugated gate looks like, when it may be trivial to compute for a specific gate type)
* in some cases this probably also allows for better parametrized gate support. The logic to conjugate arbitrary parametrized gates would probably get ugly really fast, but for e.g. `CZ**sympy.Symbol("z")` we know know conjugation still has no effect

PR #1230, opened 2025-06-23
Add condition to check notebooks only on push or from the same repository in CI

KerryZhou1024:
So that external fork PRs will pass all checks.


pgoiporia:
You've tested that this works from a fork? If so then lgtm and I'll approve.

KerryZhou1024:
Np! Will return with test results.

vtomole (approved):
Approving cause a merge is needed to test.

PR #1242, opened 2025-07-01
Remove remaining `pylint` refs + config nits

bharat-thotakura:


supermarq-benchmarks/supermarq/qcvv/su2_test.py
```diff
@@ -131,13 +129,13 @@ def test_build_circuits_old(su2_experiment: SU2) -> None:  # pragma: no cover
...
                     1: ───PhXZ───│───X─────────────────│───X───│───X─────────────────│───PhXZ───│───X─────────────────│───X───│───X─────────────────│───PhXZ───│───X─────────────────│───X───│───X─────────────────│───PhXZ───M───
-                    """  # noqa: E501  # pylint: disable=line-too-long
+                    """  # noqa: E501
                 ),
             )
 
 
 def test_analyse_results(tmp_path: pathlib.Path, su2_experiment: SU2) -> None:
-    def decay(x):  # type: ignore[no-untyped-def] # pylint: disable=W9012 # noqa: ANN001, ANN202
+    def decay(x: int | float) -> float:
```
richrines1: nit
```suggestion
    def decay(x: float) -> float:
```
(int is [formally treated](https://peps.python.org/pep-0484/#the-numeric-tower) as a subclass if float for the sake of type checking)

.github/workflows/ci.yml
```diff
@@ -57,7 +57,7 @@ jobs:
           pip install -e ./checks-superstaq -e ./general-superstaq -e ./qiskit-superstaq -e ./cirq-superstaq -e ./supermarq-benchmarks
       - name: Ruff lint
         run: |
-          checks/lint_.py
+          FORCE_COLOR=1 checks/lint_.py
```
richrines1: would it make sense to do this in `checks_superstaq.lint_.py` itself? (ditto for format_)

richrines1 (approved):
lgtm % nit and a question

.github/workflows/ci.yml
bharat-thotakura: So there was no direct CLI flag I could immediately [find](https://github.com/astral-sh/ruff/issues/2743), so I settled for the env var. I only set it in the `.yaml` so that it's not global for all `checks-superstaq` users (even then, I could put it in our `checks/`). Regardless, imo it's a nice quality of life feature so I've added it in https://github.com/Infleqtion/client-superstaq/pull/1242/commits/95329dd6a39621ce7207d491a4d5b9e841060e97

Issue #1129, opened 2024-12-13
Flakey notebook tests


cdbf1:
### What is happening?

I am frequently getting flakey notebook checks at the moment. The error message is:
```
ERROR    traitlets:client.py:567 Error occurred while starting new kernel client for kernel ea20993a-d565-43b8-adeb-a76cc20b7d91: Kernel died before replying to kernel_info
```

When I rerun the task a minute or so later it works fine

### How can we reproduce the issue?

-

### What should happen?

-

### Environment

CI/CD pipeline

stephanielee9:
Related: https://github.com/Infleqtion/client-superstaq/issues/1104

PR #1244, opened 2025-07-02
Upper bound notebook `pytest` threads to 4

bharat-thotakura:
Various notebook check timeouts/failures, like in https://github.com/Infleqtion/client-superstaq/issues/1129, may potentially be attributed to spawning too many kernels in a short amount of time, trying to use a port already in use. This PR hence limits the number of threads to a maximum of four in an effort to reduce the likelihood of such scenarios from happening.

As of this PR, the number of workers currently spawned by each notebook check is:

- `cirq-superstaq`: 9 workers
- `qiskit-superstaq:` 8 workers
- `supermarq-benchmarks`: 6 workers


richrines1 (approved):
lgtm!

Issue #1222, opened 2025-06-18
Implement missing ` __hash__` for some superstaq classes


bharat-thotakura:
### What is happening?

`ruff` `v0.12` highlighted that some superstaq classes have `__eq__` but not `__hash__` via the [PLW1641](https://docs.astral.sh/ruff/rules/eq-without-hash/#eq-without-hash-plw1641) rule. This would be good to have as far as it is possible to straightforwardly implement and hence remove the `# noqa: PLW1641` comments we have after https://github.com/Infleqtion/client-superstaq/pull/1221

### What are the current alternatives?

Keep the `# noqa: PLW1641` comments or exclude the rule

### Any additional context?

Flagged with `ruff==0.12.0`

pgoiporia:
Is there a difference between the `hash` method and the `json` method we have?

bharat-thotakura:
If you're referring to `to_json`, then yes. That method would specify the `json` serialization into a `str`, but the [PLW1641](https://docs.astral.sh/ruff/rules/eq-without-hash/#eq-without-hash-plw1641) rule is raised when there's `__eq__` but not `__hash__` at the same time. The `__hash__` is used to define the hashability for the class I believe so that it can be used in data structures such as the keys of a dictionary (which acts like a hash table).

bharat-thotakura:
Completed by https://github.com/Infleqtion/client-superstaq/pull/1257

PR #1257, opened 2025-07-31
added hash functions for two classes, removed noqa

microbe73:


microbe73:
For #1222. The only other class with this noqa is the compiler_output class, but it has several dictionary attributes and a circuit attribute so I don't see a good way to hash it.

PR #1247, opened 2025-07-04
add verbatim support to client_v3_fixes

richrines1:
to be merged into #1215 (attn: @vietphamngoc) - needed for smoke tests

PR #1267, opened 2025-08-28
Version bump

cdbf1:


richrines1:
if possible it would be awesome if we could merge #1247 before deploying

PR #1269, opened 2025-09-11
Changing to remove barriers from gate count

vietphamngoc:
In the sample data, the 1q gate and 2q gate counts include the barriers. Excluding them helps keep a more accurate count of each type of gates.


supermarq-benchmarks/supermarq/qcvv/irb.py
```diff
@@ -745,10 +745,14 @@ def _build_circuits(self, num_circuits: int, cycle_depths: Iterable[int]) -> Seq
                         "clifford_depth": depth,
                         "circuit_depth": len(rb_circuit),
                         "single_qubit_gates": sum(
-                            1 for op in rb_circuit.all_operations() if len(op.qubits) == 1
+                            1
```
cdbf1: Could we factor this out into a private `_count_gates(self, circuit: cirq.Circuit, num_qubits: int) -> int` method that gets called in each place?

cdbf1 (changes requested):

supermarq-benchmarks/supermarq/qcvv/irb.py
vietphamngoc: Good point!

PR #1292, opened 2025-10-28
Stimcirq requirement

vtomole:
This requirement is needed to compile Stim circuits with Superstaq for example

```python
import stimcirq
from qldpc import circuits, codes
import cirq_superstaq as css

code = codes.SteaneCode()
circuit = circuits.get_memory_experiment(code, num_rounds=1)
circuit= stimcirq.stim_circuit_to_cirq_circuit(circuit)
service = css.Service()
compiler_output = service.cq_compile(circuit)

```


cirq-superstaq/cirq_superstaq/serialization.py
```diff
@@ -35,7 +36,7 @@ def deserialize_circuits(serialized_circuits: str) -> list[cirq.Circuit]:
     Returns:
         The circuit or list of circuits that was serialized.
     """
-    resolvers = [*SUPERSTAQ_RESOLVERS, *cirq.DEFAULT_RESOLVERS]
+    resolvers = [*cirq.DEFAULT_RESOLVERS, stimcirq.JSON_RESOLVER, *SUPERSTAQ_RESOLVERS]
```
richrines1: nit: i think there was a reason we put our resolvers first
```suggestion
    resolvers = [*SUPERSTAQ_RESOLVERS, *cirq.DEFAULT_RESOLVERS, stimcirq.JSON_RESOLVER]
```

bharat-thotakura:
To reference in the future, can we update the PR description to give some more context as to why `stimcirq` is being made an explicit dependency of `cirq-superstaq`? That is, for example, what feature is it enabling that we couldn't make an optional dependency?

bharat-thotakura (approved):
LGTM

(might be nice to add a small test to check that the deserialization works on a circuit with, say, a `MeasureAndOrResetGate`)

PR #1293, opened 2025-10-29
target model v2

richrines1:
more compact replacement for `TargetModel`

also renamed the new model `TargetDescription`, but open to other suggestions :)

### DIFF
--- a/general-superstaq/general_superstaq/models.py
+++ b/general-superstaq/general_superstaq/models.py
@@ -8,7 +8,7 @@ import itertools
 import uuid
 from collections.abc import Mapping, Sequence
 from enum import Enum
-from typing import Annotated, Any, Self
+from typing import Annotated, Any
 
 import pydantic.functional_validators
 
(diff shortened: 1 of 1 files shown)


cdbf1:
I'm very much in favour of switching to using statuses rather than all the booleans!

general-superstaq/general_superstaq/models.py
```diff
@@ -366,9 +366,33 @@ class TargetStatus(str, Enum):
...
 
-class TargetModel(DefaultPydanticModel):
+class TargetInputType(str, Enum):
+    """The input type supported by a Superstaq target."""
+
+    CIRCUIT = "circuit"
+    QUBO = "qubo"
+
+
+class TargetDescription(DefaultPydanticModel):
```
dowusu-antwi, line 376: How about `TargetSpecification` / `TargetSpecs`?

general-superstaq/general_superstaq/models_test.py
```diff
@@ -11,6 +11,65 @@
...
+
+    with pytest.raises(pydantic.ValidationError, match=r"valid target device type"):
+        _ = gss.models.TargetDescription(
+            target="sqale_example_system",
+            status="available",
+            supported_inputs=["circuit"],
+            accessible=True,
+        )
+
+    with pytest.raises(pydantic.ValidationError, match=r"valid string format"):
```
pgoiporia, line 64: Maybe this is overkill, given that we hit full coverage, but should we be adding the same test for `status`?

Issue #1243, opened 2025-07-02
retire `CompilerOutput`, use Jobs everywhere


richrines1:
### What is happening?

with the v0.3.0 api compilation is asynchronous, so e.g. `service.compile()` should just return a `css.Job` (which already has a `compiled_circuit` attribute) instead of a specialized compilation object

(note this might require a bit of work to make sure e.g. `job.compiled_circuits()` blocks until compilation is done, like `job.counts()` does until the job has terminated - but i think this is a change we'll have to make anyway? we also probably want to add some special handling in `job.counts()` for non-submission jobs)

PR #1301, opened 2025-11-13
support jaqal subcircuits

richrines1:
temporary fix to support jaqal subcircuits before #1243

currently the output of qscout_compile will contain \*either\* `.jaqal_program` (a string for a single circuit) or `.jaqal_programs` (a list for multiple circuits). After this pr it will always include both, where `.jaqal_program` is still always a single string but may contain subcircuits, and `.jaqal_programs` is always a list (possibly of length 1)

### DIFF
--- a/docs/source/optimizations/qscout/qscout_css.ipynb
+++ b/docs/source/optimizations/qscout/qscout_css.ipynb
@@ -55,8 +55,9 @@
     "    print(\"You may need to restart the kernel to import newly installed packages.\")\n",
     "    import cirq_superstaq as css\n",
     "\n",
-    "import cirq\n",
     "import random\n",
+    "\n",
+    "import cirq\n",
     "import numpy as np"
    ]
   },
--- a/docs/source/optimizations/qscout/qscout_qss.ipynb
+++ b/docs/source/optimizations/qscout/qscout_qss.ipynb
@@ -55,9 +55,10 @@
     "    print(\"You may need to restart the kernel to import newly installed packages.\")\n",
     "    import qiskit_superstaq as qss\n",
     "\n",
-    "import qiskit\n",
     "import random\n",
-    "import numpy as np"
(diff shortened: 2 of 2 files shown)


cirq-superstaq/cirq_superstaq/compiler_output.py
```diff
@@ -141,14 +139,28 @@ def __repr__(self) -> str:
...
+
+        For multi-circuit compilation the string will contain subcircuits.
+        """
+        if not self.jaqal_programs:
+            return None
+
+        separator = "prepare_all"
+        subcircuits = [self.jaqal_programs[0]]
+        subcircuits += [program.partition(separator)[2] for program in self.jaqal_programs[1:]]
+        return f"\n{separator}".join(subcircuits)
```
bharat-thotakura, line 162: Using the `prepare_all` concatenation makes sense here for maximum compatibility; however, is there any preference to use the newer `subcircuit` notation?

Image

Also, we could return it in the `subcircuit` notation server-side, but that would make maintaining `.jaqal_programs` more annoying ...

bharat-thotakura (commented):
We should update/re-run the QSCOUT compile notebooks in light of this PR, but that can be done separately before deploying

cirq-superstaq/cirq_superstaq/compiler_output.py
richrines1: my thinking was to save breaking changes for the v3.0 api / #1243, so at least compiler_output.jaqal_circuit should stick with prepare_all? (but when we do add this functionality to v3.0 imo the server should definitely just return the circuit with subcircuits instead of doing this client side, regardless of which notation we use)

bharat-thotakura: Sounds good, that makes sense to me

PR #1302, opened 2025-11-14
machine api v0

richrines1:
simplified/expanded version of #1254

(tests should pass after server deploy)

### DIFF
--- a/general-superstaq/general_superstaq/machine_api.py
+++ b/general-superstaq/general_superstaq/machine_api.py
@@ -2,19 +2,9 @@
 
 from __future__ import annotations
 
-import dataclasses
 from collections.abc import Collection
 
 import general_superstaq as gss
-
-
-@dataclasses.dataclass
-class Task:
-    """A task to be executed on the machine."""
-
-    task_id: str
-    circuit: str
-    shots: int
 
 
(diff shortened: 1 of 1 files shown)


cirq-superstaq/cirq_superstaq/worker.py
```diff
@@ -0,0 +1,94 @@
...
+from general_superstaq.machine_api import MachineAPI
+
+import cirq_superstaq as css
+
+if TYPE_CHECKING:
+    import cirq
+
+
+@dataclasses.dataclass
+class Task:
```
bharat-thotakura, line 17: Is it worth renaming this `Task` name to distinguish it from `gss.machine_api.Task` (and I guess `gss.WorkerTask` is a different type entirely)? Or would they not really overlap?

bharat-thotakura (approved):
LGTM

cirq-superstaq/cirq_superstaq/worker.py
richrines1: oops, `gss.machine_api.Task` wasn't supposed to exist at all. just deleted.

(it's intentionally similar to WorkerTask - it's basically a container for the same info, except that the circuit has been deserialized)

PR #1314, opened 2025-12-05
Update target fields of `"aqt_keysight_qpu"` in `cirq` integration test

bharat-thotakura:
Fixes: https://github.com/Infleqtion/client-superstaq/issues/1313


vtomole:
Why was this available and now it's not?

PR #1325, opened 2025-12-27
update various qubit subspace gate protocols

richrines1:
* adds support for mixed-unitary gates (e.g. depolarizing noise in the qubit subspace)
* accounts for interchangeable qubits in `sub_gate`, when possible
* fixes a bug in `QubitSubspaceGate._apply_unitary_` when `sub_gate` doesn't have its own `_apply_unitary_` or `_unitary_` methods


cirq-superstaq/cirq_superstaq/ops/qudit_gates.py
```diff
@@ -576,6 +607,20 @@ def _trace_distance_bound_(self) -> float:
     def _has_unitary_(self) -> bool:
         return cirq.has_unitary(self._sub_gate)
 
+    def _unitary_(self) -> npt.NDArray[np.complex128] | None:
+        sub_unitary = cirq.unitary(self._sub_gate, default=None)
+
+        if sub_unitary is None:
+            return NotImplemented
```
dowusu-antwi, line 614: Why `NotImplemented` instead of `return None`?

cirq-superstaq/cirq_superstaq/ops/qudit_gates.py
```diff
@@ -586,7 +631,31 @@ def _apply_unitary_(self, args: cirq.ApplyUnitaryArgs) -> npt.NDArray[np.complex
...
+        if sub_mixture is None:
+            return NotImplemented
+
+        dim = np.prod(self._qid_shape).item()
+        grid = np.meshgrid(*self._subspaces, *self._subspaces, indexing="ij")
+        idxs = np.ravel_multi_index(grid, 2 * self._qid_shape)
+
+        mixture = []
+        for prob, sub_unitary in sub_mixture:
+            unitary = np.eye(dim, dtype=complex)
```
dowusu-antwi: nit: change `complex` to `np.complex128` for consistency?

richrines1: done.

cirq-superstaq/cirq_superstaq/ops/qudit_gates.py
richrines1: i believe most cirq gates use NotImplemented instead of None, which should tell various protocols (e.g. `cirq.unitary()`) to try another method instead of giving up. in this case there isn't really another method to try so it shouldn't matter, but for consistency i left it that way

(did realize the annotation was off though - now fixed)

richrines1: actually found a weird corner case where it matters:
```python
sub_gate = cirq.X ** sympy.Symbol("x")
op = css.qubit_subspace_op(sub_gate(cirq.q(0)), [3])
print(cirq.mixture(op, default="default"))
```
currently (correctly) prints "default", but if i change `NotImplemented` to `None` i instead get `((1.0, None),)`. this seems like it might be a cirq bug though

Issue #1321, opened 2025-12-19
consider making stim requirement optional


richrines1:
### What is happening?

installing stim adds ~10 minutes to our python 3.14 and windows checks (and in the latter case probably [won't](https://github.com/quantumlib/Stim/issues/1001) be fixed soon)

we only include the dependency so we can (de)serialize stimcirq gates, which we probably don't need to include in tests

PR #1327, opened 2025-12-31
make stimcirq requirement optional

richrines1:
closes: #1321


cirq-superstaq/pyproject.toml
```diff
@@ -37,6 +37,9 @@ file = ["dev-requirements.txt"]
 [tool.setuptools.dynamic.optional-dependencies.examples]
 file = ["example-requirements.txt"]
 
+[tool.setuptools.dynamic.optional-dependencies.stim]
+file = ["stim-requirements.txt"]
```
bharat-thotakura, line 41: Is it worth testing a stim installation in the coverage checks whose Python version has pre-built wheels? Or would this file just be local user thing?

richrines1: updated dev-requirements to install stimcirq on systems with prebuilt wheels so we get coverage everywhere it makes sense

bharat-thotakura: Thanks, sounds good! Btw, are we going to include the `stim-requirements.txt` file in the repo for Python 3.14 users?

richrines1: oops yep just forgot to git-add it :)

PR #1329, opened 2026-01-02
more verbose error reporting in `get_targets` integration tests

richrines1:
in response to recent test flakes

(also cleans up the implementation of `get_my_targets`)

### DIFF
--- a/general-superstaq/general_superstaq/service.py
+++ b/general-superstaq/general_superstaq/service.py
@@ -195,14 +195,45 @@ class Service:
         )
         return self._client.get_targets(**filters)
 
-    def get_my_targets(self) -> list[gss.Target]:
+    def get_my_targets(
+        self,
+        simulator: bool | None = None,
+        supports_submit: bool | None = None,
+        supports_submit_qubo: bool | None = None,
+        supports_compile: bool | None = None,
+        available: bool | None = None,
+        retired: bool | None = None,
+        **kwargs: bool,
+    ) -> list[gss.Target]:
         """Gets a list of Superstaq targets that the user can submit to and are available along
         with their status information.
+
+        Args:
+            simulator: Optional flag to restrict the list of targets to (non-) simulators.
+            supports_submit: Optional boolean flag to only return targets that (don't) allow
+                circuit submissions.
+            supports_submit_qubo: Optional boolean flag to only return targets that (don't)
+                allow qubo submissions.
+            supports_compile: Optional boolean flag to return targets that (don't) support
+                circuit compilation.
+            available: Optional boolean flag to only return targets that are (not) available
+                to use.
+            retired: Optional boolean flag to only return targets that are or are not retired.
+            kwargs: Any additional, supported flags to restrict/filter returned targets.
 
         Returns:
             A list of Superstaq targets that the user can currently submit to.
         """
-        return self.get_targets(accessible=True)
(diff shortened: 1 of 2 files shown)


cirq-superstaq/cirq_superstaq/daily_integration_test.py
```diff
@@ -193,7 +193,15 @@ def test_get_targets(service: css.Service) -> None:
...
     assert ibmq_target_info in result
     assert aqt_target_info in result
-    assert all(target in result for target in filtered_result)
+
+    unfiltered_targets = {t.target: t for t in result}
+    for target in filtered_result:
+        assert target.target in unfiltered_targets, f"'{target.target}' not in unfiltered result"
+        assert target == unfiltered_targets[target.target], (
+            "Divergent targets.\nFiltered: {target!r}\n"
+            "Unfiltered: {unfiltered_targets[target.target]!r}"
```
bharat-thotakura:
```suggestion
            f"Divergent targets.\nFiltered: {target!r}\n"
            f"Unfiltered: {unfiltered_targets[target.target]!r}"
```

qiskit-superstaq/qiskit_superstaq/daily_integration_test.py
```diff
@@ -50,7 +50,15 @@ def test_backends(provider: qss.SuperstaqProvider) -> None:
...
     assert ibmq_backend_info in result
     assert len(provider.backends()) == len(result)
-    assert all(target in result for target in filtered_result)
+
+    unfiltered_targets = {t.target: t for t in result}
+    for target in filtered_result:
+        assert target.target in unfiltered_targets, f"'{target.target}' not in unfiltered result"
+        assert target == unfiltered_targets[target.target], (
+            "Divergent targets.\nFiltered: {target!r}\n"
+            "Unfiltered: {unfiltered_targets[target.target]!r}"
```
bharat-thotakura:
```suggestion
            f"Divergent targets.\nFiltered: {target!r}\n"
            f"Unfiltered: {unfiltered_targets[target.target]!r}"
```

cirq-superstaq/cirq_superstaq/daily_integration_test.py
richrines1: good catch, thanks :)

qiskit-superstaq/qiskit_superstaq/daily_integration_test.py
```diff
@@ -50,7 +50,15 @@ def test_backends(provider: qss.SuperstaqProvider) -> None:
     )
     assert ibmq_backend_info in result
     assert len(provider.backends()) == len(result)
```
bharat-thotakura: Actually, I think we've had the [`len` assertion fail before](https://github.com/Infleqtion/client-superstaq/actions/runs/20372291833/job/58541622944) as well. Could we move the target `for` loop before this assertion or check backends in addition or something?

richrines1: good call. updated to check for/report missing backends in the same way as the filtered targets

richrines1:
(also updated to drop `SuperstaqClient.get_my_targets` entirely, and instead just have `gss.Service.get_my_targets` call `self.get_targets(..., accessible=True)`)

PR #1339, opened 2026-01-21
Update timeline viewer notebooks with circuit duration estimates

bharat-thotakura:
Updates the `ibmq_dd_strategies_qss.ipynb` notebook to not use the retired `"ibmq_brisbane_qpu"`, and gets circuit duration estimates for `ibmq_compile_*.ipynb` to fix the recent failing notebook checks.

Note: After https://github.com/Infleqtion/client-superstaq/pull/1334, setting circuit durations is no longer done manually. Instead, as recommended, the notebooks use the `.estimate_duration()` method to determine the circuit duration value (at least until the duration setting is supported/needed in `qiskit`). This is necessary as `qiskit`'s timeline drawer seems to call `load_program()` which [then checks `program._duration`](https://github.com/Qiskit/qiskit/blob/aedc6f0aaad5b4a51d972b276017c64a3f9c0a19/qiskit/visualization/timeline/core.py#L263)


docs/source/optimizations/ibm/ibmq_compile_css.ipynb
```diff
@@ -334,31 +337,45 @@
...
    "execution_count": 10,
+   "id": "fc6cf4cc",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Estimate circuit durations for the timeline drawer\n",
+    "pulse_circuits = []\n",
+    "for pg_circ in pulse_gate_circuits:\n",
+    "    pg_circ.duration = pg_circ.estimate_duration(backend_target, unit=\"dt\")\n",
+    "    pulse_circuits.append(pg_circ)"
```
richrines1: why do we need to create a new list here? it looks like the items in the original list are being updated in-place so we could keep using `pulse_gate_circuits` below
```suggestion
    "for pg_circ in pulse_gate_circuits:\n",
    "    pg_circ.duration = pg_circ.estimate_duration(backend_target, unit=\"dt\")\n",
```

docs/source/optimizations/ibm/ibmq_compile_qss.ipynb
```diff
@@ -307,31 +310,45 @@
...
+   "id": "ad90e8aa",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Estimate circuit durations for the timeline drawer\n",
+    "pulse_circuits = []\n",
+    "for pg_qc in pulse_gate_circuits:\n",
+    "    pg_qc.duration = pg_qc.estimate_duration(backend_target, unit=\"dt\")\n",
+    "    pulse_circuits.append(pg_qc)"
+   ]
```
richrines1, line 320: (ditto)

docs/source/optimizations/ibm/ibmq_compile_css.ipynb
bharat-thotakura: Yeah, we don't, was an oversight on my part. Updated in https://github.com/Infleqtion/client-superstaq/pull/1339/commits/4e6c25e28b643b088e1480e213d55260d6e487cc

PR #1350, opened 2026-02-24
correct job type inference

richrines1:
currently jobs submitted to external targets ending in "_simulator" are marked as `JobType.SIMULATE`, however these should be treated as normal submissions ("simulate" is reserved for local simulations by superstaq)


general-superstaq/general_superstaq/superstaq_client_test.py
```diff
@@ -422,10 +422,12 @@ def test_supertstaq_client_create_job(
     else:
         job_type = "submit"
         sim_method = None
-        if target.endswith("_simulator") or method == "sim":
+        if method == "sim":
             job_type = "simulate"
             if method == "sim":
```
cdbf1: Is this second conditional necessary as it's within a conditional block where `method` is already `"sim"`?

cdbf1 (approved):
One nit otherwise lgtm

general-superstaq/general_superstaq/superstaq_client_test.py
richrines1: oops nope good catch

PR #1353, opened 2026-02-27
Sqalesim

vtomole:


vtomole:
This PR has 2 main implementations: The simulator which can be a plug-and-play and the the actual specifications of the noise model.

Issue #774, opened 2023-09-18
Create an abstract `gss.Job` parent class 


bharat-thotakura:
### What is happening?

With PR https://github.com/Infleqtion/client-superstaq/pull/691, `cirq_superstaq/job.py` and `qiskit_superstaq/superstaq_job.py` share a lot of common implementations. Motivated by Rich's [comment](https://github.com/Infleqtion/client-superstaq/pull/691#pullrequestreview-1620127681), it would be useful to have a parent `gss.Job` class to factor out standard functionality in `css` & `qss`. 

### What are the current alternatives?

Update `css` and `qss` individually.

PR #1357, opened 2026-03-03
factor out common JobV3 components to `gss.job.Job`

richrines1:
Towards https://github.com/Infleqtion/client-superstaq/issues/774

### DIFF
--- a/general-superstaq/general_superstaq/job.py
+++ b/general-superstaq/general_superstaq/job.py
@@ -225,7 +225,7 @@ class Job:
         """
         return self.job_data.target
 
-    def repetitions(self) -> int:
+    def _repetitions(self) -> int:
         """Gets the number of repetitions requested for this job.
 
         Returns:
--- a/cirq-superstaq/cirq_superstaq/job.py
+++ b/cirq-superstaq/cirq_superstaq/job.py
@@ -510,6 +510,7 @@ class JobV3(gss.job.Job):
     """
 
     status = gss.job.Job._status
+    repetitions = gss.job.Job._repetitions
 
     @overload
     def num_qubits(self, index: int) -> int: ...
--- a/cirq-superstaq/cirq_superstaq/job_test.py
+++ b/cirq-superstaq/cirq_superstaq/job_test.py
@@ -649,6 +649,7 @@ def test_job_countsV3(
     assert jobV3.counts(index=0) == {"10": 1}
     assert jobV3.counts(index=0, qubit_indices=[0]) == ({"1": 1})
     assert jobV3.counts() == [{"10": 1}]
+    assert jobV3.repetitions() == 1
 
 
 def test_job_counts_failed(job: css.Job, job_dict: dict[str, object]) -> None:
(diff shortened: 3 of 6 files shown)


general-superstaq/general_superstaq/job.py
```diff
@@ -0,0 +1,273 @@
...
+
+        Args:
+            client: The client used for calling the API.
+            job_id: Unique identifier for the job.
+
+        Raises:
+            TypeError: If `JobV3` is used with `v0.2.0` of the Superstaq API.
+        """
+        if not isinstance(client, _SuperstaqClientV3):
+            raise TypeError("JobV3 job can only be used with v0.3.0 of the Superstaq API.")
```
bharat-thotakura: `gss.Job` is technically not `JobV3`. Do we still want to have this exception since that will be the default in the future?

general-superstaq/general_superstaq/job.py
```diff
@@ -0,0 +1,273 @@
...
+        """Gets the Superstaq target associated with this job.
+
+        Returns:
+            The target to which this job was submitted.
+
+        Raises:
+            ~gss.SuperstaqUnsuccessfulJobException: If the job failed or has been
+                canceled or deleted.
+            ~gss.SuperstaqServerException: If unable to get the status of the job
+                from the API.
```
bharat-thotakura: Could this method potentially raise these exceptions just to get the job's target?

general-superstaq/general_superstaq/job.py
```diff
@@ -0,0 +1,273 @@
...
+        """
+        return self.job_data.target
+
+    def repetitions(self) -> int:
```
bharat-thotakura: Given this is in `gss`, is `repetitions()` too biased for `css.Job`  vs. having `shots()` for `qss.SuperstaqJob`? That is, should these convenience methods be implemented in their respective subclasses?

general-superstaq/general_superstaq/job.py
richrines1: ah good point. I just removed the exception (we already tell the user not to instantiate the class directly anyway)

general-superstaq/general_superstaq/job.py
richrines1: good catch - i think `job_data` could possibly raise the second one but not the first. updated.

general-superstaq/general_superstaq/job.py
richrines1: good question, i'm not really sure (tbh i kinda prefer "repetitions" over "shots" generally, but not sure how much they're associated with cirq/qiskit respectively)

bharat-thotakura: I think we mainly distinguish it when doing `backend.run(shots=100)` in `qss` vs. `service.create_job(repetitions=100)` in `css` (and they can't be used interchangeably, so I've occasionally forgotten to specifiy the right `kwarg` when switching from `css` <> `qss`  lol so I thought I'd bring it up (& funnily enough, `repetitions()` is a shortcut for `job_data.shots`)). 

But this might just be a question for later on standardization as we introduce breaking changes in `v0.3.0`

richrines1: ah that's a good point, we should be consistent with the args we require when creating the jobs. fixed.

PR #1360, opened 2026-03-10
Use `pytest-split` to add some concurrency to integration test

bharat-thotakura:
Presently, the daily integration test workflow uses only 2 workers from `pytest-xdist` and generally can take a while to finish, waiting on some slower tests relative to others. Anticipating https://github.com/Infleqtion/client-superstaq/pull/1358, which [will slow down](https://github.com/Infleqtion/client-superstaq/actions/runs/22884111478) the workflow even more, this PR adds `pytest-split` to break up the (current) total of 44 unit tests from the `css` and `qss` integration tests combined into 4 concurrent workflows, each using 2 workers. (Note: the tests are split up evenly across the 4 workflows (called 'groups'), but `pytest-split` can unevenly group time-consuming tests if given `pytest` timing info) 

Additionally, on a more speculative & hopeful note, this PR also applies `pytest-split` on the three notebook checks in the regular CI to continue to mitigate against https://github.com/Infleqtion/client-superstaq/issues/1129 (with the idea being that each concurrent workflow should now only assign 1 notebook per worker (of course, subject to change if new notebooks are added)). In [initial testing](https://github.com/Infleqtion/client-superstaq/actions/runs/22887601930/job/66403774452), this did not seem to offer much in time reduction due to the small number of notebooks being tested to begin with (& the worker overhead perhaps), but this approach might still be worth experimenting with.

Lastly, to confer any potential speed benefits, this PR also updates all relevant CI jobs to use Python 3.14 from the previous Python 3.13

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -255,7 +255,6 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install ${{ matrix.dependencies }}
-          pip install pytest-split
       - name: Check notebooks
         if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name == github.repository
         run: |


.github/workflows/ci.yml
```diff
@@ -223,38 +223,54 @@ jobs:
...
+      - name: Set up Python 3.14
         uses: actions/setup-python@v6
         with:
-          python-version: 3.13
+          python-version: 3.14
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
           pip install ${{ matrix.dependencies }}
+          pip install pytest-split
```
richrines1: let's just make pytest-split a requirement of checks-superstaq, it's pretty lightweight

.github/workflows/daily-integration-check.yml
```diff
@@ -6,34 +6,39 @@ on:
...
+        steps:
+            - uses: actions/checkout@v6
+            - name: Set up Python 3.14
+              uses: actions/setup-python@v6
+              with:
+                  python-version: 3.14
+            - name: Install dependencies
+              run: |
+                  python -m pip install --upgrade pip
+                  python -m pip install ./checks-superstaq ./general-superstaq[dev] ./qiskit-superstaq[dev] ./cirq-superstaq[dev]
```
richrines1: getting outside the scope of this pr, but it'd probably be nice to also split up the css and qss integration tests (with each installing just the relevant packages)

.github/workflows/ci.yml
bharat-thotakura: Done in https://github.com/Infleqtion/client-superstaq/pull/1360/commits/26f35ca228414a0c69acfe92d331840da17d7ae8

.github/workflows/daily-integration-check.yml
bharat-thotakura: Added (see this [workflow](https://github.com/Infleqtion/client-superstaq/actions/runs/23204714836) for an example)

.github/workflows/daily-integration-check.yml
```diff
@@ -1,39 +1,55 @@
...
+            - name: Pytest integration check (group ${{ matrix.group }} of 2)
+              run: |
+                  checks/pytest_.py --integration --splits=2 --group=${{ matrix.group }} -v ${{ matrix.type }}
+            - name: Create issue if integration check failed
+              if: ${{ failure() }}
+              uses: JasonEtco/create-an-issue@v2
+              env:
+                  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+                  TEST_USER_IBMQ_TOKEN: ${{ secrets.TEST_USER_IBMQ_TOKEN }}
+                  TEST_USER_IBMQ_INSTANCE: ${{ secrets.TEST_USER_IBMQ_INSTANCE }}
```
richrines1: nit (predates this pr): we shouldn't need ibm creds to file a github issue
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -223,39 +223,55 @@ jobs:
...
```
richrines1: (shouldn't need to install manually anymore)
```suggestion
```

PR #1361, opened 2026-03-16
Exposing the `qubits` argument. 

vietphamngoc:
Exposing the `qubits` argument in the different benchmark to allow for targeting of specific qubits.

### DIFF
--- a/supermarq-benchmarks/supermarq/qcvv/irb.py
+++ b/supermarq-benchmarks/supermarq/qcvv/irb.py
@@ -621,20 +621,6 @@ class IRB(QCVVExperiment[_RBResultsBase]):
 
         return qubits
 
-        # Error if qubits is not sequence (for benefit of type checker)
-        assert isinstance(qubits, Sequence)
-
-        if interleaved_gate is not None and cirq.num_qubits(interleaved_gate) != len(qubits):
-            raise ValueError(
-                "The length of targeted qubits must match the number of qubits interleaved_gate"
-                " is acting on."
-            )
-        if isinstance(interleaved_gate, cirq.Operation) and list(interleaved_gate.qubits) != list(
-            qubits
-        ):
-            raise ValueError("The qubits provided do not match interleaved_gate's qubits")
-        return qubits
-
     def _clifford_gate_to_circuit(
         self,
         clifford: cirq.CliffordGate,


supermarq-benchmarks/supermarq/qcvv/irb.py
```diff
@@ -542,20 +542,21 @@ def __init__(
...
-        elif not isinstance(qubits, Sequence):
-            qubits = cirq.LineQubit.range(qubits)
+        if not isinstance(qubits, Sequence):
+            if isinstance(interleaved_gate, cirq.Gate):
+                qubits = cirq.LineQubit.range(interleaved_gate.num_qubits())
+            elif isinstance(interleaved_gate, cirq.Operation):
+                qubits = interleaved_gate.qubits
+                interleaved_gate = interleaved_gate.gate
+            else:
+                qubits = cirq.LineQubit.range(qubits)
```
richrines1: do we want to throw errors if a user passes in other conflicting arguments, e.g.
```python
IRB(..., interleaved_gate=cirq.CZ, qubits=1)
```
or
```python
IRB(..., interleaved_gate=cirq.Z(cirq.LineQubit(1)), qubits=[cirq.GridQubit(2,3)])
```
?

supermarq-benchmarks/supermarq/qcvv/ssb.py
```diff
@@ -196,10 +197,16 @@ def __init__(
...
             random_seed: An optional seed to use for randomization.
             kwargs: Additional kwargs passed to the base QCVVExperiment.
         """
-        qubits = cirq.LineQubit.range(2)
+        if qubits is None:
+            qubits = cirq.LineQubit.range(2)
+        else:
+            if len(qubits) != 2:
+                raise ValueError("SSB benchmarking is only designed for 2 qubits.")
+            qubits = qubits
```
richrines1:
```suggestion
        elif len(qubits) != 2:
            raise ValueError("SSB benchmarking is only designed for 2 qubits.")
```

supermarq-benchmarks/supermarq/qcvv/su2.py
```diff
@@ -246,8 +247,17 @@ def __init__(
...
         """The two qubit gate to be benchmarked"""
 
+        if qubits is None:
+            qubits = cirq.LineQubit.range(2)
+        else:
+            if len(qubits) != 2:
+                raise ValueError(
+                    "SU2 benchmarking is currently only implemented for 2 qubits gates."
+                )
+            qubits = qubits
```
richrines1:
```suggestion
        elif len(qubits) != 2:
            raise ValueError("SU2 benchmarking is currently only implemented for 2 qubits gates.")
```

supermarq-benchmarks/supermarq/qcvv/irb.py
vietphamngoc: Probably for the your first point we should raise an error. For your second point, in the case where a user wants to test a given interleaved gate on different set of qubits, they could do it just by changing the `qubits` arguments and without redefining the circuit

richrines1: my thought in the second case was that if the user is passing in `qubits`, they should be passing in a `cirq.Gate`, not a `cirq.Operation` (which already has its own qubits)

vietphamngoc: Sorry missed that point. I was thinking about the case where for some reason a user wanted to benchmark a Clifford circuit with IRB but that's off the mark here.

supermarq-benchmarks/supermarq/qcvv/irb.py
```diff
@@ -599,6 +595,37 @@ def __repr__(self) -> str:
...
+    ) -> Sequence[cirq.Qid]:
+        if qubits is None:
+            if interleaved_gate is None:
+                raise ValueError("Please provide either interleaved_gate or qubits.")
+            if isinstance(interleaved_gate, cirq.Gate):
+                return cirq.LineQubit.range(interleaved_gate.num_qubits())
+            return interleaved_gate.qubits
+
+        if isinstance(qubits, int):
+            if interleaved_gate is not None and interleaved_gate._num_qubits_() != qubits:
```
richrines1: nit
```suggestion
            if interleaved_gate is not None and cirq.num_qubits(interleaved_gate) != qubits:
```

supermarq-benchmarks/supermarq/qcvv/irb.py
```diff
@@ -599,6 +595,37 @@ def __repr__(self) -> str:
...
+                raise ValueError(
+                    "The number of qubits must match the number of qubits interleaved_gate is "
+                    "acting on."
+                )
+            return cirq.LineQubit.range(qubits)
+
+        if interleaved_gate is not None and interleaved_gate._num_qubits_() != len(qubits):
```
richrines1:
```suggestion
        if interleaved_gate is not None and cirq.num_qubits(interleaved_gate) != len(qubits):
```

supermarq-benchmarks/supermarq/qcvv/irb.py
```diff
@@ -599,6 +595,37 @@ def __repr__(self) -> str:
...
+            raise ValueError(
+                "The length of targeted qubits must match the number of qubits interleaved_gate"
+                " is acting on."
+            )
+        if isinstance(interleaved_gate, cirq.Operation) and set(interleaved_gate.qubits) != set(
+            qubits
+        ):
```
richrines1: i think we care about order here
```suggestion
        if isinstance(interleaved_gate, cirq.Operation) and list(interleaved_gate.qubits) != list(
            qubits
        ):
```

supermarq-benchmarks/supermarq/qcvv/irb.py
```diff
@@ -599,6 +595,37 @@ def __repr__(self) -> str:
...
+        qubits: int | Sequence[cirq.Qid] | None = None,
+    ) -> Sequence[cirq.Qid]:
+        if qubits is None:
+            if interleaved_gate is None:
+                raise ValueError("Please provide either interleaved_gate or qubits.")
+            if isinstance(interleaved_gate, cirq.Gate):
+                return cirq.LineQubit.range(interleaved_gate.num_qubits())
+            return interleaved_gate.qubits
+
+        if isinstance(qubits, int):
```
richrines1: to make this work correctly with other int-like types (e.g. `np.int*`):
```suggestion
        if isinstance(qubits, numbers.Integral):
```

supermarq-benchmarks/supermarq/qcvv/irb.py
vietphamngoc: You are right

supermarq-benchmarks/supermarq/qcvv/irb.py
```diff
@@ -599,6 +596,40 @@ def __repr__(self) -> str:
...
+        if interleaved_gate is not None and cirq.num_qubits(interleaved_gate) != len(qubits):
+            raise ValueError(
+                "The length of targeted qubits must match the number of qubits interleaved_gate"
+                " is acting on."
+            )
+        if isinstance(interleaved_gate, cirq.Operation) and list(interleaved_gate.qubits) != list(
+            qubits
+        ):
+            raise ValueError("The qubits provided do not match inteleaved_gate's qubits")
+        return qubits
```
richrines1: nit: i think this method can be simplified a bit, while also narrowing the return type to `list[cirq.Qid]` and avoiding the type assertion and `numbers` import:
```suggestion
    def _assign_qubits(
        self,
        interleaved_gate: cirq.Gate | cirq.Operation | None = None,
        qubits: int | Sequence[cirq.Qid] | None = None,
    ) -> list[cirq.Qid]:
        if qubits is None:
            if interleaved_gate is None:
                raise ValueError("Please provide either interleaved_gate or qubits.")
            if isinstance(interleaved_gate, cirq.Operation):
                qubits = interleaved_gate.qubits
            else:
                qubits = cirq.num_qubits(interleaved_gate)

        if not isinstance(qubits, Sequence):
            qubits = cirq.LineQubit.range(qubits)
        qubits = list(qubits)

        if interleaved_gate is not None and cirq.num_qubits(interleaved_gate) != len(qubits):
            raise ValueError(
                "The number of qubits must match the number of qubits interleaved_gate acts on."
            )
        if isinstance(interleaved_gate, cirq.Operation) and list(interleaved_gate.qubits) != qubits:
            raise ValueError("The qubits provided do not match interleaved_gate's qubits")
        return qubits
```

vietphamngoc: That's a good simplification. I'll take it

PR #1364, opened 2026-03-31
Add support for direct Jaqal ingestion with `JaqalService()`

bharat-thotakura:


general-superstaq/general_superstaq/validation.py
```diff
@@ -98,6 +99,27 @@ def validate_target(target: str) -> str:
...
+
+    Returns:
+        The max qubit register size needed for all `jaqal_programs`.
+
+    Raises:
+        ValueError: If no qubit count could be inferred from `jaqal_programs`.
+    """
+    jaqal_programs = [jaqal_programs] if isinstance(jaqal_programs, str) else jaqal_programs
+    pattern = re.compile(r"^\s*register\b.*?\[(\d+)\]", re.MULTILINE)
+    register_sizes = (int(m.group(1)) for jp in jaqal_programs for m in [pattern.search(jp)] if m)
```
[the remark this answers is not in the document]
bharat-thotakura: A jaqal program is usually expected to only have a single register declaration at the start, so this shouldn't be a concern

general-superstaq/general_superstaq/compiler_output.py
```diff
@@ -0,0 +1,224 @@
...
+        num_eca_circuits: Number of logically equivalent random circuits to generate for each
+            input circuit.
+
+    Returns:
+        A `CompilerOutput` object with the compiled Jaqal program(s).
+    """
+    compiled_circuits = json.loads(json_dict["jaqal_strs"])
+
+    # Convert a single subcircuit input back to single subcircuit output
+    if not circuits_is_list and len(compiled_circuits) != 1:
```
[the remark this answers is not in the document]
bharat-thotakura: This check has been removed in https://github.com/Infleqtion/client-superstaq/pull/1364/commits/8efc41d818fa75f7a03dda2c9a1f860056e032b6, but the population of the `CompilerOutput` object checks for `num_eca_circuits`

general-superstaq/general_superstaq/compiler_output.py
```diff
@@ -0,0 +1,224 @@
...
+            f"{self.seq!r}, {self.jaqal_programs!r})"
+        )
+
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, BaseCompilerOutput):
+            return False
+
+        if self.has_multiple_circuits() != other.has_multiple_circuits():
+            return False
+        if self.has_multiple_circuits():
```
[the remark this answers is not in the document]
bharat-thotakura: Going with `type(self) is type(other)` would make it so that `gss.compiler_output.BaseCompilerOutput(...) != qss.compiler_output.CompilerOutput(...)` even if they contained the _same_ contents. The main concern should be content equality, and different `CompilerOutput`'s are unlikely to be equal anyway due to the different circuit types they use. It also makes type checking easier to continue using `isinstance()` checks.

qiskit-superstaq/qiskit_superstaq/compiler_output.py
```diff
@@ -417,7 +347,9 @@ def read_json_qscout(
             for i in range(0, len(final_logical_to_physicals_list), num_eca_circuits)
         ]
         jaqal_programs = [
-            _jaqal_programs_to_subcircuits(jaqal_programs[i : i + num_eca_circuits])
+            gss.compiler_output._jaqal_programs_to_subcircuits(
+                jaqal_programs[i : i + num_eca_circuits]
+            )
             for i in range(0, len(jaqal_programs), num_eca_circuits)
```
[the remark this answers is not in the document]
bharat-thotakura: Jaqal is not the predominant input used in `cirq-superstaq` or `qiskit-superstaq`, so that function is unlikely to be a user-facing functionality broadly used (hence the underscore). Consolidating it in `general-superstaq` helps reduce the need to update the function in multiple places.

cirq-superstaq/cirq_superstaq/compiler_output.py
```diff
@@ -371,7 +325,9 @@ def read_json_qscout(
             for i in range(0, len(final_logical_to_physicals_list), num_eca_circuits)
         ]
         jaqal_programs = [
-            _jaqal_programs_to_subcircuits(jaqal_programs[i : i + num_eca_circuits])
+            gss.compiler_output._jaqal_programs_to_subcircuits(
+                jaqal_programs[i : i + num_eca_circuits]
+            )
             for i in range(0, len(jaqal_programs), num_eca_circuits)
         ]
```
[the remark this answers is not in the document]
bharat-thotakura: Same as https://github.com/Infleqtion/client-superstaq/pull/1364#discussion_r3048012431

general-superstaq/general_superstaq/service.py
```diff
@@ -575,3 +581,114 @@ def submit_atom_picture(self, bitmap: npt.ArrayLike) -> str:
...
+        Returns:
+            Object whose .circuit(s) attribute contains optimized Jaqal program(s).
+
+        Raises:
+            ValueError: If `base_entangling_gate` is not a valid gate option.
+            ValueError: If provided `num_qubits` is less than the register size required by
+                `jaqal_programs`.
+        """
+        target = gss.validation.validate_target(target)
+        circuits_is_list = not isinstance(jaqal_programs, str)
```
richrines1: if this is a string could it still contain subcircuits? if so maybe we just make this
```suggestion
        circuits_is_list = True
```
to avoid randomly changing return types between one and two+ subcircuits

general-superstaq/general_superstaq/service.py
```diff
@@ -575,3 +581,114 @@ def submit_atom_picture(self, bitmap: npt.ArrayLike) -> str:
         """
         request_id = self._client.submit_atom_picture(bitmap=bitmap).get("request_id")
         return f"Submitted request for atom picture with ID: {request_id}"
+
+
+class JaqalService(BaseService):
+    """This class contains services relating to Superstaq and Jaqal input."""
+
+    def compile(
```
richrines1: for now let's make this
```suggestion
    def qscout_compile(
```
because it's using the `/qscout_compile` endpoint. in the future we can add a generic `compile()` (which for the most part should hopefully just work otob)

general-superstaq/general_superstaq/compiler_output.py
```diff
@@ -0,0 +1,235 @@
...
+
+import json
+from collections.abc import Sequence
+from typing import Any, Generic, TypeVar
+
+C = TypeVar("C")
+Q = TypeVar("Q")
+
+
+class BaseCompilerOutput(Generic[C, Q]):  # noqa: PLW1641
```
richrines1, line 24: love the use of generics here - wondering if we could do something similar for `BaseService` (obviously not for this pr though)

general-superstaq/general_superstaq/service.py
bharat-thotakura: Currently, the server will always return a `list`, so if there were subcircuits, that is handled via the check in `read_json_jaqal`:
https://github.com/Infleqtion/client-superstaq/blob/2fccf755d5290af83aefa8559685f0dc19746da9/general-superstaq/general_superstaq/compiler_output.py#L223

If `circuits_is_list` is always `True`, we wouldn't be able to use the singular `.circuit` attribute right?

general-superstaq/general_superstaq/service.py
bharat-thotakura: Sure, it currently switches between `/qscout_compile` and `/compile` actually, with the latter relevant in the (currently an edge case) scenario where you want to compile a Jaqal string to a non-QSCOUT backend

general-superstaq/general_superstaq/compiler_output.py
bharat-thotakura: Yeah probably, but feels like we would end up implementing the specific overrides anyway (e.g., because of different serializations etc), so would have to see how much more is consolidated and not duplicated. Regardless, will leave for another PR when the `SuperstaqClientV3` becomes the norm

general-superstaq/general_superstaq/service.py
richrines1: ah, missed the check in `read_json_jaqal`, and also forgot we provide `.jaqal_program(s)` regardless of `circuits_is_list` (though maybe that raises a question about whether the compiler output from jaqal service should just use `.circuit(s)` instead of `.jaqal_program(s)`?)

> we wouldn't be able to use the singular .circuit attribute right?

my thinking was that this would make sense in this case, because if i pass a single jaqal string there's no way to tell if i intended it to be a standalone circuit or a program with a single subcircuit (and imo it's not great to change switch just given the number of circuits, because it makes the result type inconsistent in any sort of scripting). but feel less strongly due to the existence of `.jaqal_program(s)` (and the coming deprecation of compiler outputs anyway)

general-superstaq/general_superstaq/service.py
bharat-thotakura: Resolved by https://github.com/Infleqtion/client-superstaq/pull/1364/commits/4f8991c61f0a9c3c5fcfcb80af4db1ccf14eb885

general-superstaq/general_superstaq/service.py
bharat-thotakura: I initially went with `circuits_is_list = True`, so I'm not opposed to it since the attribute change isn't elegant (I think I didn't want to index by 0 for single-str inputs). However, in light of the review, I've just made it always a `list` with https://github.com/Infleqtion/client-superstaq/pull/1364/commits/77ca880613620abcfcd6aec86c6df1b0fde00459. Let me know if that works better

general-superstaq/general_superstaq/compiler_output.py
richrines1: ah good point, that's probably true

richrines1 (approved):
lgtm! (assuming tests will pass after #1380)

bharat-thotakura:
> (assuming tests will pass after #1380)

Yup, they should (except for the unrequired notebook checks)

PR #1365, opened 2026-04-02
support fractional counts in jobs

richrines1:
some jobs may have non-integer "counts" (for example due to MEM). this pr updates various annotations to reflect this

### DIFF
--- a/cirq-superstaq/cirq_superstaq/service.py
+++ b/cirq-superstaq/cirq_superstaq/service.py
@@ -1159,7 +1159,9 @@ class Service(gss.service.Service):
             noise=noise_dict,
         )
 
-    def process_cb(self, job_id: str | uuid.UUID, counts: Sequence[Mapping[str, float]] | None = None) -> dict[str, Any]:
+    def process_cb(
+        self, job_id: str | uuid.UUID, counts: Sequence[Mapping[str, float]] | None = None
+    ) -> dict[str, Any]:
         """Processes the data from the Cycle Benchmarking protocol.
         Generates SPAM and decay parameter estimations in addition to the process infidelity.
 


general-superstaq/general_superstaq/models.py
```diff
@@ -180,7 +180,7 @@ class JobData(DefaultPydanticModel):
     """The input circuits as serialized strings."""
     circuit_type: CircuitType
     """The circuit type used for representing the circuits."""
-    counts: list[dict[str, int] | None]
+    counts: list[dict[str, int | float] | None]
```
bharat-thotakura, line 183: Is this not:
```suggestion
    counts: list[dict[str, float] | None]
```
for `mypy`?

cirq-superstaq/cirq_superstaq/service.py
bharat-thotakura, line 1: While on this topic, should the counts of `process_cb()` also be updated? Or would those counts not use something like MEM? Relatedly, but somewhat unrelated to this PR, I'm also unsure why the docstring for `process_cb()` in `general-superstaq/general_superstaq/superstaq_client.py` says that the counts are a `dict` but the type annotation is `str | None = None`?

general-superstaq/general_superstaq/models.py
richrines1: for mypy that would be fine (because int is treated as a subclass of float for typing purposes), but it would have a slightly different effect due to pydantic's validation: with `dict[str, float]` pydantic would transform all values into floats, e.g. i pass in `counts=[{"00": 100}], ...` pydantic would change it to `counts=[{"00": 100.0}]`, whereas with `dict[str, int | float]` both floats and ints will be unchanged

bharat-thotakura: Yeah, I found `mypy` didn't complain, but I didn't consider the `pydantic` part. Thanks for the explanation!

cirq-superstaq/cirq_superstaq/service.py
richrines1: ah yep looks like i missed `process_cb`, fixed!

for `superstaq_client`, apparently it's supposed to take a serialized sequence of counts dictionaries so the docstring seems wrong..

that said, really all those qcvv methods do now is raise this:
```
SuperstaqServerException: Superstaq's ACES, CB, and DFE endpoints have been disabled as we move this functionality to our open-source clients. Stay tuned! (Status code: 400, non-retriable error making request to Superstaq API)
```
so it might be time to start deleting them? (i believe we left the aces endpoints in place because they're referenced in [a paper](https://arxiv.org/abs/2403.12857), but afaik the cb/dfe endpoints are not referenced anywhere and so could be removed now - maybe @vtomole can confirm.) All this is definitely getting outside the scope of this pr though so i'll punt for now :)

PR #1367, opened 2026-04-03
ingestion of qscout results objects

richrines1:


qiskit-superstaq/dev-requirements.txt
```diff
@@ -1 +1,3 @@
 checks-superstaq~=0.5.64
+jaqalpaq>=1.3.0
```
cdbf1, line 2: Do these need to be in the css and qss requirements if they get pulled in anyway from gss?

cirq-superstaq/cirq_superstaq/job.py
```diff
@@ -726,6 +726,29 @@ def counts(
             return _get_marginal_counts(single_counts, qubit_indices)
         return single_counts
 
+    def _terminal_measurement_qubit_indices(self, index: int) -> list[int]:
+        """Returns the ordered physical qubit indices for each measurement in a compiled circuit.
```
cdbf1, line 730: Nit: args and return docs (although its a private function so I guess somewhat optional)

qiskit-superstaq/dev-requirements.txt
richrines1: currently the dev options don't pull from one another (so e.g. `cirq-superstaq[dev]` still just requires `general-superstaq` and not `general-superstaq[dev]`). i think the idea is that each `<package>[dev]` should only add the dependencies necessary to run checks on <package>, and not on other `*-superstaq` packages.. (but not sure how much of a difference this makes in practice atm)

cirq-superstaq/cirq_superstaq/job.py
richrines1: done.

general-superstaq/general_superstaq/job.py
```diff
@@ -16,11 +16,26 @@
...
 import general_superstaq as gss
 
+if TYPE_CHECKING:
+    import jaqalpaq.run.result
+
+
+class Endian(str, enum.Enum):
+    """Endianness to use when mapping quantum objects to matrices, state vectors, and bitstrings."""
+
+    BIG = "big"  # BQSKit, Braket, Cirq, Jaqal, TKET
```
bharat-thotakura: μ-nit
```suggestion
    BIG = "big"  # BQSKit, Braket, Cirq, Jaqal, TKET, CUDA-Q
```

bharat-thotakura (approved):
LGTM

PR #1371, opened 2026-04-08
Adding ty type checking (experimental)

cdbf1:


vtomole:
To get this to pass, install something like types-requests.

general-superstaq/general_superstaq/machine_api.py
```diff
@@ -68,7 +68,9 @@ def post_result(
             status=status,
             status_message=status_message,
             successful_shots=len(bitstrings) if bitstrings is not None else None,
-            measurements=compressed_bitstrings,
+            measurements={k: set(v) for k, v in compressed_bitstrings.items()}
```
richrines1: nit: can we just update the type of/how we construct `compressed_bitstrings` above instead of constructing it with lists and then transforming it here?

general-superstaq/general_superstaq/models.py
```diff
@@ -153,7 +153,8 @@ def _validate_cq_token(cls, cq_token: object) -> object:
         Previously CQ tokens were specified via a dict, e.g. `cq_token={"access_token": "token"}`.
         """
         if isinstance(cq_token, Mapping):
-            return cq_token.get("access_token")
+            cq_token_mapping = cast("Mapping[str, object]", cq_token)
```
richrines1: nit (doesn't affect anything but technically we don't know the keys are strings so we shouldn't cast them as such):
```suggestion
            cq_token_mapping = cast(Mapping[object, object], cq_token)
```

(tbh i'm kinda surprised this is necessary though?)

general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -303,8 +305,11 @@ def _handle_status_codes(self, response: requests.Response) -> None:
             if isinstance(json_content, dict) and set(json_content.keys()).intersection(
                 {"message", "detail"}
             ):
-                alternative: str = json_content.get("detail", "")
-                message: str = json_content.get("message", alternative)
+                error_content = {str(key): value for key, value in json_content.items()}
+                detail = error_content.get("detail")
+                alternative = detail if isinstance(detail, str) else ""
+                message_content = error_content.get("message")
+                message = message_content if isinstance(message_content, str) else alternative
```
richrines1: if we change the annotation of `_handle_response` to `pydantic.JsonValue` (which is more precise anyway) we can revert these changes

general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -386,6 +391,22 @@ def _extract_credentials(kwargs: dict[str, Any]) -> dict[str, str]:
 
         return credentials
 
+    @staticmethod
+    def _normalize_job_ids(job_ids: Sequence[str] | Sequence[uuid.UUID]) -> list[uuid.UUID]:
```
richrines1: pydantic is supposed to do this kind of type coercion for us, i don't think we should be adding these manual validations/conversions on top of it (for example the pydantic plugin for mypy tells it to treat model arguments [as `Any`](https://pydantic.dev/docs/validation/latest/integrations/mypy/#init_typed) by default, but unfortunately it seems like ty's pydantic support is still a wip..)

not sure how to best circumvent this though.. one possible (but definitely suboptimal) workaround would be to add something like
```
[tool.ty.analysis]
replace-imports-with-any = ["pydantic.main.**"]
```
to our pyproject.yaml?

general-superstaq/general_superstaq/models.py
cdbf1: Updated as you suggest to be the most generic. I think this is a quirk of `ty` being especially strict. Without the cast, the type checker has no way to infer the key, value types of the mapping and so `ty` seems to default to `Never` for the key parameter.

general-superstaq/general_superstaq/superstaq_client.py
cdbf1: I agree with updating annotation of `_handle_response` to `pydantic.JsonValue` and while this satisfies `ty` it introduces an error from `mypy` instead (since `json_content.get("detail", "")` is not garunteed to be a string.
Ultimately, either way some handling of non-string response values is required.

I've modified a little bit to now satisfy both

general-superstaq/general_superstaq/superstaq_client.py
cdbf1: Pydantic does the coercion for us, but `ty` is correct that we are technically passing in a value of the wrong type since pydantic doesnt have a way of annotating the coercion it does. For now my suggestion is targeted ignore lines.

.github/workflows/ci.yml
```diff
@@ -93,6 +93,27 @@ jobs:
...
+        run: |
+          python -m pip install --upgrade pip
+          pip install -e ./checks-superstaq -e ./general-superstaq[dev] -e ./qiskit-superstaq[dev] -e ./cirq-superstaq[dev] -e ./supermarq-benchmarks[dev]
+      - name: Type check
+        run: |
+          checks/ty_.py general-superstaq
+
   coverage:
     name: Pytest and Coverage check
     strategy:
```
dowusu-antwi, line 119: Do we set a minimum supported Python version for client? We say 3.9 in the [docs](https://superstaq.readthedocs.io/en/latest/get_started/installation.html#installation-guide). Or, is the goal to bring all type-syntax up to date with 3.14?

checks-superstaq/checks_superstaq/ty_.py
```diff
@@ -0,0 +1,64 @@
...
+        silent: If True, restrict printing to warning and error messages.
+
+    Returns:
+        Terminal exit code. 0 indicates success, while any other integer indicates a test failure.
+    """
+    parser = check_utils.get_check_parser()
+    parser.description = textwrap.dedent(
+        """
+        Runs ty on the repository (typing check).
+        Ignores files in the [repo_root]/examples directory.
```
dowusu-antwi: Is this handled somewhere? Or are we ignoring it for now since `gss` has no `examples`?

general-superstaq/general_superstaq/superstaq_client.py
richrines1: hmm now i'm a little confused. shouldn't the type checker only care about the annotation of `__init__` when checking the arguments passed to the class constructor? `pydantic.BaseModel.__init__` [has this annotation](https://github.com/pydantic/pydantic/blob/1078a0516b1854708915c0157df5fab059cc453b/pydantic/main.py#L253):
```python
def __init__(self, /, **data: Any) -> None:
```
which should mean that all input arguments accept anything

.github/workflows/ci.yml
bharat-thotakura: Relevant: [`ty docs`](https://docs.astral.sh/ty/python-version/#:~:text=By%20default%2C%20the,ty%20(currently%203.14)) on inferring the Python version. 

We may want to check explicitly setting this (like you can in `mypy`)

checks-superstaq/checks_superstaq/ty_.py
cdbf1: Nope, that was because I copied it from somewhere else - thanks for catching!

dowusu-antwi:
Dev discussion: Consider doing some kind of `ty ignore` (not sure about syntax) for the expected type errors, b/c `ty` is still in alpha and so this might change.

dowusu-antwi:
Consider pyrefly: https://htmlpreview.github.io/?https://github.com/python/typing/blob/main/conformance/results/results.html 

https://github.com/quantumlib/Cirq/issues/8182

cdbf1:
Oh yeah, pyrefly looks interesting. I read a useful comparison blog which highlights some of the differences in approach nicely: https://blog.edward-li.com/tech/comparing-pyrefly-vs-ty/

PR #1378, opened 2026-04-17
add `job.combined_counts()`

richrines1:
requires #1367


cirq-superstaq/cirq_superstaq/ops/qudit_gates_test.py
```diff
@@ -269,7 +269,7 @@ def test_qutrit_z_pow_gate() -> None:
     )
 
     cirq.testing.assert_has_diagram(
-        cirq.Circuit(css.QutritZ1(cirq.LineQid(0, 3)) ** 1.2),
```
bharat-thotakura, line 272: Was it intentional previously to check circuit diagram equivalence when raised to 1.2? If so, should both assertions, 1.2 and -0.8, be worthwhile to have?

bharat-thotakura (approved):
LGTM

cirq-superstaq/cirq_superstaq/ops/qudit_gates_test.py
bharat-thotakura: From dev meeting: we cannot use the 1.2 exponent as newer `cirq` versions will not produce the same diagram

PR #1383, opened 2026-04-29
Transition to job objects for all `v0.3.0` compile endpoints

bharat-thotakura:
Towards https://github.com/Infleqtion/client-superstaq/issues/1243


general-superstaq/general_superstaq/typing.py
```diff
@@ -15,12 +15,21 @@
 from __future__ import annotations
 
 from collections.abc import Hashable
-from typing import Any, TypedDict, TypeVar
+from typing import TYPE_CHECKING, Any, Literal, TypedDict
 
 import pydantic
+from typing_extensions import TypeVar
```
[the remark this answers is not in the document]
bharat-thotakura: It shouldn't cause an `ImportError` because `general-superstaq` requires `typing_extensions` via the following chain:
```bash
general-superstaq==0.5.66
├── numpy [required: >=1.21.0, installed: 2.4.4]
├── pydantic [required: >=2.6.0, installed: 2.13.4]
│   ├── annotated-types [required: >=0.6.0, installed: 0.7.0]
│   ├── pydantic_core [required: ==2.46.4, installed: 2.46.4]
│   │   └── typing_extensions [required: >=4.14.1, installed: 4.15.0]
│   ├── typing_extensions [required: >=4.14.1, installed: 4.15.0]
│   └── typing-inspection [required: >=0.4.2, installed: 0.4.2]
│       └── typing_extensions [required: >=4.12.0, installed: 4.15.0]
```
Ultimately, the choice to import from `typing_extensions` and not `typing` is for <Python 3.13 compatibility in using `default=`  (see PEP 696)  for the `TypeVar` (though not used here specifically)

general-superstaq/general_superstaq/typing.py
```diff
@@ -15,12 +15,21 @@
...
+
+import general_superstaq as gss
+
+if TYPE_CHECKING:
+    from typing_extensions import TypeAlias
 
 TQuboKey = TypeVar("TQuboKey", bound=Hashable)
 
+ApiV2Like: TypeAlias = Literal[gss.superstaq_client.ApiVersion.V0_2_0, "v0.2.0"]
+ApiV3Like: TypeAlias = Literal[gss.superstaq_client.ApiVersion.V0_3_0, "v0.3.0"]
```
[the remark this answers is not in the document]
bharat-thotakura: Revised with https://github.com/Infleqtion/client-superstaq/pull/1383/commits/dd837c16a1c8268138a41ab877803b996795b148 & https://github.com/Infleqtion/client-superstaq/pull/1383/commits/ab056b3e7d5cb393bcc843b82e6c1c1ff3167221

cirq-superstaq/cirq_superstaq/service_test.py
```diff
@@ -103,11 +104,11 @@ def test_counts_to_results() -> None:
 
 def test_service_wrong_version() -> None:
     with pytest.raises(ValueError, match=r"`api_version` can only take value 'v0.2.0' or 'v0.3.0'"):
-        css.Service(api_version="v4")
+        css.Service(api_version="v0.1.0")  # type: ignore [call-overload]
```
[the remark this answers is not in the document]
bharat-thotakura: Fixed in https://github.com/Infleqtion/client-superstaq/pull/1383/commits/ebebd1c3ab6efb029097cff3dcbf3bc738f2742a

qiskit-superstaq/qiskit_superstaq/superstaq_provider.py
```diff
@@ -252,7 +322,7 @@ def aqt_compile(
         pulses: object = None,
         variables: object = None,
         **kwargs: Any,
-    ) -> qss.compiler_output.CompilerOutput:
+    ) -> qss.compiler_output.CompilerOutput | qss.SuperstaqJobV3:
```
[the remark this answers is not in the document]
bharat-thotakura: Corrected in https://github.com/Infleqtion/client-superstaq/pull/1383/commits/607b2f9b6130d3dcbb3c69cb57177aa466830b82

qiskit-superstaq/qiskit_superstaq/superstaq_backend.py
```diff
@@ -246,16 +249,17 @@ def compile(
         self,
         circuits: qiskit.QuantumCircuit | Sequence[qiskit.QuantumCircuit],
         **kwargs: Any,
-    ) -> qss.compiler_output.CompilerOutput:
+    ) -> qss.compiler_output.CompilerOutput | qss.SuperstaqJobV3:
         """Compiles the given circuit(s) to the backend's native gateset.
```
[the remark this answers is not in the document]
bharat-thotakura: Corrected in https://github.com/Infleqtion/client-superstaq/pull/1383/commits/607b2f9b6130d3dcbb3c69cb57177aa466830b82

qiskit-superstaq/qiskit_superstaq/superstaq_job.py
```diff
@@ -501,28 +501,30 @@ def _terminal_measurement_qubit_indices(self, index: int) -> list[int]:
     def result(
         self,
         index: int | None = None,
-        timeout: float | None = None,
-        wait: float = 5,
+        timeout_seconds: float = 7200,
+        polling_seconds: float = 1.0,
```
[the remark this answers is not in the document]
bharat-thotakura: Reverted in https://github.com/Infleqtion/client-superstaq/pull/1383/commits/c5dcf023047a1dd4a425cd728d6c5c3a6a60ad21

cirq-superstaq/cirq_superstaq/service_test.py
```diff
@@ -1037,6 +1038,101 @@ def test_service_ibmq_compile(mock_post: mock.MagicMock) -> None:
...
+)
+def test_service_compile_jobV3(
+    compile_call: Callable[[css.Service[css.JobV3], cirq.Circuit, str], css.JobV3],
+    target: str,
+) -> None:
+    service = css.Service(api_key="key", remote_host="http://example.com", api_version="v0.3.0")
+
+    mock_client = mock.MagicMock(spec=_SuperstaqClientV3)
+    mock_client.client_kwargs = {}
+    service._client = mock_client
```
bharat-thotakura: This mock is hiding away that `qscout_compile()` and `aqt_compile()` should technically not work due to deprecation

bharat-thotakura: Fixed by https://github.com/Infleqtion/client-superstaq/pull/1383/commits/2c5b1fd80dad48a77156807228c04e7537d3aec7

qiskit-superstaq/qiskit_superstaq/superstaq_backend.py
```diff
@@ -490,8 +520,11 @@ def qscout_compile(
             kwargs: Other desired `/qscout_compile` options.
 
         Returns:
-            Object whose .circuit(s) attribute contains optimized `qiskit.QuantumCircuit`(s), and
-            `.jaqal_program` attribute contains the corresponding Jaqal program(s).
+            For the v0.2.0 API, a `qss.CompilerOutput` object whose .circuit(s) attribute contains
+            optimized `qiskit.QuantumCircuit`(s), and`.jaqal_program` attribute contains the
+            corresponding Jaqal program(s). Otherwise (for the v0.3.0 API), an asynchronous
+            `qss.SuperstaqJobV3` on which compiled circuits can be queried via
+            `.compiled_circuits()`.
```
bharat-thotakura: Update docstring to mention `.jaqal_program()` can be queried on the job object

bharat-thotakura: Done in https://github.com/Infleqtion/client-superstaq/pull/1383/commits/ee333f96263d4175b5164e6595183e7239a65fb8

qiskit-superstaq/qiskit_superstaq/superstaq_provider.py
```diff
@@ -143,7 +178,35 @@ def __str__(self) -> str:
...
+             legacy behavior will be preserved and return a `qss.CompilerOutput`.
+
+        Raises:
+            TypeError: If `json_dict` is missing a job ID for the v0.3.0 API version.
+        """
+        if isinstance(self._client, gss.superstaq_client._SuperstaqClientV3):
+            job_id = json_dict.get("job_id")
+            if not isinstance(job_id, str):
+                raise TypeError("No valid job id was found in the compile request.")
+            return cast(
```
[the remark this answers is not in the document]
bharat-thotakura: Good catch, fixed in https://github.com/Infleqtion/client-superstaq/pull/1383/commits/fc35a881d42e59029059da35b6c7aed26a5e41ac

general-superstaq/general_superstaq/job.py
```diff
@@ -111,15 +111,16 @@ def wait_until_complete(
...
         time_waited: float = 0.0
         timeout_seconds = timeout_seconds or self._client.max_retry_seconds
         while (status := self._status(index)) not in gss.models.TERMINAL_CIRCUIT_STATES:
             if time_waited > timeout_seconds:
                 raise TimeoutError(
-                    f"Timed out while waiting for results. Final status was '{status}'"
+                    f"Timed out while waiting for results. Final status was '{status}'."
                 )
             time.sleep(polling_seconds)
             time_waited += polling_seconds
```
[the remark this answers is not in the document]
bharat-thotakura: That's a design choice independent of the goal of this PR. I think a solution would be to use:
```python
time_wait += polling_seconds or 5 # where '5' is the default if not specified
```
to avoid the indefinite loop till terminal.

For now, `polling_seconds = 0` should be taken to mean poll indefinitely till terminal

qiskit-superstaq/qiskit_superstaq/superstaq_job.py
```diff
@@ -501,28 +501,29 @@ def _terminal_measurement_qubit_indices(self, index: int) -> list[int]:
     def result(
         self,
         index: int | None = None,
-        timeout: float | None = None,
-        wait: float = 5,
+        timeout: float = 7200,
+        wait: float = 1.0,
         qubit_indices: Sequence[int] | None = None,
```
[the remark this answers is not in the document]
bharat-thotakura: Updated for parity in https://github.com/Infleqtion/client-superstaq/pull/1383/commits/9940aaea93d9767be9aefc11a82c2a2fa1f681db

qiskit-superstaq/qiskit_superstaq/superstaq_job.py
```diff
@@ -618,16 +624,26 @@ def compiled_circuits(self, index: int) -> qiskit.QuantumCircuit: ...
     def compiled_circuits(self, index: None = None) -> list[qiskit.QuantumCircuit]: ...
 
     def compiled_circuits(
-        self, index: int | None = None
+        self,
+        index: int | None = None,
+        *,
+        timeout: float | None = None,
+        wait: float = 5,
     ) -> qiskit.QuantumCircuit | list[qiskit.QuantumCircuit]:
```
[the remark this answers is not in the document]
bharat-thotakura: Good spot, updated in https://github.com/Infleqtion/client-superstaq/pull/1383/commits/ee1e1035833543e526b5863960bb9394fa67bfc2

cirq-superstaq/cirq_superstaq/job.py
```diff
@@ -542,44 +546,72 @@ def num_qubits(self, index: int | None = None) -> int | list[int]:
...
-        serialized_circuit = self.job_data.compiled_circuits[index]
-        if serialized_circuit is None:
+        if (serialized_circuit := self.job_data.compiled_circuits[index]) is None:
             raise gss.SuperstaqException(
                 f"Circuit {index} of job {self._job_id} does not have a compiled circuit."
             )
-
         return css.deserialize_circuits(serialized_circuit)[0]
 
+    def jaqal_program(
```
richrines1, line 596: could this go in gss.job?

cirq-superstaq/cirq_superstaq/job.py
```diff
@@ -542,44 +546,72 @@ def num_qubits(self, index: int | None = None) -> int | list[int]:
...
+        Raises:
+            gss.SuperstaqException: If the circuit at the provided `index` has no compiled circuit
+                or, when `index` is not specified, if any of the circuits in the job are missing a
+                compiled circuit.
         """
-        if index is None:
-            if all(c is None for c in self.job_data.compiled_circuits):
-                raise gss.SuperstaqException(f"The job {self._job_id} has no compiled circuits.")
+        self.wait_until_terminal_state(index, timeout_seconds, polling_seconds)
+        self._check_if_unsuccessful(index)
```
richrines1: I think we might want to loosen these - if e.g. we're submitting to hardware, the compiled circuits may be available long before the job is in a terminal state (and could even be available if the job eventually fails)

cirq-superstaq/cirq_superstaq/job.py
```diff
@@ -616,14 +652,20 @@ def initial_logical_to_physical(
...
+            polling_seconds: The time interval (in seconds) with which to poll job data retrieval.
+                Defaults to 1.0.
 
         Returns:
             A single logical to physical map (if `index` is passed) or list of maps for all input
             circuits.
         """
+        self.wait_until_terminal_state(index, timeout_seconds, polling_seconds)
         if index is None:
             return [self.initial_logical_to_physical(i) for i in range(self.job_data.num_circuits)]
```
richrines1: nit (avoids extra wait_for_terminal_state call if the index is None)
```suggestion
        if index is None:
            return [self.initial_logical_to_physical(i) for i in range(self.job_data.num_circuits)]

        self.wait_until_terminal_state(index, timeout_seconds, polling_seconds)
```

cirq-superstaq/cirq_superstaq/job.py
```diff
@@ -616,14 +652,20 @@ def initial_logical_to_physical(
...
 
+        # TODO: is it possible for a job failure to not populate this?
```
richrines1: i think this fails in the same conditions as compiled_circuits, e.g. if the job is still has status RECEIVED, AWAITING_COMPILE, or COMPILING, or fails during compilation

cirq-superstaq/cirq_superstaq/job.py
```diff
@@ -647,14 +693,20 @@ def final_logical_to_physical(
...
+            polling_seconds: The time interval (in seconds) with which to poll job data retrieval.
+                Defaults to 1.0.
 
         Returns:
             A single logical to physical map (if `index` is passed) or list of maps for all input
             circuits.
         """
+        self.wait_until_terminal_state(index, timeout_seconds, polling_seconds)
         if index is None:
             return [self.final_logical_to_physical(i) for i in range(self.job_data.num_circuits)]
```
richrines1:
```suggestion
        if index is None:
            return [self.final_logical_to_physical(i) for i in range(self.job_data.num_circuits)]

        self.wait_until_terminal_state(index, timeout_seconds, polling_seconds)
```

cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -142,12 +149,48 @@ class Service(gss.service.Service):
     `SUPERSTAQ_API_KEY`, or setting an API key in a configuration file.
     """
 
+    @overload
```
richrines1, line 152: this is pretty neat :)

general-superstaq/general_superstaq/typing.py
```diff
@@ -15,10 +15,18 @@
 from __future__ import annotations
 
 from collections.abc import Hashable
-from typing import Any, TypedDict, TypeVar
+from typing import TYPE_CHECKING, Any, Literal, TypedDict
 
 import pydantic
+from typing_extensions import TypeVar
```
richrines1: imo we should stick with importing this from `typing` where we don't need the `default` arg

cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -222,6 +265,32 @@ def __init__(
             **kwargs,
         )
 
+    def _map_compile_request_to_client_result(
+        self,
+        json_dict: dict[str, Any],
+        *,
+        legacy_parser: Callable[[dict[str, Any]], css.compiler_output.CompilerOutput],
```
richrines1, line 272: fwiw it's pretty redundant that we have `read_json_aqt` and `read_json_qscout` in addition to `read_json()` - it's be pretty easy to cover all three cases with one function (e.g. below), in which case it seems like we could avoid passing `legacy_parser` around/the corresponding casts/etc (instead we'd just need to pass `circuits_is_list` and `num_eca_circuits` to this function):
```python
def read_json(
    json_dict: dict[str, Any], circuits_is_list: bool, num_eca_circuits: int | None = None
) -> CompilerOutput:
    """Reads out returned JSON from Superstaq API's AQT compilation endpoint.

    Args:
        json_dict: A JSON dictionary matching the format returned by aqt_compile endpoint.
        circuits_is_list: A bool flag that controls whether the returned object has a .circuits
            attribute (if `True`) or a .circuit attribute (`False`).
        num_eca_circuits: Number of logically equivalent random circuits to generate for each
            input circuit.

    Returns:
        A `CompilerOutput` object with the compiled circuit(s). If `qtrl` is available locally,
        the returned object also stores the pulse sequence in the .seq attribute.
    """
    compiled_circuits: list[cirq.Circuit] | list[list[cirq.Circuit]]
    compiled_circuits = css.serialization.deserialize_circuits(json_dict["cirq_circuits"])

    initial_logical_to_physicals_list: list[dict[cirq.Qid, cirq.Qid]] = list(
        map(dict, cirq.read_json(json_text=json_dict["initial_logical_to_physicals"]))
    )
    initial_logical_to_physicals: (
        list[dict[cirq.Qid, cirq.Qid]] | list[list[dict[cirq.Qid, cirq.Qid]]]
    ) = initial_logical_to_physicals_list

    final_logical_to_physicals_list: list[dict[cirq.Qid, cirq.Qid]] = list(
        map(dict, cirq.read_json(json_text=json_dict["final_logical_to_physicals"]))
    )
    final_logical_to_physicals: (
        list[dict[cirq.Qid, cirq.Qid]] | list[list[dict[cirq.Qid, cirq.Qid]]]
    ) = final_logical_to_physicals_list

    jaqal_programs: list[str] | None = json_dict.get("jaqal_programs")

    serialized_pulse_gate_circuits = json_dict.get("pulse_gate_circuits")
    pulse_gate_circuits = (
        css.serialization.deserialize_qiskit_circuits(
            serialized_pulse_gate_circuits,
            circuits_is_list,
            pulse_start_times=json_dict.get("pulse_start_times"),
        )
        if serialized_pulse_gate_circuits
        else None
    )

    seq = None
    if "state_jp" in json_dict:
        if not importlib.util.find_spec("qtrl"):
            warnings.warn(
                "This output only contains compiled circuits. The `qtrl` package must be installed "
                "in order to deserialize compiled pulse sequences.",
                stacklevel=2,
            )
        else:  # pragma: no cover, b/c qtrl is not open source so it is not in cirq-superstaq reqs

            def _sequencer_from_state(state: dict[str, Any]) -> qtrl.sequencer.Sequence:
                seq = qtrl.sequencer.Sequence(n_elements=1)
                seq.__setstate__(state)
                seq.compile()
                return seq

            state = gss.serialization.deserialize(json_dict["state_jp"])

            if "readout_jp" in json_dict:
                readout_state = gss.serialization.deserialize(json_dict["readout_jp"])
                readout_seq = _sequencer_from_state(readout_state)

                if "readout_qubits" in json_dict:
                    readout_qubits = json.loads(json_dict["readout_qubits"])
                    readout_seq._readout = qtrl.sequence_utils.readout._ReadoutInfo(
                        readout_seq, readout_qubits, n_readouts=len(compiled_circuits)
                    )

                state["_readout"] = readout_seq

            seq = _sequencer_from_state(state)

    if num_eca_circuits is not None:
        compiled_circuits = [
            compiled_circuits[i : i + num_eca_circuits]
            for i in range(0, len(compiled_circuits), num_eca_circuits)
        ]
        initial_logical_to_physicals = [
            initial_logical_to_physicals_list[i : i + num_eca_circuits]
            for i in range(0, len(initial_logical_to_physicals_list), num_eca_circuits)
        ]
        final_logical_to_physicals = [
            final_logical_to_physicals_list[i : i + num_eca_circuits]
            for i in range(0, len(final_logical_to_physicals_list), num_eca_circuits)
        ]

        if jaqal_programs:
            jaqal_programs = [
                gss.compiler_output._jaqal_programs_to_subcircuits(
                    jaqal_programs[i : i + num_eca_circuits]
                )
                for i in range(0, len(jaqal_programs), num_eca_circuits)
            ]

        if pulse_gate_circuits:
            pulse_gate_circuits = [
                pulse_gate_circuits[i : i + num_eca_circuits]
                for i in range(0, len(pulse_gate_circuits), num_eca_circuits)
            ]

    if circuits_is_list:
        return CompilerOutput(
            compiled_circuits,
            initial_logical_to_physicals,
            final_logical_to_physicals,
            jaqal_programs=jaqal_programs,
            pulse_gate_circuits=pulse_gate_circuits,
            seq=seq,
        )

    return CompilerOutput(
        compiled_circuits[0],
        initial_logical_to_physicals[0],
        final_logical_to_physicals[0],
        jaqal_programs=jaqal_programs,
        pulse_gate_circuits=None if pulse_gate_circuits is None else pulse_gate_circuits[0],
        seq=seq,
    )
```

richrines1 (commented):
a few questions/comments but mostly looking good!

cirq-superstaq/cirq_superstaq/job_test.py
```diff
@@ -758,24 +753,20 @@ def test_job_counts_poll_timeout(
...
+    jobV3._job_data = gss.models.JobData(
+        **modifiy_job_result(job_dictV3, statuses=["running"], counts=[None])
+    )
 
-    with mock.patch("requests.Session.get", side_effect=[running_mock, running_mock]) as mock_get:
-        with pytest.raises(
-            TimeoutError, match=r"Timed out while waiting for results. Final status was 'running'"
-        ):
-            jobV3.counts(index=0, timeout_seconds=5, polling_seconds=10)
+    timeout_exc = TimeoutError("Timed out while waiting for results. Final status was 'running'.")
```
bharat-thotakura: Note: update test similar to https://github.com/Infleqtion/client-superstaq/pull/1396/changes/dfce3559a8f91be1a7a4af011128a4d481d564a2

general-superstaq/general_superstaq/typing.py
bharat-thotakura: Yeah I agree, I'm not sure why I decided to change it here anymore so reverted in https://github.com/Infleqtion/client-superstaq/pull/1383/commits/b8f745161ea22794c3438900a03d13ecc758c334

cirq-superstaq/cirq_superstaq/job.py
bharat-thotakura: Yeah, that's a fair point, the annoying thing is that the `css.Job` and `qss.SuperstaqJob` methods have different kwarg naming conventions. That is, `css.Job` uses `timeout_seconds` and `polling_seconds` and `qss.SuperstaqJob` uses `timeout` and `wait` (in parity with `qiskit` (as flagged in https://github.com/Infleqtion/client-superstaq/pull/1383#discussion_r3231390129). If it's factored into `gss.Job` the inherited `jaqal_program` will use different kwargs than the rest of `qss.SuperstaqJob` so we'd have to introduce a wrapper for naming consistency. At that point, I'm inclined to just keep duplicates since it's minimal lines of code and this function might not even be the final way to retrieve translated circuits. We could I guess rename everything to use `qiskit'`s naming convention, but I don't know if that breaks backwards compatibility for `cirq-superstaq`...

cirq-superstaq/cirq_superstaq/service.py
bharat-thotakura: Definitely, funnily enough, I do end up doing such a refactor for https://github.com/Infleqtion/client-superstaq/pull/1387: https://github.com/Infleqtion/client-superstaq/blob/d07733085c4c3dbea028534265383bfc2687a34b/cirq-superstaq/cirq_superstaq/service.py#L268-L304

Though, I don't go all the way like you're suggesting in consolidating everything into a `read_json` and instead tack on an extra identifier string label `parser`. Anticipating changes present in https://github.com/Infleqtion/client-superstaq/pull/1387, I'm thinking of doing this change there (or in a separate mini-PR) if that sounds good. Also, even with that refactoring, I wasn't able to get rid of the casting, though I'll look into it further after consolidation

cirq-superstaq/cirq_superstaq/job.py
bharat-thotakura: Thanks for the catch, I've added two types of terminal states to wait for (one is the pre-existing `TERMINAL_CIRCUIT_STATES` and a new `TERMINAL_COMPILE_STATES`) in https://github.com/Infleqtion/client-superstaq/pull/1383/commits/ad8f3b7f615ce1ce4558964f77f68896eaf97fde

cirq-superstaq/cirq_superstaq/job.py
bharat-thotakura: Ah got it, thanks! ~Is it worthwhile then to add a further warning/exception noting that job data is unavailable for `initial_logical_to_physical` or `final_logical_to_physical` if the job fails during compilation, since `CircuitStatus.FAILED` is considered a terminal state by the `self.wait_until_terminal_state()` guard?~ 

Just realized we have a `_check_if_unsuccessful()` helper for this so added it in https://github.com/Infleqtion/client-superstaq/pull/1383/commits/121cd2fce48f51fda9cb8fdcec6bc11e12b1fa49 to account for job failure cases

cirq-superstaq/cirq_superstaq/job_test.py
bharat-thotakura: Done in https://github.com/Infleqtion/client-superstaq/pull/1383/commits/025061b5eff0bd1d43f4d1de102609009090e1c6

cirq-superstaq/cirq_superstaq/service.py
richrines1: ah yeah maybe it doesn't avoid the casting

leaving for #1387/another pr is fine w/ me!

general-superstaq/general_superstaq/job.py
```diff
@@ -100,26 +100,35 @@ def _refresh_job(self) -> None:
...
         Args:
             index: An optional index of the specific circuit to wait for (otherwise waits for all
                 circuits to complete).
-            timeout_seconds: The total number of seconds to poll for.
-            polling_seconds: The interval with which to poll.
+            timeout_seconds: Optional total number of seconds to poll for. Otherwise, uses the
+                client instance's max retry seconds.
+            polling_seconds: The time interval with which to poll. Defaults to 5.
+            treat_as_submit_job: Boolean flag to consider terminal states corresponding to SUBMIT or
+                SIMULATE jobs. Defaults to `True`
```
richrines1: nit: this argument name seems a little confusing to me (in practice we'd still want to use this for submit jobs to check whether the compilation part is done) - wdyt about making it something like `compile_only` to indicate we are only waiting for compilation (defaulting to False)?

cirq-superstaq/cirq_superstaq/job.py
richrines1: we maybe don't want to use `_check_if_unsuccesful()` here - it's possible compilation succeeded but the job then failed during submission, in which case i think we should still be able to retrieve the compiled circuits. But maybe we could move it down to above the existing exception below, so we only check for errors if some of the compiled circuits are missing?

(ditto a few places below)

general-superstaq/general_superstaq/job.py
bharat-thotakura: Yeah, I realized afterwards that it could be read differently, so definitely open to names for clarity here (fun fact: this was the 2nd or 3rd name for it lol). Ended up going with `check_until_compile` (defaulting to `False`) with https://github.com/Infleqtion/client-superstaq/pull/1383/commits/dc33be67aca15a8dc409122c1a2e6919230184c4. Let me know if `compile_only` is clearer though

cirq-superstaq/cirq_superstaq/job.py
bharat-thotakura: I see, yeah, that's also a concern. I reverted that commit with https://github.com/Infleqtion/client-superstaq/pull/1383/commits/df117fee935bc7fa776a878e89f483566d84b771 & https://github.com/Infleqtion/client-superstaq/pull/1383/commits/c9a0fce4ad2098600637c0cb1ff1df84c3516ae3. I'm thinking of relegating better error handling to another PR after we use V3 more and identify pain points

cirq-superstaq/cirq_superstaq/job.py
richrines1: ah makes sense, i'm fine w/ the duplication for now. that said i think ultimately i'd be in favor of breaking backwards compatibility to make the args consistent across the two clients (for v3 only) - eventually i'd like to factor out even more of this logic to gss. my only hesitation is that i like the cirq names better lol

cirq-superstaq/cirq_superstaq/job.py
```diff
@@ -705,9 +766,12 @@ def counts(
             ~gss.SuperstaqServerException: If unable to get the results from the API.
             TimeoutError: If no results are available in the provided timeout interval.
             gss.SuperstaqException: If the job counts are missing.
+            NotImplementedError: The job does not admit retrieving result counts.
         """
-        self.wait_until_complete(index, timeout_seconds, polling_seconds)
+        if self.job_data.job_type in (gss.models.JobType.COMPILE, gss.models.JobType.CONVERT):
+            raise NotImplementedError("There are no counts to be retrieved for this job type.")
```
richrines1: i think this will conflict with counts set manually via `job.set_counts()` - maybe we need to check directly whether or not counts are present?

general-superstaq/general_superstaq/job.py
richrines1: hmm it still sounds a little funny to me - should it be `check_until_compiled`?

cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -612,8 +685,17 @@ def aqt_compile(
             }
 
         request_json["options"] = cirq.to_json(options_dict)
-        json_dict = self._client.post_request("/aqt_compile", request_json)
-        return css.compiler_output.read_json_aqt(json_dict, circuits_is_list, num_eca_circuits)
+        json_dict = (
+            self._client.aqt_compile(request_json)
+            if self._client.api_version == "v0.2.0"
+            else self._client.compile(request_json)
```
richrines1, line 691: after dropping v0.2.0, it'd be nice if we could just use e.g. `return self.create_job(..., method="compile")` here instead of having separate logic in `superstaq_client`

richrines1 (commented):
one more change necessary i think, then lgtm

general-superstaq/general_superstaq/job.py
bharat-thotakura: Yeah that's clearer, updated in https://github.com/Infleqtion/client-superstaq/pull/1383/commits/c28d0fa6d2457a1886aaed3a0b52edeefb3d68d9

cirq-superstaq/cirq_superstaq/job.py
bharat-thotakura: I see, yeah I forgot about retroactive job additions. I've changed it to check if it's `None` in https://github.com/Infleqtion/client-superstaq/pull/1383/commits/ae5ada377d027a44b1f89c72987370866486eecc

cirq-superstaq/cirq_superstaq/service.py
bharat-thotakura: Sure, should we make an issue for this, or do you recommend handling it in this PR?

richrines1: definitely not necessary as part of this pr - i think it'll be easier once we drop v0.2.0 anyway

richrines1 (approved):
lgtm, just need to check system repo compatibility before deploying

PR #1398, opened 2026-05-21
Include `lxml` dependency in `gss`'s `example-requirements.txt` for Sharpe ratio notebook

bharat-thotakura:


stephanielee9 (approved):
lgtm

Issue #1381, opened 2026-04-21
Use branch coverage in CI/CD


cdbf1:
### What is happening?

Currently we use line coverage in our CI/CD pipelines. I expect we would catch more bugs if we switched to using branch coverage (achieved using `checks/coverage_.py --branch`) which checks for any missing logical branches

dowusu-antwi:
From dev: @bharat-thotakura discussed possibly doing this on client but not on server (since it's slow), but would want to discuss more widely

bharat-thotakura:
^ On it being slow, this was from a while ago, where [using `sysmon` didn't help speed up](https://nedbatchelder.com/blog/202312/coveragepy_with_sysmonitoring#:~:text=Execution%20will%20be%20faster%20for%20line%20coverage%2C%20but%20not%20yet%20for%20branch%20coverage) branch coverage. Support has since been added, but there [were more steps involved](https://nedbatchelder.com/blog/202503/faster_branch_coverage_measurement#:~:text=I%20would%20love%20for%20people%20to%20try%20it%2C%20but%20it%E2%80%99s%20a%20little%20involved%20at%20the%20moment%3A), so it wasn't included in the CI yet (with local usage instead encouraged). Things may have changed now so we could look into it

PR #1401, opened 2026-06-04
update coverage_ test to branch instead of  line checks

daradarry:
Updating coverage_.py check to branch checks instead of line checks. Fixes issue #1381


dowusu-antwi:
From @richrines1 @ Dev meeting: remove the `--branch` option and run it locally with `general_superstaq` only, and make those changes first, then proceed similarly for each package in separate PRs.

PR #1403, opened 2026-06-16
Parallelize modular coverage tests

perlinm:
Modular coverage tests currently run one-by-one.  This PR makes them run in parallel.

A point of confusion: I don't see much benefit for the coverage [tests in this PR](https://github.com/Infleqtion/client-superstaq/actions/runs/27630018230) vs. that, say, [here](https://github.com/Infleqtion/client-superstaq/actions/runs/27291828714).  I definitely see a big benefit locally though...  Are github workflows limited to one CPU?

### DIFF
--- a/checks-superstaq/checks_superstaq/coverage_.py
+++ b/checks-superstaq/checks_superstaq/coverage_.py
@@ -15,6 +15,7 @@
 
 from __future__ import annotations
 
+import concurrent.futures
 import os
 import subprocess
 import sys
@@ -104,16 +105,17 @@ def run(
         return _report(test_returncode)
 
     # Run checks on individual files, skipping repeats
-    subprocess.check_call(["python", "-m", "coverage", "erase"])
+    subprocess.check_call([sys.executable, "-m", "coverage", "erase"], cwd=check_utils.root_dir)
 
     # Move test files to the end of the file list, so if both "x.py" and "x_test.py" are in `files`
     # both will be included in the coverage report
     files.sort(
         key=lambda file: file.endswith("_test.py") or os.path.basename(file).startswith("test_")
     )
-    coverage_args.append("--append")
-    test_returncode = 0
+    coverage_args.append("--parallel-mode")
 
+    # Build (files_requiring_coverage, test_files) pairs first, then run all concurrently.
+    pairs: list[tuple[list[str], list[str]]] = []
     while files:
         file = files.pop(0)
         test_files = check_utils.get_test_files([file], exclude=exclude, silent=True)
@@ -132,9 +134,24 @@ def run(
                 files_requiring_coverage.append(test_file)
                 files.remove(test_file)
 
-        test_returncode |= _run_on_files(
-            files_requiring_coverage, test_files, coverage_args, pytest_args
-        )
+        pairs.append((files_requiring_coverage, test_files))
+
+    test_returncode = 0
+    with concurrent.futures.ThreadPoolExecutor() as executor:
+        future_to_pair = {
+            executor.submit(
+                _run_on_files_capturing,
+                files_requiring_coverage,
+                test_files,
+                coverage_args,
+                pytest_args,
+            ): test_files
+            for files_requiring_coverage, test_files in pairs
+        }
+        for future in concurrent.futures.as_completed(future_to_pair):
+            returncode, output = future.result()
+            print(output, end="")  # noqa: T201
+            test_returncode |= returncode
 
     return _report(test_returncode)
 
@@ -147,8 +164,7 @@ def _run_on_files(
 ) -> int:
     """Helper function to run coverage tests on the given files with the given pytest arguments."""
     coverage_args = ["--include=" + ",".join(files_requiring_coverage), *coverage_args]
-
-    test_returncode = subprocess.call(
+    return subprocess.call(
         [
             sys.executable,
             "-m",
@@ -162,7 +178,35 @@ def _run_on_files(
         ],
         cwd=check_utils.root_dir,
     )
-    return test_returncode
+
+
+def _run_on_files_capturing(
+    files_requiring_coverage: list[str],
+    test_files: list[str],
+    coverage_args: list[str],
+    pytest_args: list[str],
+) -> tuple[int, str]:
+    """Like _run_on_files, but captures combined stdout/stderr for safe concurrent printing."""
+    coverage_args = ["--include=" + ",".join(files_requiring_coverage), *coverage_args]
+    result = subprocess.run(
+        [
+            sys.executable,
+            "-m",
+            "coverage",
+            "run",
+            *coverage_args,
+            "-m",
+            "pytest",
+            *test_files,
+            *pytest_args,
+        ],
+        check=False,
+        text=True,
+        cwd=check_utils.root_dir,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.STDOUT,
+    )
+    return result.returncode, result.stdout
 
 
 def _report(test_returncode: int) -> int:
--- a/checks-superstaq/checks_superstaq/coverage_.py
+++ b/checks-superstaq/checks_superstaq/coverage_.py
@@ -101,7 +101,7 @@ def run(
         return 0
 
     if not parsed_args.modular:
-        test_returncode = _run_on_files(files, test_files, coverage_args, pytest_args)
+        test_returncode, _ = _run_on_files(files, test_files, coverage_args, pytest_args)
         return _report(test_returncode)
 
     # Run checks on individual files, skipping repeats
@@ -140,7 +140,7 @@ def run(
     with concurrent.futures.ThreadPoolExecutor() as executor:
         future_to_pair = {
             executor.submit(
-                _run_on_files_and_capture_output,
+                _run_on_files,
                 files_requiring_coverage,
                 test_files,
                 coverage_args,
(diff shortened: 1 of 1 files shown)


checks-superstaq/checks_superstaq/coverage_.py
```diff
@@ -100,20 +101,30 @@ def run(
...
+) -> int:
+    """Run modular coverage checks concurrently, one (source, test) pair per subprocess."""
+    subprocess.check_call([sys.executable, "-m", "coverage", "erase"], cwd=check_utils.root_dir)
+
+    # Move test files to the end of the file list, so if both "x.py" and "x_test.py" are in
+    # `files` both will be included in the coverage report.
     files.sort(
         key=lambda file: file.endswith("_test.py") or os.path.basename(file).startswith("test_")
     )
-    coverage_args.append("--append")
```
richrines1, line 114: i think this argument is necessary to run parallel (but not modular) coverage

dowusu-antwi:
Dev meeting comments:
- Coverage output is being captured somewhere -- can we add pack the prints (e.g., '.' for every passed check, etc.)
- Parallel mode is on by default (@bharat-thotakura ran it locally without that and it ran with it)
- We want to make sure that this is sufficiently reconfigurable so that it's not too restrictive

checks-superstaq/checks_superstaq/coverage_.py
perlinm: What do you mean?  This line is only reached in modular mode, since otherwise the check terminates above at
```python
if not parsed_args.modular:
    test_returncode, _ = _run_on_files(files, test_files, coverage_args, pytest_args)
    return test_returncode
```

perlinm: Do you mean "modular but not parallel"?

perlinm:
@dowusu-antwi I think I reverted to the same default behavior as before this PR, and added a `-j | --jobs` argument.  So now updates are printed as usual with `--jobs 1` (the default).  Passing `--jobs 0` uses the default number of workers chosen by `concurrent.futures.ThreadPoolExecutor`.

perlinm:
Note: the type check seems to be failing on `supermarq-benchmarks/supermarq/plotting.py`, which is not touched by this PR.

checks-superstaq/checks_superstaq/coverage_.py
perlinm: I now add `--append` when `num_jobs == 1`, and `--parallel-mode` otherwise.

dowusu-antwi:
> Note: the type check seems to be failing on `supermarq-benchmarks/supermarq/plotting.py`, which is not touched by this PR.

This should be fixed by #1405.

perlinm:
Any other requested changes?

checks-superstaq/checks_superstaq/coverage_.py
```diff
@@ -132,23 +153,49 @@ def run(
...
+    if num_jobs == 1:
+        # Run modular checks one at a time
+        coverage_args.append("--append")
+        for files_requiring_coverage, test_files in pairs:
+            result = _run_on_files(files_requiring_coverage, test_files, coverage_args, pytest_args)
+            test_returncode |= result.returncode
+        return test_returncode
+
+    # Run modular checks concurrently
+    coverage_args.append("--parallel-mode")
```
richrines1: this is already set by default via `pyproject.toml`
```suggestion
```

dowusu-antwi:
I think Rich made a live comment at today's dev meeting about the `--job 0` case, but maybe that was in reference to his earlier review. @richrines1 is that the case?

checks-superstaq/checks_superstaq/coverage_.py
richrines1: ah my bad, my original comment was mistaken. "parallel-mode" is already set by default via pyproject.toml, which afaict means we shouldn't need/want `--append` to be passed at all. so i think you had it right before.

checks-superstaq/checks_superstaq/coverage_.py
```diff
@@ -100,20 +111,30 @@ def run(
         return 0
 
     if not parsed_args.modular:
-        test_returncode = _run_on_files(files, test_files, coverage_args, pytest_args)
-        return _report(test_returncode)
+        result = _run_on_files(files, test_files, coverage_args, pytest_args)
+        return result.returncode
```
richrines1: still need to run `_report` in this case
```suggestion
        result = _run_on_files(files, test_files, coverage_args, pytest_args)
        return _report(result.returncode)
```

checks-superstaq/checks_superstaq/coverage_.py
```diff
@@ -132,23 +153,49 @@ def run(
...
                 files.remove(test_file)
 
-        test_returncode |= _run_on_files(
-            files_requiring_coverage, test_files, coverage_args, pytest_args
-        )
+        pairs.append((files_requiring_coverage, test_files))
 
-    return _report(test_returncode)
+    test_returncode = 0
+    if num_jobs == 1:
```
richrines1: for consistency with xdist (see above)
```suggestion
    if num_jobs == 0:
```

checks-superstaq/checks_superstaq/coverage_.py
```diff
@@ -48,6 +49,16 @@ def run(
         action="store_true",
         help="Check that each file is covered by its own test file.",
     )
+    parser.add_argument(
+        "-j",
+        "--jobs",
+        type=int,
+        default=1,
```
richrines1: currently we use pytest-xdist to enable parallelism in non-modular mode. it would be nice if we could make this option more consistent with the way that works:
* `-j 0` should disable parallelism
* `-j 1` should enable "parallel" mode but with a single thread (which can be useful for debugging)
* if no `-j` flag is passed, we should default to something sensible like `min((os.cpu_count() or 2) // 2, len(files))` (the ThreadPoolExecutor default is way too many to be useful for something that isn't io-limited)

in this case imo this should just default to None (which we can interpret as "default" downstream)
```suggestion
        default=None,
```

checks-superstaq/checks_superstaq/coverage_.py
richrines1: we can pass the jobs flag to xdist in non-modular mode so the same flag works in both cases
```suggestion
        parsed_args.jobs
        if not parsed_args.modular and parsed_args.jobs is not None
        else "auto"
```

checks-superstaq/checks_superstaq/coverage_.py
```diff
@@ -132,23 +153,49 @@ def run(
...
-        test_returncode |= _run_on_files(
-            files_requiring_coverage, test_files, coverage_args, pytest_args
-        )
+        pairs.append((files_requiring_coverage, test_files))
 
-    return _report(test_returncode)
+    test_returncode = 0
+    if num_jobs == 1:
+        # Run modular checks one at a time
+        coverage_args.append("--append")
```
richrines1: (as per thread)
```suggestion
```

checks-superstaq/checks_superstaq/coverage_.py
```diff
@@ -132,23 +153,49 @@ def run(
...
+        for files_requiring_coverage, test_files in pairs:
+            result = _run_on_files(files_requiring_coverage, test_files, coverage_args, pytest_args)
+            test_returncode |= result.returncode
+        return test_returncode
+
+    # Run modular checks concurrently
+    coverage_args.append("--parallel-mode")
+    with concurrent.futures.ThreadPoolExecutor(max_workers=num_jobs or None) as executor:
```
richrines1: (more sensible default)
```suggestion
    max_workers = num_jobs or min((os.cpu_count() or 2) // 2, len(test_files))
    with concurrent.futures.ThreadPoolExecutor(max_workers=max_workers) as executor:
```

checks-superstaq/checks_superstaq/coverage_.py
perlinm: Done

checks-superstaq/checks_superstaq/coverage_.py
perlinm: Done

checks-superstaq/checks_superstaq/coverage_.py
perlinm: Done

checks-superstaq/checks_superstaq/coverage_.py
perlinm: Done

checks-superstaq/checks_superstaq/coverage_.py
perlinm: Done

checks-superstaq/checks_superstaq/coverage_.py
perlinm: Done

checks-superstaq/checks_superstaq/coverage_.py
perlinm: Does this not need to first change the number of jobs from `None` to `min((os.cpu_count() or 2) // 2, len(test_files))` in this case?

perlinm: Oh nevermind, I see it passes `"auto"` in that case

richrines1 (approved):
> Are github workflows limited to one CPU?

i think they give you two? but would need to confirm.. (also fwiw they also might not be picked up correctly by e.g. `os.cpu_count()` so probably a good idea to pass `-j<n>` directly in CI scripts)

richrines1:
lgtm!

perlinm:
Thanks for the review @richrines1!

PR #1404, opened 2026-06-23
Update copyright licensee to Infleqtion, Inc. everywhere

bharat-thotakura:


bharat-thotakura:
(type check to be fixed by #1405)

dowusu-antwi (approved):

PR #1405, opened 2026-06-23
Update `matplotlib` related typing and import style

bharat-thotakura:
Fixes recently [failing type check](https://github.com/Infleqtion/client-superstaq/actions/runs/28055515228/job/83056762405) and conforms some `mpl` imports to `ty` and Superstaq dev guide

### DIFF
--- a/supermarq-benchmarks/supermarq/plotting.py
+++ b/supermarq-benchmarks/supermarq/plotting.py
@@ -347,11 +347,11 @@ def annotate_heatmap(
     Returns:
         List of the text annotations.
     """
-    if data is None:
-        data = np.asarray(im.get_array(), dtype=np.float64)
-    else:
-        data = np.asarray(data, dtype=np.float64)
-
+    data = (
+        np.asarray(im.get_array(), dtype=np.float64)
+        if data is None
+        else np.asarray(data, dtype=np.float64)
+    )
     normalized_data = np.asarray(im.norm(data), dtype=np.float64)
 
     def _norm_scalar_val(value: float) -> float:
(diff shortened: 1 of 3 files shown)


supermarq-benchmarks/supermarq/plotting.py
```diff
@@ -18,12 +18,16 @@
 from typing import TYPE_CHECKING, Any
 
 import matplotlib as mpl
+import matplotlib.axes
```
bharat-thotakura, line 21: Note: These specific imports were added due to `ty` warnings such as:
```bash
Submodule `axes` might not have been imported
help: Consider explicitly importing `matplotlib.axes (ty possibly-missing-submodule)`
```

supermarq-benchmarks/supermarq/plotting.py
```diff
@@ -341,11 +345,24 @@ def annotate_heatmap(
...
+            np.asarray(im.norm(np.asarray([value], dtype=np.float64)), dtype=np.float64).item()
+        )
 
     # Normalize the threshold to the images color range.
-    threshold = im.norm(threshold) if threshold is not None else im.norm(data.max()) / 2.0
+    normalized_threshold: float = (
+        _norm_scalar_val(threshold)
+        if threshold is not None
+        else _norm_scalar_val(float(np.max(data))) / 2.0
+    )
```
richrines1: this seems to pass for me, with (somewhat) less casting:
```suggestion
    # Normalize the threshold to the images color range.
    normalized_threshold = im.norm(np.max(data) / 2 if threshold is None else np.asarray(threshold))
```

richrines1 (approved):
one (possible) suggestion, otherwise lgtm!

supermarq-benchmarks/supermarq/plotting.py
bharat-thotakura: TIL that Github won't let me apply this suggestion due to the following:
> Applying suggestions on deleted lines is currently not supported. 

But updated in https://github.com/Infleqtion/client-superstaq/pull/1405/commits/bf1433ebc01c7bc24208927e0caefe9d31be1039

PR #1412, opened 2026-07-01
Adding unit tests for full branch coverage of general superstaq

daradarry:
Including full branch for coverage for `general superstaq`: Part of [#1381](https://github.com/Infleqtion/client-superstaq/issues/1381) .


daradarry:
@bharat-thotakura, I guess allowing the coverage test here fixes the other "pytest and coverage check" here. So, this should be good to go right?

checks/coverage_.py
```diff
@@ -21,6 +21,7 @@
 
 if __name__ == "__main__":
     args = [*sys.argv[1:]]
+    args += ["--branch"]
```
bharat-thotakura: @daradarry, yes, the checks should pass after you remove this line (we'll only include it _after_ all the branch coverage PRs are merged in

The plan is to manually pass `--branch` locally to test this PR

cirq-superstaq/cirq_superstaq/ops/qudit_gates_test.py
```diff
@@ -234,8 +234,6 @@ def test_movement_op() -> None:
     op = css.movement_op({q2: q3, q3: q4})
     assert op.gate == css.MovementGate({0: 1, 1: 2})
     assert op == css.MovementGate({0: 1, 1: 2}).on(q2, q3, q4)
-    assert op == op**1
-    assert op**-1 == css.movement_op({q3: q2, q4: q3})
```
bharat-thotakura, line 238: Also, is there a reason to delete these assertions? I think they're good to have

daradarry: they were producing a lint error:
`error: Unsupported operand types
for ** ("Operation" and "int")  [operator]
        assert op == op**1`

Seems extra lint requirement was added in a recent branch.

bharat-thotakura: I don't believe it's a linting issue, so it's coming from `mypy`. However, I don't seem to get it locally, so can you revert this deletion to see if that error shows up in the CI checks?

general-superstaq/general_superstaq/job_test.py
```diff
@@ -308,6 +307,12 @@ def test_update_status_queue_info(mock_client: gss.superstaq_client._SuperstaqCl
     job._update_status_queue_info()
     assert job._overall_status == "failed"
 
+    # tests the 'failed' status branch option for test_update_status_queue_info in job.py
```
bharat-thotakura: µ-nit: [Superstaq dev guide has a style for writing comments](https://superstaq.readthedocs.io/en/latest/resources/developer_guide.html#inline-comments). So, a better practice here (& elsewhere) would be:
```suggestion
    # Tests the 'failed' status branch option for `test_update_status_queue_info()` in `job.py`
```

checks/coverage_.py
bharat-thotakura: Resolved by https://github.com/Infleqtion/client-superstaq/pull/1412/commits/c12632aa5d165bc03ac771d19b14312b69ed7b27

general-superstaq/general_superstaq/service_test.py
```diff
@@ -527,5 +527,20 @@ def test_aces(
...
+    # tests the 'false' noise branch option for submit_aces in service.py
+    assert (
+        service.submit_aces(
+            target="ss_unconstrained_simulator",
+            qubits=[0, 1],
+            shots=100,
+            num_circuits=10,
+            mirror_depth=5,
+            extra_depth=5,
+            noise="",
```
bharat-thotakura: Just to understand this better, is this testing for the case when `noise=None` or the case where `noise` is given a (falsy, empty) `str` that is not in the valid set: `("symmetric_depolarize", "phase_flip", "bit_flip", "asymmetric_depolarize")`? If the latter, we'd ideally want to throw an error for invalid string options, but `submit_aces` is somewhat deprecated, so that might be out of scope for this PR

general-superstaq/general_superstaq/superstaq_client_test.py
```diff
@@ -1315,6 +1315,80 @@ def test_superstaq_client_compile_v3(
     )
     assert job_json == {"job_id": str(job_id)}
 
+    job_id = uuid.UUID(int=0)
+    mock_post.return_value.json.return_value = {"job_id": job_id, "num_circuits": 1}
+
+    client_v3.compile({"cirq_circuits": "Hello", "target": "ss_example_qpu"})
+
+    client_v3.compile({"qiskit_circuits": "Hello", "target": "ss_example_qpu"})
```
bharat-thotakura: Could you say more about what these lines are for? The `mock_post` isn't invoked for the rest of the function, for example? Locally, I was able to achieve full branch coverage just by deleting them:
```suggestion
```

general-superstaq/general_superstaq/superstaq_client_test.py
```diff
@@ -1315,6 +1315,80 @@ def test_superstaq_client_compile_v3(
...
+
+    with pytest.raises(RuntimeError, match=r"No recognized circuits found."):
+        client_v3.compile({"unknown_circuits": "Hello", "target": "ss_example_qpu"})
+
+
+@mock.patch("requests.Session.get")
+@mock.patch("requests.Session.post")
+def test_superstaq_client_compile_v3_with_wait(
```
bharat-thotakura: Similar to this [comment](https://github.com/Infleqtion/client-superstaq/pull/1412/changes#r3761401770), it looks like some updates from `main` may have covered this case. That is, is this test doing something not already covered? Locally, I seem to maintain full branch coverage without it

general-superstaq/general_superstaq/superstaq_client_test.py
```diff
@@ -1315,6 +1315,80 @@ def test_superstaq_client_compile_v3(
...
 
+    job_id = uuid.UUID(int=0)
+    mock_post.return_value.json.return_value = {"job_id": job_id, "num_circuits": 1}
+
+    client_v3.compile({"cirq_circuits": "Hello", "target": "ss_example_qpu"})
+
+    client_v3.compile({"qiskit_circuits": "Hello", "target": "ss_example_qpu"})
+
+    with pytest.raises(RuntimeError, match=r"No recognized circuits found."):
+        client_v3.compile({"unknown_circuits": "Hello", "target": "ss_example_qpu"})
```
bharat-thotakura: Looks like this error checking is already covered by:
https://github.com/Infleqtion/client-superstaq/blob/1fcb8c6da25b4b90cbaaaf19cb8bc0e042b950b5/general-superstaq/general_superstaq/superstaq_client_test.py#L1269-L1273
so it might be redundant:
```suggestion
```

general-superstaq/general_superstaq/service_test.py
daradarry: You are right, I only needed to set `Noise=None`. I am not sure of what to do with the deprecated `submit_aces` though. Should it be removed from `service.py` ?

bharat-thotakura: I think the plan for it hasn't been decided, so you can ignore it for now

daradarry: Okay

general-superstaq/general_superstaq/superstaq_client_test.py
daradarry: Yes, I removed the redundancies.

bharat-thotakura (approved):
LGTM

PR #1413, opened 2026-07-01
Adding unit tests for full branch coverage of qiskit superstaq

daradarry:
Including full branch for coverage for `qiskit superstaq`: Part of #[1381](https://github.com/Infleqtion/client-superstaq/issues/1381)  .


qiskit-superstaq/qiskit_superstaq/serialization.py
```diff
@@ -104,7 +104,7 @@ def _mcphase(
 }
 
 
-if hasattr(qiskit.circuit.library, "QFTGate"):
+if hasattr(qiskit.circuit.library, "QFTGate"):  # pragma: no branch
```
daradarry: Do we still need this part?

qiskit-superstaq/qiskit_superstaq/serialization.py
```diff
@@ -382,7 +382,7 @@ def _resolve_gate(gate: qiskit.circuit.Instruction) -> qiskit.circuit.Instructio
     if type(gate) is qiskit.circuit.ControlledGate:
         gate.base_gate = _resolve_gate(gate.base_gate)
 
-        if gate.definition and gate._definition:
+        if gate.definition and gate._definition:  # pragma: no branch
```
daradarry: ..not sure of how to test the branch here ..seems qiskit do not use these kind of gates again.

qiskit-superstaq/qiskit_superstaq/serialization.py
```diff
@@ -394,7 +394,7 @@ def _resolve_gate(gate: qiskit.circuit.Instruction) -> qiskit.circuit.Instructio
             if (
                 trial_gate.definition == gate.definition
                 or trial_gate._definition == gate._definition
-            ):
+            ):  # pragma: no branch
```
daradarry: ...same as above

qiskit-superstaq/qiskit_superstaq/serialization.py
richrines1: nope :) looks like `QFTGate` was added in qiskit 1.2.0, and we already require qiskit>=1.3.0

we can delete this block and instead add 
```python
    qiskit.circuit.library.QFTGate: lambda gate: qiskit.circuit.library.QFTGate(
        gate.num_qubits - gate.num_ctrl_qubits
    ).control(
        num_ctrl_qubits=gate.num_ctrl_qubits,
        ctrl_state=gate.ctrl_state,
        label=gate.label,
    ),
```
to `_controlled_gate_resolvers` above.

we should also update a few corresponding tests:
* drop `test_qft_gate`
* drop the `if hasattr...` condition above in `test_nonstandard_static_gate_serialization` (`qiskit.circuit.library.QFTGate(4)`)

qiskit-superstaq/qiskit_superstaq/serialization.py
richrines1: this one is here to catch a weird corner case: someone can define a controlled version of one of the gates in `_controlled_gate_resolvers`, but give it an unusual (or incorrect) definition. in this case its definition would be different from that of `trial_gate`, and so returning the latter would be incorrect

something like this should cover the branch:
```
    gate = qiskit.circuit.ControlledGate(
        "cgate",
        num_qubits=3,
        params=[1.23],
        base_gate=qiskit.circuit.library.PhaseGate(1.23),
        definition=qiskit.circuit.library.CCXGate().definition,
    )
    _check_serialization(gate)
```
(which works because `qiskit.circuit.library.PhaseGate` is in `_controlled_gate_resolvers` but here we're giving it a different (wrong) definition)

btw this is maybe a good example of why branch coverage is really important - apparently we wrote code to handle this corner case but weren't testing it at all :)

qiskit-superstaq/qiskit_superstaq/serialization.py
richrines1: to cover this we need a controlled gate without a `definition`, which we can construct like this:
```python
    gate = qiskit.circuit.ControlledGate(
        "cgate",
        num_qubits=3,
        params=[],
        base_gate=qiskit.circuit.library.YGate(),
    )
    assert qss.serialization._resolve_gate(gate) == gate
```
however, it turns out such a gate will cause the serialization pipeline to fail:
```python
circuit = qiskit.QuantumCircuit(3)
circuit.append(gate)
_ = qss.serialize_circuits(circuit)  # AttributeError: 'NoneType' object has no attribute 'metadata'
```
we can add a hacky workaround for this in `_prepare_gate()` (see [comment](https://github.com/Infleqtion/client-superstaq/pull/1413/changes#r3565032154)), after which we can test both using
```python
_check_serialization(gate)
```
with the same gate as above

qiskit-superstaq/qiskit_superstaq/serialization.py
richrines1, line 335: workaround for issue in https://github.com/Infleqtion/client-superstaq/pull/1413/changes#r3565030447:
```suggestion
        definition = (
            qiskit.QuantumCircuit()
            if gate._definition is None
            else _prepare_circuit(gate._definition)
        )
        return qiskit.circuit.ControlledGate(
            name=gate._name,
            num_qubits=gate.num_qubits,
            params=gate.params,
            label=gate.label,
            num_ctrl_qubits=gate.num_ctrl_qubits,
            definition=definition,
            ctrl_state=gate.ctrl_state,
            base_gate=_prepare_gate(gate.base_gate),
        )
```

qiskit-superstaq/qiskit_superstaq/serialization.py
daradarry: Ok, will do

qiskit-superstaq/qiskit_superstaq/serialization.py
daradarry: Makes sense now. Thanks.

qiskit-superstaq/qiskit_superstaq/serialization.py
daradarry: @richrines1, I see that function `result` was defined twice in  'qiskit_superstaq/superstaq_job'  (lines 127 and 508) . Is there a reason for that?

richrines1: there are two different job classes in that module: `SuperstaqJob` and `SuperstaqJobV3` (where the former is returned when using the old v0.2.0 server and the latter for the v0.3.0 server). They use different data structures to store job data so each has to define its own `result` method

daradarry: .. I skipped that. 

Okay,  I need to work on  `test_resultV3` function for that of `SuperstaqJobV3` then (I've been editing the wrong one). 
Thanks.

qiskit-superstaq/qiskit_superstaq/superstaq_backend.py
```diff
@@ -260,7 +260,7 @@ def compile(
         Raises:
             ValueError: If this backend does not support compilation.
         """
-        if self._provider._client.api_version == "v0.2.0":
+        if self._provider._client.api_version == "v0.2.0":  # pragma: no branch
```
daradarry: Is there another version `qss.SuperstaqProvider` that will satisfy this branch condition? I tried "v0.3.0", but ran into Pydantic errors.

richrines1: "v0.2.0" and "v0.3.0" are the only supported versions (without a bunch of hacky mocking at least)

do you have an example of the pydantic errors? it \*should\* work with both versions, so if it's not there might be a bug somewhere we have to fix..

qiskit-superstaq/qiskit_superstaq/superstaq_job.py
```diff
@@ -586,7 +593,7 @@ def combined_result(
         result = self.result(timeout=timeout, wait=wait, qubit_indices=qubit_indices)
         counts = result.get_counts()
 
-        if not isinstance(counts, Mapping):
+        if not isinstance(counts, Mapping):  # pragma: no branch
```
daradarry: Here, line #593 always return a dict (either empty or not, which is a mapping).

qiskit-superstaq/qiskit_superstaq/superstaq_backend.py
daradarry: I get this error with "v0.3.0":

E       pydantic_core._pydantic_core.ValidationError: 2 validation errors for NewJobResponse
E       job_id
E         Field required [type=missing, input_value={'qiskit_circuits': 'UUlT...hysicals': '[[[0, 0]]]'}, input_type=dict]
E           For further information visit https://errors.pydantic.dev/2.13/v/missing
E       num_circuits
E         Field required [type=missing, input_value={'qiskit_circuits': 'UUlT...hysicals': '[[[0, 0]]]'}, input_type=dict]
E           For further information visit https://errors.pydantic.dev/2.13/v/missing

../general-superstaq/general_superstaq/superstaq_client.py:1521: ValidationError

qiskit-superstaq/qiskit_superstaq/superstaq_job.py
richrines1: not always: if the job contains multiple circuits, `result.get_counts()` will instead be a list of dicts. for example:
```python
backend = provider.get_backend("ss_unconstrained_simulator")

circuit = qiskit.QuantumCircuit(2)
circuit.measure_all()

job = backend.run([circuit, circuit], shots=10)
print(job.result().get_counts())  # [{'00': 10}, {'00': 10}]
```
(fwiw most superstaq client methods are supposed to support either a single circuit or a list of circuits)

it looks like the multi-circuit is already covered by the `multi_job.combined_counts()` calls in `superstaq_job_test.py::test_resultV3`, so to cover this branch we'd just need to do the same for a single-circuit `job`, e.g. by adding something like this
```python
        assert job.combined_result().get_counts() == job.result().get_counts()
        assert job.combined_result().to_dict() == job.result().to_dict()
```
to the first `with patched_requests(...)` block in test_resultV3

qiskit-superstaq/qiskit_superstaq/superstaq_backend.py
richrines1: ah this might just be a v2/v3 mocking issue. can you show the code you added that caused the error?

(sorry the v2/v3 stuff is super confusing, hopefully we'll be able to get rid of v2 altogether soon :)

daradarry: Here is the code block (Similar to the one in  `test_compile` in `qiskit_superstaq/superstaq_backend_test.py`):

    provider = qss.SuperstaqProvider(api_key="MY_TOKEN", api_version="v0.3.0")
    backend = provider.get_backend("aws_sv1_simulator")

    qc = qiskit.QuantumCircuit(1)
    qc.h(0)
    mock_post.return_value.json = lambda: {
        "qiskit_circuits": qss.serialization.serialize_circuits([qc]),
        "initial_logical_to_physicals": json.dumps([[(0, 0)]]),
        "final_logical_to_physicals": json.dumps([[(0, 0)]]),
    }
    out = backend.compile([qc], test_options="yes")

richrines1: ok so the problem is that the v3 client makes different requests/expects different responses, which don't align with that mocked return value. in particular the v3 client uses two different requests here: a POST to submit the circuit for compilation, and then a GET to retrieve the result (this difference was a major reason for building v3 - in v2 the original POST stays open until the compilation is done)

unfortunately v3 requests also expect much more detailed responses so the tests become a lot more verbose. e.g. i think this should work:
```python
@patch("requests.Session.get")
@patch("requests.Session.post")
def test_compileV3(mock_post: MagicMock, mock_get: MagicMock) -> None:
    provider = qss.SuperstaqProvider(api_key="MY_TOKEN", api_version="v0.3.0")
    backend = provider.get_backend("cq_sqale_simulator")

    qc = qiskit.QuantumCircuit(1)
    qc.h(0)

    fake_job_id = str(uuid.uuid4())
    mock_post.return_value.json = lambda: gss.models.NewJobResponse(
        job_id=fake_job_id, num_circuits=1
    ).model_dump(mode="json")

    fake_job_data = gss.models.JobData(
        job_type="compile",
        statuses=["completed"],
        status_messages=[],
        user_email="test@email.com",
        target="cq_sqale_simulator",
        provider_id=[fake_job_id],
        num_circuits=1,
        compiled_circuits=[qss.serialize_circuits(qc)],
        input_circuits=[qss.serialize_circuits(qc)],
        circuit_type="qiskit",
        counts=[None],
        results_dicts=[None],
        shots=[0],
        dry_run=True,
        submission_timestamp=datetime.datetime.now(tz=datetime.timezone.utc),
        last_updated_timestamp=[datetime.datetime.now(tz=datetime.timezone.utc)],
        initial_logical_to_physicals=[{0: 0}],
        final_logical_to_physicals=[{0: 0}],
        logical_qubits=["0", "1"],
        physical_qubits=["0", "1"],
    )
    mock_get.return_value.json = lambda: {
        fake_job_id: fake_job_data.model_dump(mode="json")
    }
    out = backend.compile([qc], test_options="yes")
    assert out.circuits == [qc]
    assert out.initial_logical_to_physicals == [{0: 0}]
    assert out.final_logical_to_physicals == [{0: 0}]
```

qiskit-superstaq/qiskit_superstaq/superstaq_job.py
daradarry: It worked..Thanks

checks/coverage_.py
```diff
@@ -21,6 +21,7 @@
 
 if __name__ == "__main__":
     args = [*sys.argv[1:]]
+    args += ["--branch"]
```
richrines1: we can add this once everything passes; for now it's sufficient to check that `./checks/coverage_.py --branch cirq-superstaq` passes locally
```suggestion
```

qiskit-superstaq/qiskit_superstaq/serialization.py
```diff
@@ -104,7 +104,7 @@ def _mcphase(
 }
 
 
-if hasattr(qiskit.circuit.library, "QFTGate"):
+if hasattr(qiskit.circuit.library, "QFTGate"):  # pragma: no cover
```
richrines1: can as described in our conversation?

richrines1 (commented):
can we include all the changes discussed in the comments to get rid of the pragmas?

daradarry:
This should be fine fine now. Only the coverage tests are failing.

qiskit-superstaq/qiskit_superstaq/superstaq_job.py
```diff
@@ -198,6 +198,7 @@ def _check_if_stopped(self) -> None:
             raise gss.superstaq_exceptions.SuperstaqUnsuccessfulJobException(
                 self._job_id, self._overall_status
             )
+        raise gss.superstaq_exceptions.SuperstaqUnsuccessfulJobException(self._job_id, "Running")
```
richrines1: we don't want to actually raise an error in this case
```suggestion
```

qiskit-superstaq/qiskit_superstaq/superstaq_job.py
```diff
@@ -257,6 +258,7 @@ def _update_status_queue_info(self) -> None:
             if temp_status in status_occurrence:
                 self._overall_status = temp_status
                 return
+            self._overall_status = "Not Queued"
```
richrines1: this should always use one of the statuses recognized by `SuperstaqJob.status`

if this occurs, the job either has no circuit data, or they all have unrecognized statuses. in the first case i think we can call this "done". in the latter it's probably an error.
```suggestion
            if not status_occurrence:  # No circuits, so nothing to do
                self._overall_status = "Done"
            self._overall_status = "Failed"
```

qiskit-superstaq/qiskit_superstaq/serialization.py
richrines1: can we apply [this change](https://github.com/Infleqtion/client-superstaq/pull/1413/changes#r3565032154) in this pr as well?

qiskit-superstaq/qiskit_superstaq/superstaq_job_test.py
```diff
@@ -537,6 +570,12 @@ def test_check_if_stopped(backend: qss.SuperstaqBackend) -> None:
         with pytest.raises(gss.SuperstaqUnsuccessfulJobException, match=status):
             job._check_if_stopped()
 
+        # test to  include `_check_if_stopped"!=("Cancelled", "Failed")` branch for ` in superstaq
+        job = qss.SuperstaqJob(backend=backend, job_id="123abc")
+        job._overall_status = "Running"
+        with pytest.raises(gss.SuperstaqUnsuccessfulJobException, match="Running"):
+            job._check_if_stopped()
```
richrines1: to test without adding the exception
```suggestion
        assert job._check_if_stopped() is None
```

qiskit-superstaq/qiskit_superstaq/superstaq_job.py
daradarry: I guess, it was one of the first few cases I attempted when I wasn't sure about testing the coverage branch :) . It just returns `None` if job is neither 'Canceled' nor 'Failed'

qiskit-superstaq/qiskit_superstaq/custom_gates_test.py
```diff
@@ -62,6 +62,11 @@ def test_acecr() -> None:
     assert repr(gate) == "qss.AceCR(rads=3.141592653589793)"
     assert str(gate) == "AceCR(1.0π)"
 
+    # test to  include `np.close() = False` branch  for `__str__` in `custom_gates.py`
```
richrines1: nit
```suggestion
    # test to  include `np.isclose() = False` branch  for `AceCR.__str__()`
```

qiskit-superstaq/qiskit_superstaq/serialization_test.py
```diff
@@ -428,11 +432,6 @@ def test_gate_serialization(
 )
 def test_nonstandard_static_gate_serialization(gate: qiskit.circuit.Instruction) -> None:
     gates = [gate]
-    if hasattr(gate, "ctrl_state"):
-        gate2 = gate.copy().to_mutable()
-        gate2.ctrl_state = 1
-        gates.append(gate2)
```
richrines1, line 434: why was this removed?

richrines1 (commented):
one more question and a nit

qiskit-superstaq/qiskit_superstaq/custom_gates_test.py
daradarry: I was supposed to have removed the comment anyways. It was supposed to be a reminder when I started the PR. I got rid of it.

qiskit-superstaq/qiskit_superstaq/serialization_test.py
daradarry: Seems you advised to drop it in the comment [here](qiskit_superstaq/custom_gates_test.py
) . I think you meant to remove only the `if` condition. I have fixed it.

richrines1 (approved):
lgtm! this was a big project, thanks for tackling it :)

PR #1419, opened 2026-07-08
Adding unit tests for full branch coverage of cirq superstaq

daradarry:
Including full branch for coverage for `cirq superstaq`: Part of #[1381](https://github.com/Infleqtion/client-superstaq/issues/1381) .


cirq-superstaq/cirq_superstaq/job.py
```diff
@@ -144,7 +144,7 @@ def _check_if_unsuccessful(self, index: int | None = None) -> None:
         if status in self.UNSUCCESSFUL_STATES:
             job_ids = self._job_id.split(",")
             ids_to_check = [job_ids[index]] if index else job_ids
-            for job_id in ids_to_check:
+            for job_id in ids_to_check:  # pragma: no branch
```
daradarry, line 147: @richrines1, Not sure how create an empty branch for ids_to_check since it returns a non-empty list from line 146

cirq-superstaq/cirq_superstaq/ops/qubit_gates.py
```diff
@@ -551,7 +551,7 @@ def qubit_index_to_equivalence_group_key(self, index: int) -> int:
         if isinstance(indexed_gate, cirq.InterchangeableQubitsGate):
             gate_key = indexed_gate.qubit_index_to_equivalence_group_key(index_in_gate)
             for i in range(index_in_gate):
-                if gate_key == indexed_gate.qubit_index_to_equivalence_group_key(i):
+                if gate_key == indexed_gate.qubit_index_to_equivalence_group_key(i):  # pragma: no branch
```
daradarry: @richrines1, could not find a suitable branch test because gate_key outputs 0

cirq-superstaq/cirq_superstaq/ops/qubit_gates.py
```diff
@@ -513,7 +513,7 @@ def __init__(self, *component_gates: cirq.Gate) -> None:
             else:
                 self.component_gates += (gate,)
 
-    def qubit_index_to_gate_and_index(self, index: int) -> tuple[cirq.Gate, int]:
+    def qubit_index_to_gate_and_index(self, index: int) -> tuple[cirq.Gate, int]: #  This does not return index
```
daradarry: @richrines1, Function  does not return index as stated in description

cirq-superstaq/cirq_superstaq/ops/qubit_gates.py
richrines1: to get coverage here, we need a gate for which \*some\* qubits can be interchanged (so that it's a subclass of `InterchangeableQubitsGate`, but `gate_key`), but not all (so that `gate_key` isn't the same for every qubit index). So e.g. this should cover that branch:
```python
    gate = css.ParallelGates(cirq.Y, cirq.CCX)
    assert (
        gate(qubits[0], qubits[1], qubits[2], qubits[3])
        == gate(qubits[0], qubits[2], qubits[1], qubits[3]) 
    )
```
(where the two control qubits of the CCX can be interchanged, but not the target)

cirq-superstaq/cirq_superstaq/ops/qubit_gates.py
richrines1: this seems to be working how i'd expect?

`ParallelGates` are composite gates comprising multiple sub-gates acting on different qubits. the purpose of this function is to take a qubit index of the full gate, and use it to determine (1) which of the sub-gates is acting on that qubit, and (2) which of that sub-gate's qubits it corresponds to. So the "index" in the function name refers to the qubit index of the particular sub-gate, whereas the input `index` is that of the overall parallel gate. for example:
```python
# create a 5-qubit ParallelGates with CX, Y, and SWAP sub-gates
gate = css.ParallelGates(cirq.CX, cirq.Y, cirq.SWAP)

assert gate.qubit_index_to_gate_and_index(0) == (cirq.CX, 0)  # qubit 0 of the parallel gate is the first qubit of the CX sub-gate
assert gate.qubit_index_to_gate_and_index(1) == (cirq.CX, 1)  # qubit 1 of the parallel gate is the second qubit of the CX sub-gate
assert gate.qubit_index_to_gate_and_index(2) == (cirq.Y, 0)  # qubit 2 of the parallel gate is the first/only qubit of the Y sub-gate
assert gate.qubit_index_to_gate_and_index(3) == (cirq.SWAP, 0)  # qubit 3 of the parallel gate is the first qubit of the SWAP sub-gate
assert gate.qubit_index_to_gate_and_index(4) == (cirq.SWAP, 1)  # qubit 4 of the parallel gate is the second qubit of the SWAP sub-gate
```

it does seem like this could use a better docstring to make this more clear...

cirq-superstaq/cirq_superstaq/job.py
richrines1: i think in this case it's actually impossible, so fine to just keep the `pragma: no branch`

cirq-superstaq/cirq_superstaq/ops/qubit_gates.py
daradarry: I see. The docstring is definitely  confusing. I will update the docstring with your explanation.

Thanks.

cirq-superstaq/cirq_superstaq/ops/qubit_gates.py
```diff
@@ -513,7 +513,9 @@ def __init__(self, *component_gates: cirq.Gate) -> None:
             else:
                 self.component_gates += (gate,)
 
-    def qubit_index_to_gate_and_index(self, index: int) -> tuple[cirq.Gate, int]:
+    def qubit_index_to_gate_and_index(
+        self, index: int
+    ) -> tuple[cirq.Gate, int]:  #  This does not return index
```
richrines1:
```suggestion
    def qubit_index_to_gate_and_index(self, index: int) -> tuple[cirq.Gate, int]:
```

cirq-superstaq/cirq_superstaq/service.py
```diff
@@ -1299,11 +1299,18 @@ def _objective(
...
+        circuits_and_metadata: dict[str, Any],
+        max_legend_labels: int = 4,
+        legend_labels_count: int = 0,
+    ) -> None:
         """Generates plot and fit data estimating decay parameters.
 
         Args:
             circuits_and_metadata: Dictionary containing cycle benchmarking data.
+            max_legend_labels: int
+            legend_labels_count: int
```
richrines1: can we add full descriptions for these args?

(or alternatively remove them - see below)

cirq-superstaq/cirq_superstaq/service.py
richrines1: would we ever need to change legend_labels_count or max_legend_labels? if not we could also just remove the `if` so we don't have to add the extra arguments, e.g. 
```suggestion
            truncated_label = "A_" + str(ps) + f"={A:.2f} \np_{ps}={p:.2f}"
            legend_labels.append(truncated_label)
            legend_labels_count += 1
            legend_colors.append(plt.gca().lines[-1].get_color())
```

daradarry: I don't think we need to change them. I will remove the `if` coindition instead.

cirq-superstaq/cirq_superstaq/ops/qubit_gates.py
daradarry: Fixed.

cirq-superstaq/cirq_superstaq/ops/qubit_gates.py
```diff
@@ -520,7 +520,8 @@ def qubit_index_to_gate_and_index(self, index: int) -> tuple[cirq.Gate, int]:
             index: The index into a particular member of the `ParallelGates` operation.
 
         Returns:
-            A tuple of the gate at the given index and the index itself.
+            A tuple of the gate at the given index and the qubit index (of the subgate within the
+            parallel gate) .
```
richrines1, line 524:
```suggestion
            parallel gate).
```

richrines1 (approved):
lgtm! (% typo)

PR #1421, opened 2026-07-10
add a movement gate

richrines1:
mostly just a wrapper for a permutation gate, but doesn't require destinations for empty sites. Used like:
```python
op = css.movement_op({q0: q1, q1: q2})
```
indicates a shift of two qubits (initially `q0` and `q1`) by one step (to positions `q1` and `q2`). `q2` is assumed to be an empty site at the start of the operation. `q0` will be empty after the move.

builds on #1420. test requires #1422

css only - will add a qss version in the future


cirq-superstaq/cirq_superstaq/ops/qudit_gates_test.py
```diff
@@ -150,6 +150,123 @@ def test_permutation_gate() -> None:
...
+    with pytest.raises(ValueError, match=r"Missing indices: {1}"):
+        _ = css.MovementGate({0: 2})
+
+    with pytest.raises(ValueError, match=r"same site"):
+        _ = css.MovementGate({0: 2, 1: 2})
+
+
+def test_movement_op() -> None:
+    q2, q3, q4 = cirq.LineQubit.range(2, 5)
+    op = css.movement_op({q2: q3, q3: q4})
```
bharat-thotakura, line 234: Checked that an inverse for `op` (i.e., `op**-1`) here works, but might be good to add a test checking that the moves indeed cancel out

bharat-thotakura (approved):
LGTM

cirq-superstaq/cirq_superstaq/ops/qudit_gates_test.py
richrines1: ah good call, turns out it was only working via `cirq.._InverseCompositeGate`, so it was losing information from the move dictionary as well. just updated both `PermutationGate` and `MovementGate` to implement `__pow__` and added corresponding tests

PR #1422, opened 2026-07-10
bump minimum cirq version

richrines1:
python 3.9 supports up to cirq==1.3.0. this is ~2.5 years old, so afaict no reason to support anything less


bharat-thotakura (approved):
LGTM

PR #1423, opened 2026-07-13
delete temporary `.coverage*` files and exit gracefully if coverage is interrupted

richrines1:
currently an incomplete coverage check (e.g. due to a keyboard interrupt) will leave around some `.coverage*` files, which can be picked up by the subsequent run and lead to an incorrect coverage report


bharat-thotakura (approved):
LGTM

PR #1424, opened 2026-07-13
update stim requirement and fix min-version checks

richrines1:
* fix handling of requirements with python_version constraints in min-version coverage checks
* stim 1.16.0 now has prebuilt wheels for python >= 3.10 on all platforms, so require that and drop the requirement altogether for python 3.9 (it's just a dev requirement anyway)

together this fixes the currently very slow cirq-superstaq min version coverage check


bharat-thotakura (approved):
LGTM (I was also bothered by the slow build of the Cirq standalone coverage check!)

PR #1425, opened 2026-07-16
fix bug in `css.JobV3._terminal_measurement_qubit_indices`

richrines1:
this is used to parse Jaqal results. currently it's incorrect if the compiled circuit uses a subset of the device's qubits (jaqal results are always constructed as if every qubit is measured)

### DIFF
--- a/cirq-superstaq/cirq_superstaq/job.py
+++ b/cirq-superstaq/cirq_superstaq/job.py
@@ -822,9 +822,9 @@ class JobV3(gss.job.Job):
         return dict(combined)
 
     def _terminal_measurement_qubit_indices(self, index: int) -> list[int]:
-        """Determines the ordered physical qubit indices for each measurement in a compiled circuit.
+        """Determines the physical qubit index corresponding to each bit in a measured bitstring.
 
-        Assumes all measurements are terminal.
+        The given circuit must already be compiled, and contain only terminal measurements.
 
         Args:
             index: The index of the compiled circuit for which to return qubit indices.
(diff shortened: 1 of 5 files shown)


cirq-superstaq/cirq_superstaq/job.py
```diff
@@ -833,23 +833,25 @@ def _terminal_measurement_qubit_indices(self, index: int) -> list[int]:
             A list of measured qubit indices, ordered as they should appear in (big-endian)
             bitstrings.
         """
-        compiled_circuit = self.compiled_circuits(index)
-        assert compiled_circuit.are_all_measurements_terminal()
+        indices = super()._terminal_measurement_qubit_indices(index)
+
+        circuit = self.input_circuits(index)
+        if circuit.has_measurements():
+            assert circuit.are_all_measurements_terminal()
```
cdbf1: Nit: Could this PR be a good opportunity to remove bare assertions?

cirq-superstaq/cirq_superstaq/job.py
cdbf1: Is "Assumes" the right word here since we explicitly check for it?

cdbf1: The docs state that this function processes the compiled circuit but the new code uses the input circuit

cdbf1:
Maybe this wasn't ready for review, but it would be worth documenting better what this is trying to achieve

cirq-superstaq/cirq_superstaq/job.py
richrines1: refactored/updated comments and docstrings to hopefully make this more clear

richrines1: probably not, updated!

cirq-superstaq/cirq_superstaq/job.py
richrines1: good call, done.

cdbf1 (approved):
LGTM

PR #1426, opened 2026-07-17
stop autodetecting integration tests in pytest_.py

richrines1:
currently pytest_.py will automatically switch to integration mode if it detects `*integration_test.py` in any passed arg, even when the passed args are e.g. `--exclude *_integration_test.py` :)

instead we should require `--integration` to be passed manually


bharat-thotakura (approved):
LGTM

bharat-thotakura:
Hmm, looks like merging this PR might have affected the daily integration test: https://github.com/Infleqtion/client-superstaq/issues/1427

PR #1429, opened 2026-07-27
Lint check update to support `ruff` `v0.16.0`

bharat-thotakura:
Fixes failing lint check due to the release of `ruff==0.16.0` (changelog [here](https://astral.sh/blog/ruff-v0.16.0))

Applicable changes include the CPY001, RUF036, and PLR0917 rules coming out of preview and associated config updates

### DIFF
--- a/general-superstaq/general_superstaq/superstaq_client.py
+++ b/general-superstaq/general_superstaq/superstaq_client.py
@@ -387,7 +387,7 @@ class _BaseSuperstaqClient:
 class _AbstractUserClient(_BaseSuperstaqClient, ABC):
     """Abstract base class for V2 and V3 clients."""
 
-    def __init__(
+    def __init__(  # noqa: PLR0917
         self,
         client_name: str,
         api_key: str | None = None,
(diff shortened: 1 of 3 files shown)


general-superstaq/general_superstaq/superstaq_client.py
```diff
@@ -387,7 +387,7 @@ def __str__(self) -> str:
 class _AbstractUserClient(_BaseSuperstaqClient, ABC):
     """Abstract base class for V2 and V3 clients."""
 
-    def __init__(
+    def __init__(  # noqa: PLR0917
         self,
         client_name: str,
         api_key: str | None = None,
```
richrines1, line 393: wdyt about just making the rest into keyword-only arguments?
```suggestion
    def __init__(
        self,
        client_name: str,
        api_key: str | None = None,
        *,
```

richrines1 (approved):
lgtm % possible but not necessary change

general-superstaq/general_superstaq/superstaq_client.py
bharat-thotakura: I was going to bring this up during dev meeting for preferences. I just set it to allow 12 positional args to minimize changes to the existing structure (this noqa is the only case with more than that)

I can adjust that in a follow-up PR if we want to discuss the 12 setting in the `pyproject.toml`, otherwise, I'm pro using more keyword arguments

PR #1431, opened 2026-07-30
fix overridden pytest options

richrines1:
our global pytest settings (e.g. passing `--disable-socket`) seem to have been overridden in #1192


bharat-thotakura (approved):
LGTM

PR #1433, opened 2026-07-31
Enforce max five positional args for all of `client-superstaq` 

### LINKED ISSUE #1430: Set allowed number of positional arguments for `ruff` lint back to the default
wdyt about just making the rest into keyword-only arguments?
```suggestion
    def __init__(
        self,
        client_name: str,
        api_key: str | None = None,
        *,
```

_Originally posted by @richrines1 in https://github.com/Infleqtion/client-superstaq/pull/1429#discussion_r3660890193_

bharat-thotakura:
Closes: https://github.com/Infleqtion/client-superstaq/issues/1430 (+ lifts some `ruff` lint rules exclusions for `*conftest.py`)


bharat-thotakura:
(tagged as a breaking change due to potential `TypeError: [...] takes X positional arguments but Y were given` when using `v0.5.68` (or higher) `client-superstaq` packages)

maggie-bao202 (approved):
makes sense-- looks cleaner to have the distinction!

PR #1439, opened 2026-08-10
Add standalone `ty` check in `checks-superstaq`

### LINKED ISSUE #1437: Adding `./checks/ty_.py`?
I propose to add and expose a short script like [`mypy_.py`](https://github.com/Infleqtion/client-superstaq/blob/436833decc3f13cda247f5725dacc036250a90da/checks-superstaq/checks_superstaq/mypy_.py), but for [`ty`](https://docs.astral.sh/ty/).  Similarly to the `ruff` migration, step 1 is to add `ty` as an option/experimental feature; an actual `mypy -> ty` migration of `superstaq` would be deferred to the future.

Tagging @richrines1 and @bharat-thotakura for visibility, but no rush on this 🙂.  And feel free to close the issue with "not planned" if you're against the idea.

bharat-thotakura:
Factors out the addition of `checks-superstaq/checks_superstaq/ty_.py` from #1371 for independent use. Closes https://github.com/Infleqtion/client-superstaq/issues/1437

(also includes a tangential type check fix with https://github.com/Infleqtion/client-superstaq/pull/1439/commits/59c42b72ad523d6a20a74087612fc09d22740b05)


dowusu-antwi (approved):
LGTM

PR #1443, opened 2026-09-07
feat: Use Cirq's Gzip serialization

drwillcharles:
We recently had some very large circuits that caused Superstaq server to run out of memory and crash. I think this change would help significantly to reduce the amount of data stored in the database.

Testing a very large circuit (with depth 36) this change reduces the amount of space used by a factor of 50.

We will also need to do some kind of integration test with the Superstaq server to ensure this doesn't break any tests.


drwillcharles:
Another possible solution instead of implementing `cirq.to_json_gzip()`:
- Use gzip on the complete HTTP request bodies (`Content-Encoding: gzip`). This would require support from the server-side to unzip the payload but would make things simpler.
- On the server-side: change the circuit database fields to GZIP and migrate existing data.

dowusu-antwi:
Dev meeting: Consider closing this.

drwillcharles:
Closing this as it makes sense to GZIP the body on the POST request code instead of the serialization code.

